PR #9, opened 2025-12-01
Start HM as a thread in LM instead of a separate process

eelcoem:
# Improvement

> [!IMPORTANT]
> Use this template only for improvement that do not influence topics covered by contribution requests or bug fixes.

> [!CAUTION]
> Make sure to submit your pull-request as **Draft** until you are ready to have it reviewed by the Committers.

## Description

- Start HM as a thread inside LM instead of as a separate process.
- Added RecoveryClient library to send recovery requests between threads

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,6 +7,10 @@
     // Use the same image as the project's devcontainer
     "image": "ghcr.io/eclipse-score/devcontainer:1.0.0",
 
+    "features": {
+        "ghcr.io/devcontainers/features/docker-outside-of-docker:1": {}
+    },
+
     // Initialize bazel cache on host
     "initializeCommand": "mkdir -p ${localEnv:HOME}/.cache/bazel",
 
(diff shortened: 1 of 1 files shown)


health_monitor/source/score/lcm/saf/daemon/PhmDaemon.hpp
```diff
@@ -154,39 +143,28 @@ class PhmDaemon
...
     /// or add a strategy for recovery.
     /// @todo Add more sophisticated time handling (deviation reporting, reporting on sleep errors)
     /// @todo Rework the signal handling
     /// @todo Monitor the correct increment of the sleep interval
     /* RULECHECKER_comment(0, 4, check_cheap_to_copy_in_parameter, "f_terminateCond is passed as reference\
        for signal handling", true_no_defect) */
-    template <typename LifecycleClientType, typename TerminationSignalPredType>
-    bool startCyclicExec(LifecycleClientType& f_client, const TerminationSignalPredType& f_terminateCond) noexcept
+    template <typename TerminationSignalPredType>
+    bool startCyclicExec(const TerminationSignalPredType& f_terminateCond, std::atomic<EInitCode>& init_status) noexcept
```
MaciejKaszynski: why is `init_status` needed? It seems we are just storing `KNoError` and not using it?

health_monitor/source/score/lcm/saf/daemon/PhmDaemon.hpp
```diff
@@ -87,30 +88,18 @@ class PhmDaemon
 
     /// @brief Wraps the initialization steps of the PHM daemon
     /// (command line parsing, Constructing the workers, adjusting the cycle time, initialization of fixed step timer)
```
MaciejKaszynski: arg parsing is not done anymore

health_monitor/source/score/lcm/saf/daemon/PhmDaemon.hpp
eelcoem: I noticed that the alive monitoring would sporadically fail on startup, so I added the wait on this variable in HealthMonitorThread::start to make sure LM doesn't try to start processes before the HM thread is properly running. I saw that you added a comment about using a cv there, agree that would make it nicer

PR #51, opened 2026-01-26
log: add Rust log init for C++ tests

arkjedrz:
Add logging init as an env in GTest.


src/health_monitoring_lib/cpp/tests/log_init.cpp
```diff
@@ -0,0 +1,44 @@
...
+class LogEnvironment final : public ::testing::Environment
+{
+  protected:
+    void SetUp() override
+    {
+        using namespace score::mw::log::rust;
+
+        StdoutLoggerBuilder builder;
+        builder.Context("TEST")
+            .LogLevel(LogLevel::Info)
```
pawelrutkaq: Verbose please

PR #53, opened 2026-01-29
Worker thread implementation for HealthMonitor

pawelrutkaq:


src/health_monitoring_lib/rust/lib.rs
```diff
@@ -10,50 +10,236 @@
...
+                    // Should not fail since we preallocated enough capacity
+                },
+                Some(DeadlineMonitorState::Available(_)) => {
+                    panic!(
+                        "All monitors must be taken before starting HealthMonitor but {:?} is not taken.",
+                        tag
+                    );
+                },
+                None => {
+                    panic!(
```
pawelrutkaq: in start I am using panic as all errors here are a health monitoring issue, and we force program termination (idea behind is that user cannot handle it and continue program execution). Other options is to return Result and leave it all up to the user. Any opinions?

.vscode/settings.json
```diff
@@ -27,8 +24,12 @@
        "editor.defaultFormatter": "llvm-vs-code-extensions.vscode-clangd",
        "editor.formatOnSave": true
      },
-     "[c]": {
+    "[c]": {
        "editor.defaultFormatter": "llvm-vs-code-extensions.vscode-clangd",
        "editor.formatOnSave": true
      },
+    // Still to align with bazel
```
arkjedrz: Remove this.
- Comment explains nothing.
- If it's not meant to be merged this time - provide it once ready.

examples/cpp_supervised_app/main.cpp
```diff
@@ -36,109 +39,122 @@ struct Config
...
-                std::cerr << "Failed to set QNX thread name" << std::endl;
-            }
-    #elif defined(__linux__)
-            if (prctl(PR_SET_NAME, identifier) < 0) {
-                std::cerr << "Failed to set process name to " << identifier << std::endl;
-            }
-    #endif
+    if (identifier != nullptr)
+    {
+#ifdef __QNXNTO__
```
arkjedrz: Inconsistent -> use `ifdef` or `if defined`

examples/cpp_supervised_app/main.cpp
```diff
@@ -12,20 +12,23 @@
...
 #endif
 
 #include "score/lcm/Monitor.h"
 #include "score/lcm/lifecycle_client.h"
+#include "score/mw/log/rust/stdout_logger_init.h"
+#include <score/hm/common.h>
+#include <score/hm/health_monitor.h>
+#include <thread>
 
 using score::lcm::Monitor;
```
arkjedrz: Unused.

examples/cpp_supervised_app/main.cpp
```diff
@@ -36,109 +39,122 @@ struct Config
...
 void signalHandler(int signal)
 {
-    if (signal == SIGINT || signal == SIGTERM) {
+    if (signal == SIGINT || signal == SIGTERM)
+    {
         exitRequested = true;
-
-    } else if(signal == SIGUSR1) {
+    }
+    else if (signal == SIGUSR1)
```
arkjedrz: Either restore `signal(SIGUSR1, signalHandler);` or remove it along with `stopReportingCheckpoints`.

examples/cpp_supervised_app/main.cpp
```diff
@@ -36,109 +39,122 @@ struct Config
...
+                break;
+
+            case 's':
+                // std::cout << "Instance Specifier is: " << optarg << std::endl;
+                config.instanceSpecifier = optarg;
+                break;
+
+            case 'h':
+                std::cout << "Usage: \n\
+                            -d <The app is configured to measure deadline between 50ms and 150ms. You configure the delay inside this deadline measurement> \n\
```
arkjedrz: So what is `"deadline_2"` for?

src/health_monitoring_lib/cpp/include/score/hm/health_monitor.h
```diff
@@ -40,11 +40,22 @@ class HealthMonitorBuilder final
     /// Adds a deadline monitor to the builder to construct DeadlineMonitor instances during HealthMonitor build.
     HealthMonitorBuilder add_deadline_monitor(const IdentTag& tag, deadline::DeadlineMonitorBuilder&& monitor) &&;
 
+    /// Sets the cycle duration for supervisor API notifications.
+    /// This duration determines how often the health monitor notifies the supervisor that the system is alive.
+    HealthMonitorBuilder with_supervisor_api_cycle(std::chrono::milliseconds cycle_duration) &&;
```
arkjedrz: Make a template out of it, so that any duration type can be used, with internal `duration_cast`.

src/health_monitoring_lib/rust/lib.rs
```diff
@@ -10,50 +10,224 @@
...
 pub struct HealthMonitor {
-    deadline_monitors: HashMap<IdentTag, deadline::DeadlineMonitor>,
+    deadline_monitors: HashMap<IdentTag, Option<DeadlineMonitorState>>,
+    worker: worker::UniqueThreadRunner,
+    supervisor_api_cycle: core::time::Duration,
 }
 
 impl HealthMonitor {
+    /// Retrieves and removes (hand over to user) a deadline monitor associated with the given identifier tag.
+    /// /// # Arguments
```
arkjedrz: Double comment start.

src/health_monitoring_lib/cpp/include/score/hm/health_monitor.h
```diff
@@ -40,11 +40,22 @@ class HealthMonitorBuilder final
...
+    /// This duration determines how often the health monitor checks deadlines.
+    HealthMonitorBuilder with_internal_processing_cycle(std::chrono::milliseconds cycle_duration) &&;
+
     /// Builds and returns the HealthMonitor instance.
     HealthMonitor build() &&;
 
   private:
     internal::DroppableFFIHandle health_monitor_builder_handle_;
+
+    std::chrono::milliseconds supervisor_api_cycle_duration_;
```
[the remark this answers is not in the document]
pawelrutkaq: yes

examples/cpp_supervised_app/main.cpp
pawelrutkaq: All this may be needed later, i intend not to change much in examples until we stabilize all apis

examples/cpp_supervised_app/main.cpp
pawelrutkaq: All this may be needed later, i intend not to change much in examples until we stabilize all apis

examples/cpp_supervised_app/main.cpp
```diff
@@ -12,20 +12,23 @@
...
 
 #ifdef __linux__
-    #include <linux/prctl.h>
-    #include <sys/prctl.h>
+#include <linux/prctl.h>
+#include <sys/prctl.h>
 #endif
 
 #include "score/lcm/Monitor.h"
 #include "score/lcm/lifecycle_client.h"
```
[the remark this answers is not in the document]
pawelrutkaq: All this may be needed later, i intend not to change much in examples until we stabilize all apis

examples/cpp_supervised_app/BUILD
```diff
@@ -30,7 +30,9 @@ cc_binary(
     }),
     visibility = ["//visibility:public"],
     deps = [
+        "//src/health_monitoring_lib:health_monitoring_lib_cc",
         "//src/launch_manager_daemon/health_monitor_lib:hm_shared_lib",
         "//src/launch_manager_daemon/lifecycle_client_lib:lifecycle_client",
+        "@score_baselibs_rust//src/log/stdout_logger_cpp_init",
```
NicolasFussberger: Why do we need the logger from baselibs_rust in a C++ application?
If we want to change the example to use the score logger I would have expected that this is `@score_baselibs//score/mw/log`

src/health_monitoring_lib/rust/lib.rs
SimonKozik: As a library you never know in which context you are executing.
Is that a not so important process that can crash, and all will be OK, or you are part of a process that should perform controlled degradation or a shutdown.

As a rule of a thumb, we can probably assume that crashes are OK during initialization period, so probably reporting running state. Reason for that will be, a safety application should probably fully initialize and start or not start at all.

However I would expect a high level guidelines on this from the project. Does S-CORE have any guidelines on architecture level for this?

src/health_monitoring_lib/rust/worker.rs
```diff
@@ -0,0 +1,321 @@
...
+                    std::thread::sleep(next_sleep_time);
+
+                    let now = std::time::Instant::now();
+
+                    if !monitoring_logic.run() {
+                        info!("Monitoring logic failed, stopping thread.");
+                        break;
+                    }
+
+                    next_sleep_time = interval - now.elapsed();
```
NicolasFussberger: What happens if executing the `monitoring_logic.run()` + scheduling delays takes longer than the configured `ìnterval`?

Would it be a good idea to log a warning if the configured interval is too small?

examples/rust_supervised_app/src/main.rs
```diff
@@ -65,42 +55,48 @@ fn set_process_name() {
...
+        ),
+    );
 
-    let args = Args::parse();
-    let stop = Arc::new(AtomicBool::new(false));
-    let stop_reporting_checkpoints = Arc::new(AtomicBool::new(false));
-    flag::register(signal_hook::consts::SIGTERM, Arc::clone(&stop))?;
-    flag::register(signal_hook::consts::SIGUSR1, Arc::clone(&stop_reporting_checkpoints))?;
-
-    if !lifecycle_client_rs::report_execution_state_running() {
```
NicolasFussberger: We will need the reporting of the Running state, as this is the starting point for launch manager to start checking alive notifications

examples/cpp_supervised_app/main.cpp
```diff
@@ -36,109 +39,122 @@ struct Config
...
-    signal(SIGUSR1, signalHandler);
 
     const auto config = parseOptions(argc, argv);
     if (!config)
     {
         return EXIT_FAILURE;
     }
 
-    score::lcm::Monitor<Checkpoints> monitor(config->instanceSpecifier);
-    score::lcm::LifecycleClient{}.ReportExecutionState(score::lcm::ExecutionState::kRunning);
```
NicolasFussberger: We will need the reporting of the Running state, as this is the starting point for launch manager to start checking alive notifications

src/health_monitoring_lib/rust/lib.rs
```diff
@@ -10,50 +10,224 @@
...
+mod worker;
 
 pub mod deadline;
 pub use common::{IdentTag, TimeRange};
 
 #[derive(Default)]
 pub struct HealthMonitorBuilder {
     deadlines: HashMap<IdentTag, deadline::DeadlineMonitorBuilder>,
+    supervisor_api_cycle: core::time::Duration,
+    internal_processing_cycle: core::time::Duration,
```
NicolasFussberger: I think we should document somewhere the interdependency between those two values.

For example, if I configure internal_processing_cycle=200 and supervisor_api_cycle=500, then I'll always report the alive notification 100ms too late.

Maybe we should also do some validation and log a warning or even return an error if `supervisor_api_cycle` is not a multiple of `internal_processing_cycle`?

examples/rust_supervised_app/src/main.rs
pawelrutkaq: this is done, check few lines below.

src/health_monitoring_lib/rust/worker.rs
pawelrutkaq: From the logic side, nothing, as it will simply continue. From a safety side, the reaction will be delayed. Adding a log is a bit problematic because there are many factors: OS scheduler type, priorities, OS type itself etc. So we can assume that the cycle should be with 10% margin and then log, however, for now I added only ToDo because thread params are not yet exposed.

pawelrutkaq:
Reverted examples as requested.

PR #55, opened 2026-02-02
Implementing report running state in lifecycle manager

shegazyy:
Implementing report running state in lifecycle manager  as per discussion with @NicolasFussberger  on slack.

### DIFF
--- a/examples/cpp_lifecycle_app/main.cpp
+++ b/examples/cpp_lifecycle_app/main.cpp
@@ -13,8 +13,6 @@
 
 #include <optional>
 #include <unistd.h>
-#include <csignal>
-#include <atomic>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


NicolasFussberger:
Changes look good to me👍
One idea would be to adapt the example at https://github.com/eclipse-score/lifecycle/blob/c7d58e953342ce87a7b9f1d389447dafe9001307/examples/cpp_lifecycle_app/main.cpp#L134 to see this also working in running code.

We can then reuse this example app for an automated integration test.

NicolasFussberger:
I cannot build it and test it right now due to a build problem coming from mw::log, but all changes look good to me.

ramceb (approved):
based on approval form @NicolasFussberger

Issue #57, opened 2026-02-03
Improvement: LaunchManager shall log real names for process groups and process group state, instead of hashes


TimoSteuerwaldETAS:
The `IdentifierHash` class is used to identify process group (states) in a memory efficient way, as an object gets also transferred between processes (e.g. via shared memory). This is why it stores only a hash value of the string which is given at construction time. However it confusing to print these hash values in log messages. 

**Proposal:** One possible solution would be a central lookup table where the string which was used during construction time can be retrieved.
**Limitations:** An `IdentifierHash` which was constructed in process A but logged in process B will still only log its hash value, of course. From our point of view this is acceptable. As IdentifierHashs are constructed in LaunchManager, inside LaunchManager it will always output its string value.

PR #58, opened 2026-02-03
 Log real names instead of hashes in IdentifierHash class in Launch Manager

TimoSteuerwaldETAS:
The `IdentifierHash` class is used to identify process group (states) in a memory efficient way. This is why it stores only a hash value of the string which is given at construction time. 

As it confusing to print hash values in log messages, a static map and a operator<< is introduced in the class which provides a easy way to print understandable strings for an IdentifierHash object, by still leaving the object as memory efficient as before.

Related to #57

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,37 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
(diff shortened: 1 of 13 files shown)


SimonKozik (approved):
Changes looks OK to me

FScholPer (approved):

PR #63, opened 2026-02-05
Local test execution setup

WilliamRoebuck:
Add a basic setup to run system-level tests for lifecycle on the host machine. Resolves https://github.com/eclipse-score/score/issues/2403

The minimal example shows a GTEST executable launched by LM that tests simple state transitions.


MaciejKaszynski (approved):

pawelrutkaq (changes requested):
This shall be synced with @PiotrKorkus  hwoto approach tests. There are score tools for it. We shall make it consistent.

PR #99, opened 2026-03-03
hmon: `loom` enablement

arkjedrz:
Enable `loom` in health monitor.


.github/workflows/code_coverage.yml
```diff
@@ -87,7 +87,7 @@ jobs:
             --config=x86_64-linux \
             --config=ferrocene-coverage \
             --nocache_test_results \
```
pawelrutkaq: Whole workflows shall be done using https://github.com/eclipse-score/cicd-workflows/blob/main/.github/workflows/rust-coverage.yml

.github/workflows/code_coverage.yml
```diff
@@ -87,7 +87,7 @@ jobs:
...
-            "$(bazel query 'kind(rust_test, //src/...)')"
+            $(bazel query 'kind(rust_test, //src/...)' | tr '\n' ' ')
```
pawelrutkaq: This is wrongly integrated as we shall use rust_coverage macro in BUILD file.

.github/workflows/code_coverage.yml
arkjedrz: Won't do in this scope, can make another PR once some other changes are merged.

.github/workflows/code_coverage.yml
arkjedrz: Even workflow You provided still uses `bazel test` with some query, but still - changed to a better version.

PR #113, opened 2026-03-11
lifecycle: Add exception handling for CPU affinity

SangKyeong-Jeong:
When there are 32 CPU cores, an issue arises where the process cannot be executed, and when there are more than 32 cores, a problem occurs where fewer cores than expected are used. To resolve this issue, exception handling is added to set the maximum number of cores using the uint32_t type when there are more than 32 cores.


pawelrutkaq (changes requested):
rebase branch

pawelrutkaq (approved):

PR #135, opened 2026-03-27
hmon: add C++ HMON unit tests

arkjedrz:
C++ wrapper unit tests.

Resolves #121

### DIFF
--- a/score/health_monitor/src/cpp/tests/deadline_monitor_test.cpp
+++ b/score/health_monitor/src/cpp/tests/deadline_monitor_test.cpp
@@ -18,13 +18,25 @@
 using namespace score::mw::health;
 using namespace score::mw::health::deadline;
 
-TEST(DeadlineMonitorBuilder, New_Succeeds)
+class DeadlineMonitorBuilderFixture : public ::testing::Test
 {
+  protected:
+    void SetUp() override
+    {
+        RecordProperty("TestType", "interface-test");
+        RecordProperty("DerivationTechnique", "explorative-testing");
+    }
+};
+
+TEST_F(DeadlineMonitorBuilderFixture, New_Succeeds)
+{
+    RecordProperty("Description", "Object successfully constructed.");
     DeadlineMonitorBuilder deadline_monitor_builder;
 }
 
-TEST(DeadlineMonitorBuilder, AddDeadline_Succeeds)
+TEST_F(DeadlineMonitorBuilderFixture, AddDeadline_Succeeds)
 {
+    RecordProperty("Description", "Deadline successfully added.");
     using namespace std::chrono_literals;
     DeadlineTag deadline_tag{"deadline"};
     TimeRange range{50ms, 150ms};
@@ -38,6 +50,9 @@ class DeadlineMonitorFixture : public ::testing::Test
 
     void SetUp() override
     {
+        RecordProperty("TestType", "interface-test");
+        RecordProperty("DerivationTechnique", "explorative-testing");
+
         // Monitor must be obtained from HMON.
         // Initialize deadline monitor builder.
         using namespace std::chrono_literals;
@@ -62,6 +77,7 @@ class DeadlineMonitorFixture : public ::testing::Test
 
 TEST_F(DeadlineMonitorFixture, GetDeadline_Succeeds)
 {
+    RecordProperty("Description", "Deadline successfully obtained using known tag.");
     // Get deadline.
     auto get_deadline_result{deadline_monitor_->get_deadline(DeadlineTag{"deadline"})};
     ASSERT_TRUE(get_deadline_result.has_value());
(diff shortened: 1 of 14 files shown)


score/health_monitor/src/cpp/tests/deadline_monitor_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+ ********************************************************************************/
+
+#include "score/hm/deadline/deadline_monitor.h"
+#include "score/hm/health_monitor.h"
+#include <gtest/gtest.h>
+
+using namespace score::hm;
+using namespace score::hm::deadline;
+
+TEST(DeadlineMonitorBuilder, New_Succeeds)
```
NicolasFussberger: Please use the RecordProperty calls defined in [Score Process](https://eclipse-score.github.io/process_description//main/process_areas/verification/guidance/verification_templates.html#c-properties-template) to define TestType, DerivationTechnique, the Description and possible requirements links.

src/health_monitoring_lib/cpp/tests/time_range_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "gtest/gtest.h"
+#include "score/hm/common.h"
+// #include <gtest/gtest.h>
```
MaciejKaszynski: should use <> way and remove the ""

score/health_monitor/src/cpp/tests/deadline_monitor_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+{
+    DeadlineMonitorBuilder deadline_monitor_builder;
+}
+
+TEST(DeadlineMonitorBuilder, AddDeadline_Succeeds)
+{
+    using namespace std::chrono_literals;
+    DeadlineTag deadline_tag{"deadline"};
+    TimeRange range{50ms, 150ms};
+    auto deadline_monitor_builder{DeadlineMonitorBuilder{}.add_deadline(deadline_tag, range)};
```
NicolasFussberger, line 43: How to know if construction succeeded?
I am wondering if there is an assert missing. If it were to throw an exception in case of error there could also be ASSERT_NO_THROW used to make it explicit.

src/health_monitoring_lib/BUILD
```diff
@@ -160,8 +160,13 @@ rust_test(
 cc_gtest_unit_test(
     name = "cpp_tests",
     srcs = [
+        "cpp/tests/deadline_monitor_test.cpp",
         "cpp/tests/health_monitor_test.cpp",
+        "cpp/tests/heartbeat_monitor_test.cpp",
+        "cpp/tests/integrated_test.cpp",
```
MaciejKaszynski: Can we have this one as a separate bazel target as it seems this is more of a integration test, so don't think we should mix test kinds into the same binary,

MaciejKaszynski (changes requested):

score/health_monitor/src/cpp/tests/heartbeat_monitor_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+    auto hmon{std::move(hmon_build_result.value())};
+
+    // Get heartbeat monitor.
+    auto get_heartbeat_monitor_result{hmon.get_heartbeat_monitor(heartbeat_monitor_tag)};
+    ASSERT_TRUE(get_heartbeat_monitor_result.has_value());
+    auto heartbeat_monitor{std::move(get_heartbeat_monitor_result.value())};
+
+    // Check heartbeat is not failing.
+    heartbeat_monitor.heartbeat();
+}
```
NicolasFussberger, line 59: Probably more testing is required for Heartbeat monitor, but not sure if this shall be done in scope of this PR.

NicolasFussberger (changes requested):

score/health_monitor/src/cpp/tests/deadline_monitor_test.cpp
arkjedrz: It is assumed here that APIs either abort or return `expected`. In this case - abort should happen.

score/health_monitor/src/cpp/tests/heartbeat_monitor_test.cpp
arkjedrz: I should probably make it more clear in the description that those unit tests are for the **C++ wrapper** on top of Rust implementation. More detailed unit tests to heartbeat monitor are available in Rust implementation.

MaciejKaszynski (approved):

NicolasFussberger (approved):

PR #152, opened 2026-04-08
Fix leftover shared memory when testing

WilliamRoebuck:
The test framework was causing launch manager not to shut down correctly, fixed by terminating process groups rather than specific pids.

Additionally:
- Fix leftover memory in partial Socket::create() failure 
- Move memory name exists check inside socket.hpp
- Attempt to unlink existing memory if it is found


NicolasFussberger:
Looks good to me.
The majority of changes are actually only formatting.

NicolasFussberger (approved):

MaciejKaszynski (approved):

PR #153, opened 2026-04-09
Remove the supervision-failure specific recovery state in Launch Manager

NicolasFussberger:
## Motivation

Historical Context: HealthMonitor and LaunchManager used to be two separate daemon processes that were merged into a single daemon.
Due to this reason, there were two separately configured recovery states. One is triggered on supervision failure, the other in case of process crashes/startup failure.
There is no need to have differently configured recovery states, as those failures are treated the same. The new configuration schema also does not expose this configuration anymore though the distinction still exists in the code. The configuration mapping will always set both recovery state to be identical, as those failure should be treated the same way.

## Changes

* Removes the separate HealthMonitor-specific recovery state.
* The RecoveryClient interface is simplified such that HealthMonitor only communicates which process group experienced the supervision failure (not requesting a specific recovery state).

## Benefit

With this change, there is only a single recovery state and LaunchManager transitions to this recovery state the same way for supervision failure, process crash, etc.
Obsolete configuration settings are removed from the internal configuration model.
The user-facing configuration stays unaffected, making this change backwards-compatible.

Closes: https://github.com/eclipse-score/lifecycle/issues/132


src/launch_manager_daemon/health_monitor_lib/config/hm_flatcfg_generated.h
MaciejKaszynski, line 1: I created an issue to remove the generated files when the central rule to generate them is there. https://github.com/eclipse-score/lifecycle/issues/154

src/launch_manager_daemon/src/process_group_manager/processgroupmanager.cpp
```diff
@@ -527,55 +527,51 @@ inline void ProcessGroupManager::recoveryActionHandler()
...
-            recovery_client_->setResponseError(recovery_request->promise_id_, score::lcm::ExecErrc::kInvalidArguments);
+            LM_LOG_ERROR() << "recoveryActionHandler: Unknown process group " << recovery_request->process_group_identifier_;
             continue;
         }
 
-        IdentifierHash old_state = pg->getProcessGroupState();
+        const IdentifierHash old_state = pg->getProcessGroupState();
+        const IdentifierHash recovery_state =
+                configuration_manager_.getNameOfRecoveryState(pg->getProcessGroupName());
         GraphState graph_state = pg->getState();
```
MaciejKaszynski: can be const

MaciejKaszynski (changes requested):

src/launch_manager_daemon/src/process_group_manager/processgroupmanager.cpp
NicolasFussberger: done

MaciejKaszynski (approved):

TimoSteuerwaldETAS (approved):
Looks good! 👍

pawelrutkaq (approved):

Issue #181, opened 2026-04-25
[Bug][Blocker] Can not upgrade dependencies


MaximilianSoerenPollak:
### Description

Currently trying to upgrade dependencies on this repository in preperation for ref-int release. 
This is currently not possible. 

This is a blocker, as this repo needs to be updated in oder to be used in ref-int.

### Analysis results

In the [PR here](https://github.com/eclipse-score/lifecycle/pull/180) you can see my attempt to upgrade the dependencies / what has been changed. 

However, trying to update the Module.bazel.lock file is impossible as any attempt fails via: 
```
ERROR: Error computing the main repository mapping: swift_argument_parser@1.3.1.1 depends on rules_swift@1.18.0 with compatibility level 1, but flatbuffers@_ depends on rules_swift@2.1.1 with compatibility level 2 which is different
```

I did already replace the hash for the `hedron_compile_commands` as according to Moduel.bazel.lock this is where that apprently comes from, but to no avail.

### Solution

Update dependencies & Update Module.bazel.lock

### Error Occurrence  Rate

Reproducible

### How to reproduce

Checkout the branch I mentioned above and run any of these two commands: 
`bazel mod graph --lockfile_mode=update`
`bazel build --lockfile_mode=update //src/... //examples/...`

And you should see the error

### Classification

Minor

### First Affected Release

not released (main)

### Last Affected Release

not released (main)

### Expected Fixed Release

before release (main)

### Category

- [ ] Safety Relevant
- [ ] Security Relevant

MaciejKaszynski:
I have seen this before, I got a working build when I removed the score_logging dependency. Not sure if this means that the error is coming from there though.

PR #185, opened 2026-04-27
Removing score_logging dependency & update dependency versions for 0.7 release

MaciejKaszynski:
Removing the `score_logging` dependency as we do not rely on remote logging.
This resolves:
https://github.com/eclipse-score/lifecycle/issues/181

- This allowed to update other dependency versions

Note there is an issue with flatbuffers, check github.com/eclipse-score/lifecycle/issues/186


MODULE.bazel
```diff
@@ -17,25 +17,31 @@ module(
 
 # Bazel global rules
 bazel_dep(name = "rules_pkg", version = "1.1.0")
-bazel_dep(name = "rules_python", version = "1.8.3")
+bazel_dep(name = "rules_python", version = "1.8.5")
 bazel_dep(name = "rules_rust", version = "0.61.0")
```
NicolasFussberger: rules_rust 0.68.1-score
according to https://github.com/eclipse-score/score/issues/2791

MaciejKaszynski: Fixed on [b2966c6](https://github.com/eclipse-score/lifecycle/pull/185/commits/b2966c6b5dd471fbd814e446d951365bd940e822)

NicolasFussberger (dismissed):

MODULE.bazel
```diff
@@ -17,25 +17,24 @@ module(
...
-bazel_dep(name = "rules_cc", version = "0.2.16")
+bazel_dep(name = "rules_python", version = "1.8.5")
+bazel_dep(name = "rules_rust", version = "0.68.1-score")
+bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "rules_oci", version = "2.3.0")
 bazel_dep(name = "aspect_rules_lint", version = "2.0.0")
 bazel_dep(name = "buildifier_prebuilt", version = "8.2.0.2")
 bazel_dep(name = "platforms", version = "1.0.0")
-bazel_dep(name = "flatbuffers", version = "25.9.23")
+bazel_dep(name = "flatbuffers", version = "v25.12.19")
```
NicolasFussberger: Should we stay with bazel_dep(name = "flatbuffers", version = "25.9.23")
as this is used in reference integration right now

MODULE.bazel
NicolasFussberger: override to be removed, use score_crate 0.0.9

NicolasFussberger: override to be removed

NicolasFussberger: override to be removed?

pawelrutkaq (commented):
why this would make sense ? All exec shall use logging backend from score logging.

MaciejKaszynski:
> why this would make sense ? All exec shall use logging backend from score logging.

The logging interface is coming from baselibs, that is what we are already using. I also got the info from Ragha that the score_logging/mw/log target will be soon deprecated and it is expected for users to use score_baselibs.

NicolasFussberger (approved):

pawelrutkaq:
> > why this would make sense ? All exec shall use logging backend from score logging.
> 
> The logging interface is coming from baselibs, that is what we are already using. I also got the info from Ragha that the score_logging/mw/log target will be soon deprecated and it is expected for users to use score_baselibs.

Thats true, bue app needs to link backend, so we wer ejsut maybe linking wrong target from score_logging but we need to link it

PR #188, opened 2026-05-01
Fix no comms reporting crash

WilliamRoebuck:
Add test for and fix a case where opening shared memory before attempting to report running from a non-reporting process would cause a crash.
Some formatting changes have been made but are separated into their own commits.

Note: This issue only arises in case of a misconfiguration. Configuring a process as type "Native" but then wrongly reporting a "Running" state.


tests/integration/incorrect_config_no_comms/control_client_mock.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include <gtest/gtest.h>
+
+#include <score/lcm/lifecycle_client.h>
+#include <score/lcm/control_client.h>
+#include <score/lcm/identifier_hash.hpp>
+#include "tests/utils/test_helper/test_helper.hpp"
+
+score::lcm::ControlClient client;
+
+TEST(NoComms, Daemon) {
```
NicolasFussberger: This file looks like a duplicate of tests/integration/smoke/control_daemon_mock.cpp
Do we actually need a state manager for this test?

tests/integration/incorrect_config_non_reporting/non_reporting_process.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "tests/utils/test_helper/test_helper.hpp"
+#include <csignal>
+
+TEST(NoComms, Process)
+{
+    auto fd = shm_open("some_shared_memory", O_CREAT | O_RDWR, 0);
+    shm_unlink("some_shared_memory");
```
NicolasFussberger, line 36: Could you provide a comment what is the purpose of opening this shared memory in the test?

tests/integration/incorrect_config_non_reporting/test_incorrect_config_non_reporting.py
```diff
@@ -0,0 +1,43 @@
...
+from tests.utils.testing_utils.test_results import (
+    check_for_failures,
+    download_xml_results,
+)
+from attribute_plugin import add_test_properties
+
+
+@add_test_properties(
+    partially_verifies=[],
+    test_type="interface-test",
```
NicolasFussberger, line 22: could you add here also description?

Maybe something like this
```python
@add_test_properties(
    partially_verifies=[],
    test_type="interface-test",
    description="""
     Objective: Test robustness of LifecycleClient API
     Input: Component wrongly configured as `native` application type, reports the Running state to LaunchManager.
     Expected Outcome: Reporting Running state fails, LifecycleClient API returns an error.
     """
    derivation_technique="explorative-testing",
)
```

tests/integration/incorrect_config_no_comms/test_incorrect_config_no_comms.py
```diff
@@ -0,0 +1,43 @@
...
+    """Test behavior when a non-reporting process attempts to report running"""
+    run_until_file_deployed(
+        target=target,
+        binary_path=str(remote_test_dir / "launch_manager"),
+        file_path=remote_test_dir.parent / "test_end",
+        cwd=str(remote_test_dir),
+        timeout_s=2.0,
+    )
+
+    download_xml_results(target, remote_test_dir, test_output_dir)
```
NicolasFussberger: Probably out of scope here.
But would be great if we can wrap this in a fixture to be executed after each test, so we do not need to duplicate this in every test.

tests/integration/incorrect_config_non_reporting/BUILD
```diff
@@ -0,0 +1,81 @@
+# *******************************************************************************
```
NicolasFussberger, line 1: I wonder if we can simplify the wording for this test?
I think `no_comms` will not be understood if you are not deep in the launch_manager implementation. On this integration test level I think it would be great if we can the terminology from the API / configuration.
Maybe something using the terms from the configuration like applicationType, Native, Reporting ?

src/control_client_lib/src/control_client_impl.cpp
```diff
@@ -79,6 +80,16 @@ ControlClientImpl::ControlClientImpl(std::function<void(const score::lcm::Execut
             instance_created_ = true;
         }
 
+    struct stat stats;
+    const auto fstat_ret = fstat(internal::osal::IpcCommsSync::sync_fd, &stats);
+    // Check size we have access of to avoid a crash if fd is not pointing to correct data
+    const auto needed_size = sizeof(internal::osal::IpcCommsSync) + sizeof(internal::ControlClientChannel);
+    if (fstat_ret == -1 || stats.st_size != static_cast<off_t>(needed_size)) {
+        std::cerr << "Control client channel at sync_fd is not valid!" << std::endl;
```
NicolasFussberger: probably need to use the log wrapper here

tests/integration/incorrect_config_no_comms/control_client_mock.cpp
WilliamRoebuck: Fixed in ed6d753931d04e4770b200ada12d97859bce27c1

tests/integration/incorrect_config_non_reporting/non_reporting_process.cpp
WilliamRoebuck: I added a comment and made the fd acquisition more robust 8a5761b90015c5119df21137abb185d0268e650d

tests/integration/incorrect_config_non_reporting/test_incorrect_config_non_reporting.py
WilliamRoebuck: Done in 6c9ec6551dee6ab020bdd47fa73e844679e73e3f, added as a docstring rather than as part of the test properties (not a valid parameter)

tests/integration/incorrect_config_no_comms/test_incorrect_config_no_comms.py
WilliamRoebuck: Done in 3f6b9c292eb271a09be854636f21a975e16e766b

tests/integration/incorrect_config_non_reporting/BUILD
WilliamRoebuck: Changed "no_comms" to "non-reporting", I think this is clearer than Native as a human-readable description, rather than using the exact wording from the config. I've included Native in the test description to be more specific if needed. What do you think? 0771219888b28db84056e3c1194d876817e57f73

src/control_client_lib/src/control_client_impl.cpp
WilliamRoebuck: Done in fff5e93d643f3a5a45f53b0371027a567dfed957

NicolasFussberger (approved):

Issue #179, opened 2026-04-24
FIT: Launch component which involves process launch monitoring


TimoSteuerwaldETAS:
### What

**Verifies requirements:**
- [feat_req__lifecycle__failure_detect](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__failure_detect): Process launch monitoring

**Depends on: #172**

**Configuration:**
- Extend config in #172 by
  - Run Target `run_target_crashing_app_on_startup_twice` which contains a Component `component_crashing_on_startup_twice`
  - Run Target `run_target_crashing_app_on_startup_always` which contains a Component `component_crashing_on_startup_always`
- Ready Recovery Action on `component_crashing_on_startup_twice` and `component_crashing_on_startup_always` configured to restart process 3 times before failing
- Recovery Action on `run_target_crashing_app_on_startup_always` configured to switch to Run Target `run_target_initial`
- The process involved in `component_crashing_on_startup_twice` should crash before reporting kRunning at the first two tries (for example count the number of process starts via text file), at the third try it should not crash and report kRunning

**Test steps:**
- Start Launch Manager process
- Via `score::lcm::ControlClient::ActivateRunTarget(..)` activate `run_target_crashing_app_on_startup_twice`
- After `run_target_crashing_app_on_startup_twice` is activated, via `score::lcm::ControlClient::ActivateRunTarget(..)` activate `run_target_crashing_app_on_startup_always`

**Expected behavior:**
- Launch Manager is started successfully and Run Target `run_target_initial` is activated initially
- Run Target `run_target_crashing_app_on_startup_twice` is activated successfully (optional: track process restart count)
- Run Target `run_target_crashing_app_on_startup_always` can not be activated (optional: track process restart count)
- Recovery Action is executed and Run Target `run_target_initial` is activated 

### Acceptance Criteria (DoD)

Test specification and implementation exists. Test passes.

Issue #178, opened 2026-04-24
FIT: Recovery Action due to complex monitoring failure


TimoSteuerwaldETAS:
### What

**Verifies requirements:**
- [feat_req__lifecycle__lm_ext_watchdog_cfg](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__lm_ext_watchdog_cfg): Launch manager external monitoring configuration
- [feat_req__lifecycle__ext_monitor_notify](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__ext_monitor_notify): Process state notification
- [feat_req__lifecycle__monitoring_processes](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__monitoring_processes): Monitoring and recovery: adopted process monitoring

**Depends on: #172**

**Configuration:**
- Extend config in #172 by an additional Run Target `run_target_complex_monitoring` which contains a Component `component_complex_monitoring`
- Recovery Action on `run_target_complex_monitoring` configured to switch to Run Target `run_target_initial`
- The process involved in `component_complex_monitoring` has deadline supervision configured, which does involve the health monitoring lib to report its checkpoints

**Test steps:**
- Start Launch Manager process
- Via `score::lcm::ControlClient::ActivateRunTarget(..)` activate `run_target_complex_monitoring`
- 1s after `component_complex_monitoring` reports running and `run_target_complex_monitoring` is activated, process involved stops reporting checkpoints to health monitoring lib

**Expected behavior:**
- Launch Manager is started successfully and Run Target `run_target_initial` is activated initially
- Run Target `run_target_complex_monitoring` is activated successfully
- Recovery Action is executed, switch back to `run_target_initial`

### Acceptance Criteria (DoD)

Test specification and implementation exists. Test passes.

Issue #177, opened 2026-04-24
FIT: Recovery Action due to process crash on runtime


TimoSteuerwaldETAS:
### What

**Verifies requirements:**
- [feat_req__lifecycle__monitor_abnormal_term](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__monitor_abnormal_term): Process crash monitoring

**Depends on: #172**

**Configuration:**
- Extend config in #172 by an additional Run Target `run_target_crashing_app_on_runtime` which contains a Component `component_crashing_on_runtime`
- Recovery Action on `run_target_crashing_app_on_runtime` configured to switch to Run Target `run_target_initial`
- The process involved in `component_crashing_on_runtime` should crash after reporting kRunning

**Test steps:**
- Start Launch Manager process
- Via `score::lcm::ControlClient::ActivateRunTarget(..)` activate `run_target_crashing_app_on_runtime`
- 1s after `component_crashing_on_runtime` reports running and `run_target_crashing_app_on_runtime` is activated, process involved crashes

**Expected behavior:**
- Launch Manager is started successfully and Run Target `run_target_initial` is activated initially
- Run Target `run_target_crashing_app_on_runtime` is activated successfully
- Recovery Action is executed, switch back to `run_target_initial`

### Acceptance Criteria (DoD)

Test specification and implementation exists. Test passes.

Issue #174, opened 2026-04-24
FIT: Switch to another Run Target


TimoSteuerwaldETAS:
### What

**Verifies requirements:**
- [feat_req__lifecycle__request_run_target_start](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__request_run_target_start): Request run target launch
- [feat_req__lifecycle__switch_run_targets](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__switch_run_targets): Switch between run targets
- [feat_req__lifecycle__process_state_comm](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__process_state_comm): Process state
- [feat_req__lifecycle__process_termination](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__process_termination): Terminating process
- [feat_req__lifecycle__terminationn_dependency](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__terminationn_dependency): Handling process dependency in termination
- [feat_req__lifecycle__launch_manager_shutdown](https://eclipse-score.github.io/score/main/features/lifecycle/requirements/index.html#feat_req__lifecycle__launch_manager_shutdown): Normal shutdown

**Depends on: #172**

**Configuration:**
- Configuration with the following run dependency tree:
  - `run_target_initial` - which is the initial run target - depends on component `component_initial`
    - `component_initial`: No dependencies
  - `run_target_a`: Depends on run target `run_target_c` and component `component_a`
    - `component_a`: Depends on `component_b`
      - `component_b`: No dependencies
    - `run_target_c`: Depends on run_target `run_target_e` and component `component_d`
      - `component_d`: No dependencies
      - `run_target_e`: Depends on `component_e`
        - `component_e`: No dependencies

**Test steps:**
- Start Launch Manager process
- Via `score::lcm::ControlClient::ActivateRunTarget(..)` activate `run_target_a`
- 1s after `component_a` reports running and `run_target_a` is activated, via `score::lcm::ControlClient::ActivateRunTarget(..)` `run_target_initial` gets activated

**Expected behavior:**
- Launch Manager is started successfully and Run Target `run_target_initial` is activated initially
[...]

PR #194, opened 2026-05-12
Initial integration tests

WilliamRoebuck:
Implemented the following tests:
- #172
- #174
- #177
- #178
- #179

Improvements/fixes to test helper & smoke test


tests/integration/crash_on_startup/process_crashing_on_startup_twice.cpp
```diff
@@ -0,0 +1,49 @@
...
+            std::cout << "Failed to deploy marker file!" << std::endl;
+        }
+        std::abort();
+    }
+}
+
+int main()
+{
+    deploy_and_crash_if_not_present("crashed_once");
+    deploy_and_crash_if_not_present("crashed_twice");
```
NicolasFussberger: Where are these files being cleaned up again?

tests/integration/crash_on_startup/crash_on_startup.py
```diff
@@ -0,0 +1,42 @@
...
+from attribute_plugin import add_test_properties
+
+
+@add_test_properties(
+    fully_verifies=["feat_req__lifecycle__failure_detect"],
+    test_type="requirements-based",
+    derivation_technique="requirements-analysis",
+)
+def test_crash_on_startup(target, setup_test, test_output_dir, remote_test_dir):
+    """Integration test verifying that the launch manager correctly handles processes
```
NicolasFussberger: I think we need some more description of what is being tested here and what is the expected behaviour.
The [process](https://eclipse-score.github.io/process_description/main/process_areas/verification/guidance/verification_specification.html#test-specification)  is asking for this information:

> * The objective of the test.
> * Inputs
> * Expected outcome (e.g. “A success message is displayed.”)
> * Test environment (e.g. network configuration, clean system state)

I know there is more info in control_client_mock.cpp available, but I assume we need this info in a test spec as part of a test report at least at a high level.

Maybe something like:
```python
"""
Objective: Verifies that the launch manager correctly handles processes that crash before reporting kRunning.

Case 1: Process crashes before Running state but eventually starts up successfully before the configured number of restart attempts is exceeded.
Expected Behaviour: Process startup successful, RunTarget activation successful

Case 2: Process keeps crashing, exceeding the number of restart attempts.
Expected Behaviour: Process startup fails, LaunchManager executes recovery action.
"""
```

tests/integration/complex_monitoring/control_client_mock.cpp
```diff
@@ -0,0 +1,54 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include <gtest/gtest.h>
+#include <unistd.h>
+
+#include "tests/utils/test_helper/test_helper.hpp"
+#include <score/lcm/control_client.h>
+#include <score/lcm/lifecycle_client.h>
+
+score::lcm::ControlClient client;
```
NicolasFussberger: After merging https://github.com/eclipse-score/lifecycle/pull/191 we cannot create the ControlClient in global scope as mw::log might not be initialized yet and does not support logging at this point in time.
Same for all other tests

NicolasFussberger (changes requested):

tests/integration/process_launch_args/process_initial.cpp
```diff
@@ -0,0 +1,50 @@
...
+    // Then, the process is started and:
+    TEST_STEP("Report kRunning")
+    {
+        auto result = score::lcm::LifecycleClient{}.ReportExecutionState(score::lcm::ExecutionState::kRunning);
+        EXPECT_TRUE(result.has_value()) << "ReportExecutionState() failed: " << result.error().Message();
+    }
+    TEST_STEP("Check args")
+    {
+        ASSERT_GT(g_argc, 1) << "Not enough arguments";
+        EXPECT_STREQ(g_argv[1], "S-CORE rules!") << "Second argument was not as expected";
```
NicolasFussberger, line 45: Should this also check that the first argument is the binary name?

tests/integration/smoke/control_daemon_mock.cpp
```diff
@@ -10,41 +10,48 @@
...
+    ASSERT_TRUE(check_clean({test_end_location}));
     TEST_STEP("Control daemon report kRunning") {
         // report kRunning
         auto result = score::lcm::LifecycleClient{}.ReportExecutionState(score::lcm::ExecutionState::kRunning);
         ASSERT_TRUE(result.has_value()) << "client.ReportExecutionState() failed: " << result.error().Message();
     }
-    TEST_STEP("Activate RunTarget Running") {
+
+    // We have to wait for the initial state transition to fully complete, otherwise unexpected failures can occur
+    sleep(1);
```
NicolasFussberger, line 35: For these workarounds, please always link the open problem https://github.com/eclipse-score/lifecycle/issues/198

tests/integration/crash_on_startup/crash_on_startup.py
```diff
@@ -0,0 +1,49 @@
...
+        binary_path=str(remote_test_dir / "launch_manager"),
+        file_path=remote_test_dir.parent / "test_end",
+        cwd=str(remote_test_dir),
+        timeout_s=6.0,
+    )
+
+    download_xml_results(target, remote_test_dir, test_output_dir)
+    all_files, failing_files = check_for_failures(test_output_dir)
+    assert len(all_files) == 2, f"Didn't find the expected number of files {all_files}"
+    assert len(failing_files) == 0, f"Found failures in files {failing_files}"
```
NicolasFussberger: I wonder if we can abstract from this somehow.
This comparison of created files seems to be one of the main asserts deciding whether the test fails or succeeds.

Currently, it is based on comparing a raw number where it is easy to introduce one-off mistake. Maybe something like listing the components that were expected to start or something like that which is more explicit?

tests/integration/smoke/control_daemon_mock.cpp
WilliamRoebuck: Fixed in https://github.com/eclipse-score/lifecycle/pull/194/commits/a0483892e535a4a66838a211ffd47a3491df6c8f

tests/integration/process_launch_args/process_initial.cpp
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/194/commits/a0483892e535a4a66838a211ffd47a3491df6c8f

tests/integration/crash_on_startup/crash_on_startup.py
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/194/commits/55c9b98e0323aa900620dbcc6222534277400e26 , also updated them to use the fixture

NicolasFussberger (approved):

TimoSteuerwaldETAS (approved):
Looks good!

Issue #50, opened 2026-01-26
Use Sanitizers


pawelrutkaq:
Investigate and setup job for sanitizers as recommended in https://github.com/eclipse-score/lifecycle/pull/48#issuecomment-3798665679.

NicolasFussberger:
Check in mw::com they are already using sanitizers

MaciejKaszynski:
I think ideally the setup would be that we run each tests with the final build (no sanitizers) and the sanitizer version so that we are actually testing what we are releasing, and also passing the sanitizers.

I think also for running with sanitizers it's good to build without optimizations as the compiler is allowed to optimize out UB.

Maybe it would also be a good idea to make a bazel rule wrapper/macro so that we don't have to define these tests twice (once without sanitizers and once with). Also would be nice to have the sanitizer bazel tag for filtering.

FScholPer:
@MaciejKaszynski  Whats the status here?

MaciejKaszynski:
> [@MaciejKaszynski](https://github.com/MaciejKaszynski) Whats the status here?

There is one outstanding task, there was a PR to fix it however it seems the contributor isn't active anymore. I'll create a copy of the branch and apply the review fix to finish off the story.

Issue #175, opened 2026-04-24
cpp_supervised_app crashes sometimes during startup


NicolasFussberger:
### Description

While executing the demo scenario described [here](https://github.com/eclipse-score/lifecycle/blob/main/examples/README.md#demo-walkthrough) the [cpp_supervised_app](https://github.com/eclipse-score/lifecycle/tree/main/examples/cpp_supervised_app) sometimes crashes during startup.

The crash is coming from a panic in health_monitoring_library:
```
thread '<unnamed>' (110) panicked at library/std/src/io/stdio.rs:1166:9:
failed printing to stdout: Resource temporarily unavailable (os error 11)
stack backtrace:
2026/04/24 11:16:23.9383844 214149635 000 DEMO NONE DFLT log info verbose 1 Reporting kRunning to Launch Manager
2026/04/24 11:16:23.9383844 214149635 000 DEMO NONE DFLT log info verbose 1 Signal handler started
   0:           0x481225 - <<std[ea9f4e93d12430d]::sys::backtrace::BacktraceLock>::print::DisplayBacktrace as core[d1bb96a9607206a1]::fmt::Display>::fmt
   1:           0x49b997 - core[d1bb96a9607206a1]::fmt::write
   2:           0x435cd6 - std[ea9f4e93d12430d]::io::default_write_fmt::<std[ea9f4e93d12430d]::sys::stdio::unix::Stderr>
   3:           0x480102 - <std[ea9f4e93d12430d]::sys::backtrace::BacktraceLock>::print
   4:           0x452d0c - std[ea9f4e93d12430d]::panicking::default_hook::{closure#0}
   5:           0x45538b - std[ea9f4e93d12430d]::panicking::default_hook
   6:           0x45554e - std[ea9f4e93d12430d]::panicking::panic_with_hook
   7:           0x47fc68 - std[ea9f4e93d12430d]::panicking::panic_handler::{closure#0}
   8:           0x47c9e9 - std[ea9f4e93d12430d]::sys::backtrace::__rust_end_short_backtrace::<std[ea9f4e93d12430d]::panicking::panic_handler::{closure#0}, !>
   9:           0x45345d - __rustc[43d5200f6d4d6223]::rust_begin_unwind
  10:           0x407d3c - core[d1bb96a9607206a1]::panicking::panic_fmt
  11:           0x4462c8 - std[ea9f4e93d12430d]::io::stdio::_print
[...]

PR #200, opened 2026-05-15
Add sanitizer support via score_cpp_policies

### LINKED ISSUE #232: Integrate ASan/UBSan/LSan via score_cpp_policies
### What

Enable AddressSanitizer, UndefinedBehaviorSanitizer, and LeakSanitizer for this repo's tests, using score_cpp_policies' shared sanitizer infrastructure (wrapper, suppressions, env templates) instead of a bespoke setup.

### Acceptance Criteria (DoD)

- bazel test --config=asan_ubsan_lsan --config=x86_64-linux //src/... //tests/... runs clean (with asan/ubsan/lsan aliases available)
- CI job runs the config on PRs, merge groups, and pushes to main
- Any test/target excluded via no-asan is documented with the reason for the exclusion
- README documents how to run locally

### How

- Add score_cpp_policies as a MODULE.bazel dev_dependency
- Add --config=asan_ubsan_lsan (+ aliases) to .bazelrc, including --build_tests_only so wildcard expansion doesn't try to build sanitizer-incompatible targets
- Add .github/workflows/sanitizers.yml job
- Tag the small set of known-false-positive tests (e.g. Rust/uninstrumented-code links, fake-stack alignment issues) no-asan

VukPavicRTRK:
# Description

Integrates ASan, UBSan, and LSan using the centralized `score_cpp_policies` infrastructure instead of a standalone implementation.

> TSan is tracked separately in Issue #50 due to CI ASLR and suppression-file considerations.

# Changes

- **`MODULE.bazel`**
  - Added `score_cpp_policies` as a `dev_dependency` via `git_override`.

- **`.bazelrc`**
  - Added sanitizer configs:
    - `asan_ubsan_lsan`
    - `asan`
    - `ubsan`
    - `lsan`
  - Added `--build_tests_only` to prevent excluded targets from being built.
  - Added sanitizer-specific tag filters (`no-asan`, `no-ubsan`, `no-lsan`).

- **`.github/workflows/sanitizers.yml`**
  - Added CI job running sanitizer tests on PRs, merge groups, and pushes to `main`.

- **`src/health_monitoring_lib/BUILD`**
  - Marked Rust-based and mixed Rust/C++ tests as `no-asan`.

- **`src/.../common/concurrency/BUILD`**
  - Marked `mpmc_concurrent_queue_test` as `no-asan` due to a known false positive.

- **`tests/utils/bazel/integration.bzl`**
  - Excluded Docker integration tests from sanitizer runs.

# Notes

Tag exclusions are limited to known false positives or environmental constraints and can be removed once the underlying issues are resolved.

# Running Locally

```bash
bazel test --config=asan_ubsan_lsan --config=x86_64-linux //src/... //tests/...
```

Closes #232.


FScholPer:
be aware there is also going work in that direction in score_cpp_policies repo

VukPavicRTRK:
> be aware there is also going work in that direction in score_cpp_policies repo

Should lifecycle continue with the self-contained sanitizer implementation in this PR as a short-term unblocker, planning a later migration to the shared score_cpp_policies module once its sanitizer structure stabilizes or do we want to pivot now and integrate directly with the centralized sanitizer infrastructure despite the potential dependency and release-readiness risks?

FScholPer:
My opinion: I would say we should bring in our thoughts into score_cpp_policies and not create silos. @NicolasFussberger @S-MOHAMD @pawelrutkaq any thoughts on this?

NicolasFussberger:
> My opinion: I would say we should bring in our thoughts into score_cpp_policies and not create silos. @NicolasFussberger @S-MOHAMD @pawelrutkaq any thoughts on this?

If we can bring in these changes into score_cpp_policies, I agree that this should be the preferred solution. I cannot judge how realistic this is.

.bazelrc
```diff
@@ -93,3 +93,41 @@ build:build_qnx8 --config=arm64-qnx
...
+# ==============================================================================
+# Sanitizer configurations — powered by score_cpp_policies runtime infrastructure
+# (wrapper, suppression files, env templates)
+# ==============================================================================
+
+# Minimal debug info (-g1) for sanitizer stack traces; stripping disabled
+build:with_debug_symbols --copt=-g1
+build:with_debug_symbols --strip=never
+
+# AddressSanitizer + UndefinedBehaviorSanitizer + LeakSanitizer (recommended)
```
MaciejKaszynski, line 106: There is a section in the README.md about building with sanitizers, could you update to use these options

MaciejKaszynski (changes requested):

MaciejKaszynski (approved):

.github/workflows/sanitizers.yml
```diff
@@ -0,0 +1,60 @@
...
+        uses: bazel-contrib/setup-bazel@0.18.0
+        with:
+          bazelisk-cache: true
+          disk-cache: ${{ github.workflow }}-${{ matrix.config }}
+          repository-cache: true
+          cache-save: ${{ github.event_name == 'push' }}
+
+      - name: Configure kernel for TSan (ASLR breaks TSan shadow memory layout)
+        if: matrix.config == 'tsan'
+        run: sudo sysctl -w kernel.randomize_va_space=0
```
NicolasFussberger: What is the problem that shows up with ASLR?
Locally I am not encountering a problem with ASLR enabled.

There is also this ticket https://github.com/llvm/llvm-project/issues/164989 which suggests
> build and link our sanitizer tests with -no-pie

I am asking because I wonder if there is a security concern with disabling ASLR here.

VukPavicRTRK: The failure mode is: `FATAL: ThreadSanitizer: unexpected memory mapping`.

TSan pre-reserves large shadow memory regions at fixed virtual addresses. The shadow memory is approximately 8× the size of the traced memory. With full ASLR (randomize_va_space=2, the default), the kernel may place shared libraries or anonymous mmap allocations into those fixed ranges, causing TSan to abort.

This conflict is intermittent and load-dependent, which explains why it does not always reproduce locally but consistently appears if I use `--runs_per_test=10`.

Conservative ASLR (randomize_va_space=1) randomizes the stack but leaves mmap randomization enabled. I tested this setting and still observed the same `unexpected memory mapping` failure. Since the conflict originates from mmap address randomization, =1 does not resolve the issue. Only randomize_va_space=0 reliably prevents the conflict.

For the `-no-pie / -fno-pie`, our failure is different: TSan aborts when mmap randomization places shared libraries into TSan's reserved shadow-memory range. As a result, `-no-pie` does not address the root cause.

Additionally, Bazel's --linkopt applies to all link actions, including shared-library (.so) targets. Applying -no-pie globally therefore breaks shared-library builds with: `undefined reference to main`

NicolasFussberger:
What do you think about this asan finding in//score/health_monitor/src/cpp:cpp_unit_tests?
Is this a valid finding or could it be because of the interaction with rust code?
It reminds me of this open issue: https://github.com/eclipse-score/lifecycle/issues/175

VukPavicRTRK:
> What do you think about this asan finding in//score/health_monitor/src/cpp:cpp_unit_tests? Is this a valid finding or could it be because of the interaction with rust code? It reminds me of this open issue: #175

This is a valid finding, not an FFI false positive. The bug is squarely on the C++ side:

```cpp
if (context_)
{
    auto value{context_.value()};      // std::string, scoped to this block
    context_ptr = value.c_str();       // pointer into value's buffer escapes the block
    context_size = value.size();
}                                      // value destroyed -> context_ptr now dangles
...
set_default_logger(context_ptr, context_size, ...);  // reads dangling pointer
````

`value` is destroyed at the closing brace of the `if`, but `context_ptr`/`context_size` survive and get passed to the Rust FFI call afterward. Rust's `to_vec`/`spec_to_string` then performs a `memcpy` through that dangling pointer. ASan correctly flags the C++ frame (`SetAsDefaultLogger`, line 64/69) as the origin, not the Rust side.

I think there's a real chance these issues are related. The report describes the failure as memory-layout/timing-dependent ("sporadic"), which is exactly the signature you'd expect from undefined behavior.

If `set_default_logger` receives a context pointer/length pair pointing at freed stack memory, the logger could be initialized with garbage context data. Depending on what happens to occupy that stack slot at the time, this could plausibly produce intermittent I/O, buffering, or formatting failures like the one observed in #175 rather than crashing deterministically.

I wouldn't claim this is the root cause of #175 without reproducing the issue after applying a fix, but it's a strong lead worth linking. A dangling-pointer bug during logger initialization is a very plausible source of sporadic downstream failures in the same logging path.

I'd recommend mentioning this in any upstream report filed against `score_baselibs_rust` and cross-linking it to #175 so whoever investigates that issue has this thread to pull on.

NicolasFussberger:
> > What do you think about this asan finding in//score/health_monitor/src/cpp:cpp_unit_tests? Is this a valid finding or could it be because of the interaction with rust code? It reminds me of this open issue: #175
> 
> This is a valid finding, not an FFI false positive. The bug is squarely on the C++ side:
> 
> ```c++
> if (context_)
> {
>     auto value{context_.value()};      // std::string, scoped to this block
>     context_ptr = value.c_str();       // pointer into value's buffer escapes the block
>     context_size = value.size();
> }                                      // value destroyed -> context_ptr now dangles
> ...
> set_default_logger(context_ptr, context_size, ...);  // reads dangling pointer
> ```
> 
> `value` is destroyed at the closing brace of the `if`, but `context_ptr`/`context_size` survive and get passed to the Rust FFI call afterward. Rust's `to_vec`/`spec_to_string` then performs a `memcpy` through that dangling pointer. ASan correctly flags the C++ frame (`SetAsDefaultLogger`, line 64/69) as the origin, not the Rust side.
> 
> I think there's a real chance these issues are related. The report describes the failure as memory-layout/timing-dependent ("sporadic"), which is exactly the signature you'd expect from undefined behavior.
> 
> If `set_default_logger` receives a context pointer/length pair pointing at freed stack memory, the logger could be initialized with garbage context data. Depending on what happens to occupy that stack slot at the time, this could plausibly produce intermittent I/O, buffering, or formatting failures like the one observed in #175 rather than crashing deterministically.
> 
> I wouldn't claim this is the root cause of #175 without reproducing the issue after applying a fix, but it's a strong lead worth linking. A dangling-pointer bug during logger initialization is a very plausible source of sporadic downstream failures in the same logging path.
> 
> I'd recommend mentioning this in any upstream report filed against `score_baselibs_rust` and cross-linking it to #175 so whoever investigates that issue has this thread to pull on.

I linked your comment in #175.

Since you found and analysed the problem, would you like to report this directly to baselibs team?
I think creating the bug ticket here https://github.com/orgs/eclipse-score/projects/24/views/4 and posting in the score-baselibs slack channel is probably a good approach. They will probably also appreciate if you already have a fix in mind.

Should be disable asan for the failing test on our side for now, mentioning the bug ticket? Otherwise, we might be tempted to ignore the sanitizer pipeline output, because it will always show us failed.

NicolasFussberger (dismissed):

VukPavicRTRK:
## Summary of Findings and Actions

### TSan Split

TSan integration has been moved to a separate branch (`issue-50-tsan-sanitizer`) and will be tracked as a follow-up to #50.

It was split out because it raises open questions that do not affect ASan/UBSan/LSan and should not block this part from merging.

### `cpp_unit_tests` ASan Failure (eclipse-score/baselibs#253)

The sanitizer CI failure in `//score/health_monitor/src/cpp:cpp_unit_tests` is a valid finding, not an FFI false positive.

`StdoutLoggerBuilder::SetAsDefaultLogger()` in `score_baselibs_rust` passes a dangling pointer to `set_default_logger`: a `std::string` local to an `if` block is destroyed before the Rust FFI call reads from it.

This has been reported upstream as **eclipse-score/baselibs#253**.

The fix is a one-line change: hoist `value` out of the `if` block.

To keep the pipeline actionable while the fix is pending:

* The test has been tagged `no-asan`.
* The tag includes a reference to the upstream issue.

This is also a candidate root cause for the sporadic crash tracked in #175.

### `mpmc_concurrent_queue_test` `no-asan` Tag

The pre-existing `no-asan` tag had no explanation.

The tag was temporarily removed and the test was run under `--config=asan_ubsan_lsan` for verification. UBSan consistently reports a misaligned-reference error at constructor entry (`mpmc_concurrent_queue.hpp:87`) before any queue logic executes.

This confirms that the issue is a fake-stack alignment false positive caused by ASan's fake stack not honoring the `alignas(64)` members.

The tag has been restored with a comment referencing *eclipse-score/lifecycle#241*.

### Other `no-asan` Exclusions Verified as Legitimate

* `health_monitor` Rust tests (`tests`, `loom_tests`)
  Ferrocene does not support ASan instrumentation, so linking fails.

* `mpmc_concurrent_queue_tsan_test`
  Already compiled with `-fsanitize=thread`; combining ASan and TSan is unsupported.

* `integration_test` macro
  The Docker test container does not ship the sanitizer runtime.

NicolasFussberger (approved):

PR #208, opened 2026-05-19
Feature/load new lifecycle config

NicolasFussberger:
* Introduces an interface in src/launch_manager_daemon/config for loading a launch manager configuration file
 * Introduce an implementation of this interface using flatbuffer
 * Introduce unit tests for the new code

Note: The code is not actually in use yet. The python script which does the translation from the user-facing json configuration to the flatbuffer json format is not part of the PR and will be added in a follow up PR.
Furthermore, the launch manager code needs to be adapted to use the new structure.

Background:
* In the last months we had defined a new user-facing configuration format (see src/launch_manager_daemon/config/config_schema/launch_manager.schema.json).
* However, this new format is currently still mapped to the old configuration files (to processes / process groups). So the code is not actually reading the new configuration file

Relates To: https://github.com/eclipse-score/lifecycle/issues/209


MODULE.bazel
```diff
@@ -134,6 +134,13 @@ use_repo(oci, "debian-test-runtime", "debian-test-runtime_linux_amd64")
 bazel_dep(name = "score_baselibs_rust", version = "0.1.2")
 bazel_dep(name = "score_baselibs", version = "0.2.6")
 
+# Temporarily overwrite baselibs to use the new flatbuffer config loader until there is a new release
+git_override(
+    module_name = "score_baselibs",
+    commit = "498a4b256c9073602140243d30c33b106e279f75",
+    remote = "https://github.com/eclipse-score/baselibs.git",
+)
```
NicolasFussberger: @paulquiring Any idea when a new baselibs release will be available containing the new flatbuffer functionality?

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+#include <vector>
+
+namespace score::launch_manager::config
+{
+
+enum class ApplicationType : uint8_t
+{
+    Native = 0,
+    Reporting = 1,
+    ReportingAndSupervised = 2,
```
paulquiring, line 30: Would just Supervised be sufficient? I guess Supervised applications have to be reporting or else they cannot be supervised?

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+    StateManager = 3
+};
+
+enum class ProcessState : uint8_t
+{
+    Running = 0,
+    Terminated = 1
+};
+
+struct ComponentAliveSupervision
```
paulquiring, line 40: All struct members are default initialized to 0. I think to have some default values here which make sense could be helpful. 

This applies also to all other configuration structs.

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+    std::vector<RunTargetConfig> run_targets_;
+    std::string initial_run_target_;
+    FallbackRunTargetConfig fallback_run_target_;
+    AliveSupervisionConfig alive_supervision_;
+    std::optional<WatchdogConfig> watchdog_;
+};
+
+}  // namespace score::launch_manager::config
+
+#endif
```
paulquiring:
```suggestion
#endif // CONFIG_HPP
```

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+
+/// @brief Move-only value object holding the parsed launch-manager configuration.
+///
+/// Loaded once by an IConfigLoader, then individual parts are moved out via
+/// the `take_*()` accessors to transfer ownership to dedicated domain objects
+/// (e.g. ComponentConfig to Component).
+///
+/// @note As of now, everything is expected in a single config file.
+/// Even though some fields appear as optional in the json schema, they are only optional if the configuration would be
+/// split across multiple files. As only a single file is supported, the single config will contain all fields.
```
paulquiring, line 210: "Even though some fields appear as optional in the json schema, they are only optional if the configuration would be split across multiple files"

That is strange, why splitting configuration into more then one file influencing what optional means?

I would expect that everything which is std::optional in config.hpp simply does not need to configured.

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+    }
+    std::string_view initial_run_target() const
+    {
+        return initial_run_target_;
+    }
+    const FallbackRunTargetConfig& fallback_run_target() const
+    {
+        return fallback_run_target_;
+    }
+    const AliveSupervisionConfig& alive_supervision() const
```
paulquiring: it's quite a mix of styles. 
setRunTargets and fallback_run_target, just one will do the job.

src/launch_manager_daemon/config/include/config_loader.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+    /// @brief Loads and parses the configuration file at @p path.
+    /// @param path Filesystem path to the configuration file.
+    /// @return A @ref Config object on success, or an @ref Error on failure.
+    virtual score::cpp::expected<Config, Error> load(const score::filesystem::Path& path) = 0;
+};
+
+}  // namespace score::launch_manager::config
+
+#endif
```
paulquiring:
```suggestion
#endif // CONFIG_LOADER_HPP
```

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+    {
+        return std::move(alive_supervision_);
+    }
+    std::optional<WatchdogConfig> take_watchdog()
+    {
+        return std::move(watchdog_);
+    }
+
+  private:
+    friend class Builder;
```
paulquiring: A comment would have been helpful to mention that this saves the need to write getters in the builder class.

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,252 @@
...
+class Config
```
paulquiring, line 211: It should be also possible to implement the Config member functions in the config.cpp file.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+    return s ? s->str() : std::string{};
+}
+
+std::vector<std::string> convertStringVector(
+    const ::flatbuffers::Vector<::flatbuffers::Offset<::flatbuffers::String>>* vec)
+{
+    std::vector<std::string> result;
+    if (vec != nullptr)
+    {
+        result.reserve(vec->size());
```
paulquiring: I'm happy to see this line.

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+{
+    std::unordered_map<std::string, std::string> result;
+    if (vec != nullptr)
+    {
+        result.reserve(vec->size());
+        for (const auto* ev : *vec)
+        {
+            if (ev != nullptr && ev->key() != nullptr)
+            {
+                result.emplace(ev->key()->str(), safeString(ev->value()));
```
paulquiring: I guess the key cannot be nullptr?

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+    if (sa == nullptr)
+    {
+        return std::nullopt;
+    }
+    return SwitchRunTargetAction{safeString(sa->run_target())};
+}
+
+SwitchRunTargetAction convertRequiredSwitchRunTargetAction(const fb::SwitchRunTargetAction* sa)
+{
+    return convertSwitchRunTargetAction(sa).value();
```
paulquiring: value can throw.
Image

.value_or(std::string{}) as alternative?

MODULE.bazel
paulquiring: Next release of baselibs: planed on 21.05 or 22.05. See https://github.com/orgs/eclipse-score/discussions/2390?sort=new#discussioncomment-16987333

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+namespace fb = score::launch_manager::config::fb;
+
+namespace
+{
+
+constexpr int32_t kExpectedSchemaVersion = 1;
+constexpr double kSecondsToMilliseconds = 1000.0;
+
+uint32_t secondsToMs(double seconds)
```
paulquiring:
```c++
#include <iostream>
#include <cstdint>
#include <string_view>
#include <cassert>

constexpr double kSecondsToMilliseconds = 1000.0;

uint32_t secondsToMs(double seconds)
{
    return static_cast<uint32_t>(seconds * kSecondsToMilliseconds);
}

int main () {
    assert(secondsToMs(0.000500) == 0);// holds is this a problem ?
    assert(secondsToMs(0.001000) == 1);
}
```

if this is a problem we might need to change the [fbs](
https://github.com/eclipse-score/lifecycle/pull/208/changes#diff-1efc43cb4599be9a98dc6dfbf42adbf370d087bf42ec0fbc037feda8f5cabb30R23
)

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+{
+    return static_cast<uint32_t>(seconds * kSecondsToMilliseconds);
```
SimonKozik: Could this result in UB if seconds < 0 or > UINT32_MAX?

Schema validator should take care about negative input, but static analysis or compiler warnings may not know about this. However we don't have cap on max timeout in schema.

Maybe we can log warning if value is too small or too big and apply some boundary values

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,398 @@
...
+
+    if (config->schema_version() != kExpectedSchemaVersion)
+    {
+        return score::cpp::make_unexpected(IConfigLoader::Error::UnsupportedVersion);
+    }
+
+    Config::Builder builder;
+
+    // initial_run_target is a required field, guaranteed non-null by the schema and verifier.
+    builder.setInitialRunTarget(config->initial_run_target()->str());
```
SimonKozik, line 89: I have mixed feelings about defensive coding, but if somebody prepares crazy config (bypassing verifier), we will crash instead of printing error message and exiting.

However if somebody is doing this, then maybe, they should not expect much support from launch manager.

score/launch_manager/daemon/src/configuration/details/new_lm_flatcfg.fbs
```diff
@@ -0,0 +1,185 @@
+// FlatBuffers schema for S-CORE Launch Manager configuration.
```
paulquiring, line 1: It looks like this will meet the configuration requirements of lifecycle. What I miss compared to the current fbs, are file_identifier and file_extension. Any reason to leave that out?

paulquiring:
review from my side is finished.

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: Thats true. However, the user-facing config has [defined those terms](https://github.com/eclipse-score/lifecycle/blob/main/docs/module/launch_manager_deamon/user_guide/configuration.rst)

```
application_profile (object, optional)

Description: Defines the application profile that specifies the runtime behavior and capabilities of this component, particularly concerning its interaction with the Launch Manager.

Allowed Values:
* "Native": Indicates no integration with the Launch Manager.* * "Reporting": Implies the component uses Launch Manager lifecycle APIs for basic reporting.
* "Reporting_And_Supervised": Implies the component uses lifecycle APIs and sends alive notifications to the Launch Manager.
* "State_Manager": Implies the component uses lifecycle APIs, sends alive notifications, and has permission to change the active Run Target.
```

I would say it makes sense stay consistent to those terms to avoid confusion.

Since the reporting of alive notifications and reporting of Running state are different APIs, to outsiders it may seem plausible that supervision would work without without reporting a Running state. That is why we went for this verbose naming

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: I understand the idea.
Currently all the defaults are handled in the python script that does the translation. There may be user-defined defaults in the json config and there are the ["score defaults"](https://github.com/eclipse-score/lifecycle/blob/056699849d4142a20852d929babf4f60d9d33a5a/scripts/config_mapping/lifecycle_config.py#L9) in case something is not user-defined. The resulting flatbuffer file will then have all fields fully defined (either because the user defined a default, explicitly defined a value or because the global score defaults were applied)

The idea is that after the translation with the python script, all fields are filled. So that in the code we do not need to handle default values.

So I am wondering if we provide any default values here, they would not be used any would be duplicated from the score_defaults definition in the python script

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: Yes everything that is std::optional here is truly optional.

There is a conundrum with the current user-facing json configuration in that it also contains launch-manager-daemon-specific configuration that must only exist once (even once we support multiple config files) as well as the component/runTarget configuration. Since we only support a single config file at the moment, these fields are actually mandatory as of now.

Example:
"initial_run_target" must be available. It tells the launch_manager which RunTarget to start initially. However, this information must only exist within a single config file (once we have multiple config files). Therefore it is optional in the json schema, as not all config files will contain it.

I believe we should change this design and move launch_manager specific configuration to CLI flags (and having a default in case not specified).

E.g. launch_manager -i <initial_run_target>

This would avoid this situation once we start supporting multiple config files.

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Looking at the rest of the code I think CamelCase is the way to go. I'll change it to CamelCase.

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: I guess this is not an issue.

The user is configuring all [time values in seconds](https://github.com/eclipse-score/lifecycle/blob/main/docs/module/launch_manager_deamon/user_guide/configuration.rst#time-intervals)

Though we should reconsider whether the want to switch the configuration to ms.

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: In the flatbuffer schema I have defined it as

```
// A key-value pair representing an environment variable.
table EnvironmentalVariable {
  key:string (required);
  value:string (required);
}
```

So I think if an environment variable is there, the key and value must be available.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: It should never happen because the SwitchRunTargetRecoveryAction is defined as required in this context:

```
// Configuration for a Run Target representing an operational mode of the system.
table RunTarget {
 ....
  // Recovery action when a component in this Run Target fails.
  recovery_action:SwitchRunTargetAction (required);
}
```

However to be defensive against potential future changes on the schema, we could:

* Add a try/catch within parseFlatbuffer and return `score::cpp::make_unexpected(IConfigLoader::Error::InvalidFormat);` in case there is any exception.
* And/or put in an assertion
```
    auto opt=convertSwitchRunTargetAction(sa)
    assert(opt.has_value() && "SwitchRunTargetAction is null although it is expected to be a mandatory configuration");
    return opt.value();
```

What do you think?

score/launch_manager/daemon/src/configuration/details/new_lm_flatcfg.fbs
NicolasFussberger:
```
file_identifier "BLMR";
file_extension "bin";
```

I think this leftover from the codes history.
Originally, the flatbuffer files were picked up via file naming convention and only the folder was configured.
E.g. the central config loader wrapper was looking for <folder path>/lmr_**.bin to identify the correct config files to load.

However, within score this is no longer the case. The central config loader does not exist anymore.
In the past it was mentioned that eventually the files should be specific as CLI.

E.g. `launch_manager -c /.../myconfig1.bin -c /../myconfig2.bin`

Therefore, I think we can safely omit those values.

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/d523966489854093388a8abdb29dd453f1fd6523

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/eb8a04453986fc019db9edc50cea3dc6d8588870

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/24d6e5fafa2582d869a510205b2c9cd2c913bfab

src/launch_manager_daemon/config/include/config_loader.hpp
NicolasFussberger: Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/24d6e5fafa2582d869a510205b2c9cd2c913bfab

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: Discussed in meeting

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: discussed in meeting

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: Consider adding asserts for values that are never expected to be nullptr.
safestring can be removed for value() because it is required in the schema

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: Discussed: We shall add asserts

Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/99cdb9c55dfb8bcc73ad2a82a48533c134263976

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: How to safeguard that the user does not configure anything < 1ms?

NicolasFussberger: Adding an assertion for now, to be moved to schema validation later

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Discussed in meeting. The friend is needed to access the private constructor of the Config class.
This is in line with typical builder pattern implementation

NicolasFussberger: Moving the Builder class to the outer scope so that it is not a nested class anymore.
Nested class is not required and may be discouraged by some SCA rules.

Done in https://github.com/eclipse-score/lifecycle/pull/208/changes/f70d64bc5b687fc07146a27787dcc7bddb49ea72

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: Asserts added in https://github.com/eclipse-score/lifecycle/pull/208/changes/99cdb9c55dfb8bcc73ad2a82a48533c134263976

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: Assertion added in https://github.com/eclipse-score/lifecycle/pull/208/changes/a1f2860bb959b7f23a7d8a955c410277c72d757a

Assertion fails if configured seconds > 0 but is rounded down to 0ms.

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,218 @@
...
+};
+
+struct SwitchRunTargetAction
+{
+    std::string run_target;
+};
+
+struct Sandbox
+{
+    uint32_t uid{};
```
paulquiring: uid_t
```suggestion
    uid_t uid{};
```

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,218 @@
...
+    uint32_t gid{};
```
paulquiring:
```suggestion
    gid_t gid{};
```

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,218 @@
...
+    int32_t scheduling_priority{};
+    std::optional<uint64_t> max_memory_usage;
+    std::optional<uint32_t> max_cpu_usage;
+};
+
+struct DeploymentConfig
+{
+    uint32_t ready_timeout_ms{};
+    uint32_t shutdown_timeout_ms{};
+    std::unordered_map<std::string, std::string> environmental_variables;
```
paulquiring: from:  https://linux.die.net/man/3/execve
we need a format like this:

```c++
char *env[] = { "HOME=/usr/home", "LOGNAME=home", (char *)0 };
```

this will do the job

```c++
std::vector<const char*> environmental_variables = {"HOME=/usr/home", "LOGNAME=home" , nullptr };
```

```c++
execve(... , const_cast<char* const*>(environmental_variables .data()));
```

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,218 @@
...
+struct SwitchRunTargetAction
+{
+    std::string run_target;
+};
+
+struct Sandbox
+{
+    uint32_t uid{};
+    uint32_t gid{};
+    std::vector<uint32_t> supplementary_group_ids;
```
paulquiring:
```suggestion
    std::vector<gid_t > supplementary_group_ids;
```

paulquiring (changes requested):

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: I had similar discussion with Paul. For fields that are marked as required in the flatbuffer schema, flatbuffer will ensure that this is not nullptr. 

Nevertheless, we decided for adding additional asserts to check that those are not nullptr.
Adding further nullptr checks seemed redundant and would make covering this code via tests almost impossible.

Is that approach fine for you?

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: With `std::vector<const char*>` I think there is the question where do the strings live in case they are not constant char sequences. We would require a combination of 
```c++
    std::vector<const char*> pointers_;
    std::vector<std::string> strings_;
```

NicolasFussberger: Please tell me if this to too overcomplicated. But I see the use case that we might want to add an environment variable before starting the process that was not configured by the user. We had this use case already in the past.

This could be achieved rather cleanly with a dedicated class, that stores both `std::vector<const char*>` and `std::vector<std::string>`

```c++
class EnvironmentVariable
{
  public:
    EnvironmentVariable(std::string_view key, std::string_view value);

    std::string_view key() const;
    std::string_view value() const;
    const char* c_str() const;

  private:
    std::string entry_; // internal storage in "key=value" format.
    std::size_t key_length_;
};

class Environment
{
  public:
    using const_iterator = std::vector<EnvironmentVariable>::const_iterator;

    Environment() = default;

    Environment(const Environment&) = delete;
    Environment& operator=(const Environment&) = delete;
    Environment(Environment&& other) noexcept;
    Environment& operator=(Environment&& other) noexcept;

    ~Environment() = default;

    void reserve(std::size_t count);
    
    void add(std::string_view key, std::string_view value);

    const_iterator begin() const;
    const_iterator end() const;
    std::size_t size() const;

    // Access format expected by execve systemcall
    const std::vector<const char*>& envp() const;

  private:
    void rebuildPointers();
    std::vector<EnvironmentVariable> entries_;
    std::vector<const char*> pointers_;
};
```

The alternative that I see is to keep doing what we do know: 
`char* envp_[static_cast<std::size_t>(score::lcm::internal::kEnvArraySize)];`

However, than we have fixed size and need to deal with the scenario this size may be exceeded by the configuration. Still there is the problem of string ownership for which I believe we will need to wrap it in a dedicated class.

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: done here https://github.com/eclipse-score/lifecycle/pull/208/changes/efe69a05ba19a0f83685c91fb466b1fca022b57e

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: done here https://github.com/eclipse-score/lifecycle/pull/208/changes/efe69a05ba19a0f83685c91fb466b1fca022b57e

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: done here https://github.com/eclipse-score/lifecycle/pull/208/changes/efe69a05ba19a0f83685c91fb466b1fca022b57e

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,219 @@
...
+    std::string run_target;
+};
+
+struct Sandbox
+{
+    uid_t uid{};
+    gid_t gid{};
+    std::vector<gid_t> supplementary_group_ids;
+    std::optional<std::string> security_policy;
+    std::string scheduling_policy;
```
NicolasFussberger: Do we want to translate scheduling_policy already to an integer?

I think for logging it might be nice to have the string available, so you can log "SCHED_OTHER" instead of "scheduling policy: 1".

Though I don't have a string opinion. Storing the same string over and over again is also not nice.

Probably makes sense to already translate it to int

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Implemented in https://github.com/eclipse-score/lifecycle/pull/208/changes/5c25b13c5673ee5d3e1a232f70ca6792b6061172

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Moved to int in https://github.com/eclipse-score/lifecycle/pull/208/changes/dbfa690c6a1c6388877b1dae9f90f9d9c7365216

MODULE.bazel
NicolasFussberger: Baselibs release is out.
I updated the MODULE.bazel file to take the new release.

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: I added further validation of the seconds conversion here https://github.com/eclipse-score/lifecycle/pull/208/changes/dd96262ac9197f78bd5e02ea6bed35c651ae997e

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,261 @@
...
+{
+  public:
+    Config(const Config&) = delete;
+    Config& operator=(const Config&) = delete;
+    Config(Config&&) = default;
+    Config& operator=(Config&&) = default;
+
+    // Read access
+    const std::vector<ComponentConfig>& components() const;
+    const std::vector<RunTargetConfig>& runTargets() const;
```
MaciejKaszynski, line 222: Can we use an `score::cpp:span` here

src/launch_manager_daemon/config/include/flatbuffer_config_loader.hpp
```diff
@@ -0,0 +1,71 @@
...
+#include <vector>
+
+namespace score::launch_manager::config
+{
+
+/// @brief Internal helpers for FlatBuffer config parsing.
+namespace details
+{
+
+score::cpp::expected<Config, IConfigLoader::Error> parseFlatbuffer(const std::vector<uint8_t>& buffer);
```
MaciejKaszynski: should we use [[nodiscard]] on Result or expected. Alternatively we can se the IConfigLoader::Error to [[nodiscard]]

src/launch_manager_daemon/config/include/config_loader.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+/// @brief Abstract interface for loading Launch Manager configuration from a file.
+///
+/// Decouples the configuration consumer from the serialization format (e.g., JSON, FlatBuffers).
+/// A concrete implementation parses the file and returns a format-independent @ref Config object.
+class IConfigLoader
+{
+  public:
+    /// @brief Error codes returned when configuration loading fails.
+    enum class Error : std::uint32_t
```
MaciejKaszynski: Would it make sense to make just a ConfigError enum and not tie it into IConfigLoader

src/launch_manager_daemon/config/include/config.hpp
```diff
@@ -0,0 +1,261 @@
...
+#define CONFIG_HPP
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <string_view>
+#include <sys/types.h>
+#include <vector>
+
+namespace score::launch_manager::config
```
MaciejKaszynski: Think we should start using new namespaces

score/launch_manager/daemon/src/configuration/config.hpp
MaciejKaszynski, line 1: This needs more docstrings on the methods

src/launch_manager_daemon/config/include/config_loader.hpp
```diff
@@ -0,0 +1,53 @@
...
+    {
+        FileNotFound,            ///< The configuration file does not exist at the given path.
```
MaciejKaszynski: I'm not a fan of the ///< comments, I think we should use the normal @brief above each value

score/launch_manager/daemon/src/configuration/config.hpp
```diff
@@ -0,0 +1,261 @@
...
+    const_iterator begin() const;
+    const_iterator end() const;
+    std::size_t size() const;
+
+    char* const* envp() const;
+
+  private:
+    void rebuildPointers() const;
+    std::vector<EnvironmentVariable> entries_;
+    mutable std::vector<const char*> pointers_;
```
MaciejKaszynski, line 124: Since we only ever will need to get the char* const* of the env vars would it make sense to just figure out the pointers in the envp() method and not keep track of each pointer?

score/launch_manager/daemon/src/configuration/lm_flatcfg_generated.h
MaciejKaszynski, line 1: should this be removed since you added the generate_cpp bazel rule?

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,697 @@
...
+
+namespace score::launch_manager::config
+{
+
+namespace fb = score::launch_manager::config::fb;
+
+namespace
+{
+
+constexpr int32_t kExpectedSchemaVersion = 1;
```
MaciejKaszynski: Maybe we should put this as a FlatbufferConfigLoaderImpl member

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,697 @@
...
+            return ApplicationType::ReportingAndSupervised;
+        case fb::ApplicationType::State_Manager:
+            return ApplicationType::StateManager;
+        case fb::ApplicationType::Native:
+        default:
+            return ApplicationType::Native;
+    }
+}
+
+ProcessState convertProcessState(fb::ProcessState fb_state)
```
MaciejKaszynski, line 99: Would it make sense to use something like this https://godbolt.org/z/TxqEW7dcE?
Don't really like how you need to hunt for the convertXYZ method, but up to you.

MaciejKaszynski (changes requested):

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: For this const ref version it is possible. However, for the other non-const API where the data ownership is moved span will not be possible.
So I am wondering if it really makes sense to have the const ref and non-const API with a different data type. I am thinking it might be more consistent to stay with the same datatype

src/launch_manager_daemon/config/include/config_loader.hpp
NicolasFussberger: I think this config loading will be the only place where this enum is used as this will be the only config loading code.
We can still move it out of the class and make if `ConfigError` though.
Do you see any benefit it moving it out of the class scope?

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: The pointer returned by envp() is already assembled within the envp() method implementation and not updated whenever an environment variable is added. Most likely envp() is called a single time to get the pointer for passing it into execve system call.
Probably I do not understand your comment correctly.

score/launch_manager/daemon/src/configuration/lm_flatcfg_generated.h
NicolasFussberger: This is the flatbuffer file for the legacy flatbuffer config where the generated files are currently checked in. I did not change this for the legacy configs because they will be removed in the next step anyway.

src/launch_manager_daemon/config/include/flatbuffer_config_loader.hpp
NicolasFussberger: Added [[nodiscard]] on the methods and Error enum here https://github.com/eclipse-score/lifecycle/pull/208/changes/06d37941a85a70f2a6ab1b4cfb5d12f1e3c47c0c

For the Result or expected type I am not sure how we could make that nodiscard since its not defined in our codebase. But I guess having it on the public methods is probably enough.

src/launch_manager_daemon/config/include/config.hpp
NicolasFussberger: Used new namespace here https://github.com/eclipse-score/lifecycle/pull/208/changes/81712b346e4e649c32d36bdcebe30c6c6e6a6079

I suppose the new namespace is score::mw::launch_manager::configuration

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: Added further documentation on methods https://github.com/eclipse-score/lifecycle/pull/208/changes/0e6f0d82ef54831becd5326bea3c1f98c6cfd122

src/launch_manager_daemon/config/include/config_loader.hpp
NicolasFussberger: Migrated to @brief https://github.com/eclipse-score/lifecycle/pull/208/changes/0e6f0d82ef54831becd5326bea3c1f98c6cfd122

src/launch_manager_daemon/config/src/flatbuffer_config_loader.cpp
NicolasFussberger: good idea. Moved this to FlatbufferConfigLoaderImpl and also make use of this in tests to avoid duplication https://github.com/eclipse-score/lifecycle/pull/208/changes/71a4f937a8a0adcdb8a409fef2cf2a60d0354876

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: I guess its a matter of preference. Whether you search based on the type or the name.

I tend to prefer the free functions because the name of the function makes clear what is being converted. 
For example vector gids is `:flatbuffers::Vector<uint32_t>*` which based on the type is not really obvious.

Then there is also the fact that there are a variety of return types, some return the type directly others return expected<...> if conversion may potentially fail. Although also possible with the templates, it looks a bit more straightforward with the regular functions.

MaciejKaszynski: The idea would be then that you that it's easy to separate each config type and write a converter for each type and the type definition in a single header. e.g. we have a config/details/environment.hpp that holds the Environment class definition and convert<fbsEnvironment, Environment>. In the end we can just include and use the function making it more readable.
I guess we can do the same with the convertXYZ methods?

Guess the root problem is that `parseFlatbuffer` is 100 lines which is pretty long for a function.
Right now I feel like this is similar to the old ConfigManager in that there is a huge piece of code that does all the converting and adding might be a pain.

score/launch_manager/daemon/src/configuration/config.hpp
MaciejKaszynski: ah I see this is needed for the backing data when you return the const char* in envp.  Guess we could also just construct the vector in envp() and return the vector so that we don't need this member and don't have to worry about reconstructing it after a move?

score/launch_manager/daemon/src/configuration/config.hpp
MaciejKaszynski: Yeah the return type should signal to the user who owns the data. span and vector are both iterators so don't think a user should have a problem with using either. Then we get the benefit of not having to allocate a vector. But either is fine I guess.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
MaciejKaszynski: Or maybe even a static method on the type struct? e.g. Environment::FromFbs()

src/control_client_lib/include/score/lcm/control_client.h
```diff
@@ -60,6 +60,11 @@ class ControlClient final {
     /// @returns the new reference
     ControlClient& operator=(ControlClient&& rval) noexcept;
 
+
+    void OnRunTargetTransitionFinished(std::function<void(std::string_view)> callback) noexcept;
+    void OnFallbackRunTargetActivated(std::function<void(std::string_view) +  add error information> callback) noexcept;
```
TimoSteuerwaldETAS: This won't build. 🤓

src/control_client_lib/include/score/lcm/control_client.h
```diff
@@ -60,6 +60,11 @@ class ControlClient final {
...
+    std::string_view GetCurrentRunTarget() const noexcept; // when does the "current" RunTarget change? Is it when the request is sent, or when the transition is finished?
```
TimoSteuerwaldETAS: Is the current commit just used as some kind of discussion basis? If not, proper doxygen documentation is missing.

TimoSteuerwaldETAS (changes requested):

src/control_client_lib/include/score/lcm/control_client.h
NicolasFussberger: to be reverted. I should not make changes to the code for discussion purposes

src/control_client_lib/include/score/lcm/control_client.h
NicolasFussberger: fixed https://github.com/eclipse-score/lifecycle/pull/208/changes/8cd3a4b75f7fb640077e05dedadc01846638a9e0

src/control_client_lib/include/score/lcm/control_client.h
NicolasFussberger: fixed https://github.com/eclipse-score/lifecycle/pull/208/changes/8cd3a4b75f7fb640077e05dedadc01846638a9e0

score/launch_manager/daemon/src/configuration/config.hpp
NicolasFussberger: The ides was to have it directly in the format required to be used with envp() systemcall.
If we return a vector the user of this will need to do the const_cast and .data() to bring it in the right format which seems a bit inconvenient.
The vector is build directly when calling envp() so I guess it will always the valid pointers, even if the object got moved before.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: How would having the templated convert methods simplify `parseFlatbuffer`?. Would it not just be calling convert<A, B>(...) instead of a free function for each conversion?
How about further dividing up into methods, like e.g. `parseComponents` and `parseRunTargets`?

I don't really like having the static methods as part of the types because I think the types should be independent of flatbuffer. If we were to change from flatbuffer to json at some point, all the types could stay unchanged.
The only change would be to implement a JsonConfigLoader and instantiate the JsonConfigLoader implementation in main instead of the FlatbufferConfigLoader.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
```diff
@@ -0,0 +1,696 @@
...
+            {
+                return score::cpp::make_unexpected(ready_cond.error());
+            }
+            result.ready_condition = std::move(*ready_cond);
+        }
+    }
+    return result;
+}
+
+score::cpp::expected<Sandbox, IConfigLoader::Error> convertSandbox(const fb::Sandbox* fb_sb)
```
TimoSteuerwaldETAS, line 341: This method is a little bit long. How about a small common helper function for uid and gid? The checks are looking identical anyway.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader_UT.cpp
```diff
@@ -0,0 +1,1353 @@
...
+}
+
+TEST_F(FlatbufferConfigLoaderTest, MissingProcessStateReturnsInvalidFormat)
+{
+    RecordProperty("Description", "ReadyCondition without process_state returns InvalidFormat.");
+
+    ::flatbuffers::FlatBufferBuilder fbb;
+    auto app_profile = fb::CreateApplicationProfile(fbb, fb::ApplicationType::Native, false);
+    auto bin_name = fbb.CreateString("bin");
+    auto ready_cond = fb::CreateReadyCondition(fbb);
```
TimoSteuerwaldETAS: All these type of errors boils down to the same error code (`IConfigLoader::Error::InvalidFormat`), which make it hard to find the error without proper logging.

May be this is a stupid question, but if it does not make sense to create a ready condition without process_state, then we should make process_state mandatory in the schema, or?

Similar for other attributes.

TimoSteuerwaldETAS (changes requested):

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader_UT.cpp
NicolasFussberger: See the comments in src/launch_manager_daemon/config/src/new_lm_flatcfg.fbs

Example:
```
table ReadyCondition {
  process_state:ProcessState = null; // required
}
```
process_state is defined in the flatbuffer schema as optional (even though it is actually required) so that we can verify during the loading that it was actually configured.
If it was defined as mandatory in the flatbuffer schema (i.e. remove =null), then flatbuffer will just return a default value if not configured (in this case Running). 

Therefore, if it was not configured and we missed to have a default in the python translation we get the default value from flatbuffer and there is no possibility to detect that it was not configured. That is why we then rather defined values as optional in the schema and check during loading that the mandatory values have been configured. All values in the schema have a comment to denote whether they are actually optional or required.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
MaciejKaszynski: Yeah I think we should have a converter or free function to create each type.
So that the flow of the code would reflect more or less the config. In the `parseFlatbuffers` method we would just call the parsing functions on each top-level elements in our config, then those take care of everything underneath.

I think this unit is too big, 700 lines and the UT is 1300 lines.
I think splitting this up into multiple files would be nice, and for me the split I suggested is what came first, I'm ok with other splits.

I suggested templates because if we wanted to support different formats then it's pretty easy as you would just have to implement each converter and the shape would be taken care of by the same code as other config types.

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader_UT.cpp
TimoSteuerwaldETAS: Sounds reasonable. Thanks!

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: I have now introduced converter functions for the remaining top-level types `vector<Component>`, `vector<RunTarget>` and moved them to a new file score/launch_manager/daemon/src/configuration/details/flatbuffer_type_converters.hpp with separate unit test.

score/launch_manager/daemon/src/configuration/config.hpp
MaciejKaszynski: mm I think the way I was thinking would still require a const cast from the user so bad idea, so this is fine.

MaciejKaszynski (approved):

score/launch_manager/daemon/src/configuration/details/flatbuffer_config_loader.cpp
NicolasFussberger: I introduced various helper functions to make the method smaller. Its still rather large because of all the individual values that are part of the sandbox.
If you see any further improvement let me know

TimoSteuerwaldETAS (approved):

score/launch_manager/daemon/src/configuration/config/BUILD
```diff
@@ -10,3 +10,85 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
+
+#load("@score_baselibs//score/flatbuffers/bazel:codegen.bzl", "generate_cpp")
```
NicolasFussberger: to be removed

paulquiring: why is everything commented out in this BUILD file?

NicolasFussberger: Seems this file got in by rebasing after the folder restructure. Removed now in https://github.com/eclipse-score/lifecycle/pull/208/changes/39f9616f7a1c85257f034caf34decd0d240d5dae

NicolasFussberger: Removed in https://github.com/eclipse-score/lifecycle/pull/208/changes/39f9616f7a1c85257f034caf34decd0d240d5dae

score/launch_manager/daemon/src/configuration/details/config.cpp
```diff
@@ -0,0 +1,232 @@
...
+        rebuildPointers();
+    }
+    return *this;
+}
+
+void Environment::reserve(std::size_t count)
+{
+    entries_.reserve(count);
+    // +1 for nullptr terminator
+    pointers_.reserve(count + 1);
```
paulquiring, line 67: I'm happy to see this :).

score/launch_manager/daemon/src/configuration/details/flatbuffer_type_converters.cpp
```diff
@@ -0,0 +1,642 @@
...
+        LM_LOG_ERROR() << field_name << " is required but missing";
+        return score::cpp::make_unexpected(IConfigLoader::Error::InvalidFormat);
+    }
+    return *field;
+}
+
+template <typename TargetT>
+score::cpp::expected<TargetT, IConfigLoader::Error> validateRange(uint32_t value, const char* field_name)
+{
+    if (value < static_cast<uint32_t>(std::numeric_limits<TargetT>::min()) ||
```
paulquiring: This will not work as intended.
https://godbolt.org/z/8qhaG8hEz

paulquiring (changes requested):

paulquiring (approved):

anmittag (approved):

PR #217, opened 2026-05-28
Fix Deprecated score::StringLiteral in applicationcontext.h causes build failures in downstream consumers (#214)

### LINKED ISSUE #214: Deprecated score::StringLiteral in applicationcontext.h causes build failures in downstream consumers
### Description

**Problem**

    score::StringLiteral was deprecated in score_baselibs with message:

"No longer supported. Use std::string, std::string_view, safecpp::zstring_view, or const char*"

   applicationcontext.h still uses it in the constructor declaration (line 41):

    ApplicationContext(const std::int32_t argc, const score::StringLiteral argv[]);

Since application.h #includes applicationcontext.h, any translation unit compiled
with  -Werror,-Wdeprecated-declarations (COMPILER_WARNING_FEATURES) that transitively
includes application.h will fail with a build error.

**Impact**

20 targets in eclipse-score/config_management fail to build, including:

//score/config_management/config_daemon/code/app/details:app
//score/config_management/config_daemon/code/data_model:parameterset_collection
//score/config_management/config_daemon/code/factory:interface
//score/config_management/config_daemon/code/plugins:plugin
//score/config_management/config_daemon/code/json_helper:json_helper
(and 15 more targets)
Suggested Fix

Replace  score::StringLiteral with const char* in applicationcontext.h ---this is a
fully compatible change since StringLiteral is an alias for const char*:

    // Before
    #include "score/memory/string_literal.h"
    ApplicationContext(const std::int32_t argc, const score::StringLiteral argv[]);

    // After
    ApplicationContext(const std::int32_t argc, const char* const argv[]);

**References:**

https://github.com/eclipse-score/lifecycle/blob/main/src/lifecycle_client_lib/include/applicationcontext.h#L41
https://github.com/eclipse-score/baselibs/blob/main/score/memory/string_literal.h

### Analysis results

_No response_

### Solution

_No response_

### Error Occurrence  Rate

None

### How to reproduce

_No response_

### Supporting Information

_No response_

### Classification

Minor

### First Affected Release

not released (main)

### Last Affected Release

not released (main)

### Expected Fixed Release

before release (main)

### Category

- [ ] Safety Relevant
- [ ] Security Relevant

WilliamRoebuck:
- Replaced score::StringLiteral with char*
- Fixed a warning from logging a char*

Fixes #214

### DIFF
--- a/docs/module/lifecycle_client_lib/model/structural_view.uxf
+++ b/docs/module/lifecycle_client_lib/model/structural_view.uxf
@@ -55,7 +55,7 @@ environments</panel_attributes><additional_attributes></additional_attributes></
 operation
 + initialize(const ApplicationContext&amp;): std::int32_t override
 + run(amp::stop_token stop_token): std::int32_t override
-+ AasApplicationContainer(cosnt std::int32_t argc, const char* const* argv, const std::size_t count_expected_applications): noexcept
++ AasApplicationContainer(const std::int32_t argc, const char* const* argv, const std::size_t count_expected_applications): noexcept
 + With&lt;App,...Args&gt;(Args&amp;&amp;.. args) : AasApplicationContainer&amp;
 + Launch() : std::int32_t
 --
--- a/docs/module/lifecycle_client_lib/model/structural_view.uxf
+++ b/docs/module/lifecycle_client_lib/model/structural_view.uxf
@@ -6,7 +6,7 @@ Operation
 + run(amp::stop_token stop_token): std::int32_t virtual = 0</panel_attributes><additional_attributes></additional_attributes></element><element><id>UMLClass</id><coordinates><x>171</x><y>855</y><w>486</w><h>135</h></coordinates><panel_attributes>score::mw::lifecycle::ApplicationContext
 --
 Operations
-+ ApplicationContext(const int32_t argc, const score::StringLiteral argv[])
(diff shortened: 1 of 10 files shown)


src/lifecycle_client_lib/src/aasapplicationcontainer.cpp
```diff
@@ -23,9 +23,8 @@ namespace mw
 namespace lifecycle
 {
 
-
 AasApplicationContainer::AasApplicationContainer(const std::int32_t argc,
-                                                 const score::StringLiteral* argv,
+                                                 const char* const* argv,
```
MaciejKaszynski: nit pick, in some places there is the `char* argv[]`, in some places its `char ** argv` Think we should just use the `[]` way

MaciejKaszynski (changes requested):

src/lifecycle_client_lib/src/aasapplicationcontainer.cpp
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/217/commits/dd70721426ec57fad56fda0fc05c8c322542228a

MaciejKaszynski (approved):

NicolasFussberger (approved):

PR #219, opened 2026-06-01
Add tests for examples

### LINKED ISSUE #218: Make `/examples` Testable with Bazel
### What

We have the `/examples` directory for demonstration to the user.

Right now a reviewer has to keep in mind to make sure the example is working which could cause breaking changes to be merged.
We should make it so that we can test the example is working automatically with `bazel test` to prevent errors.

### Acceptance Criteria (DoD)

* `bazel test //examples/...` shall test that the demo is working correctly.

### How

* A docker image should be defined in the bazel
    * We could reuse the base of the image from our tests maybe?
       https://github.com/eclipse-score/lifecycle/blob/1d257e13114536d2538d147f0e01ca7aa3e6a5d3/tests/utils/environments/x86_64-linux/BUILD#L15-L18
* We should also see how much of the code could be shared with `/tests/integration/smoke` (not sure as the test has gtest APIs so we shall decide)
* The scripts that copies the binaries into the correct location shall be replaced with an `oci_images( ... tars = [ demo_files ])` even if we choose to use the test image as a base this is possible.
* Add the `oci_images(... entrypoint = lcm_binary_path )

VukPavicRTRK:
## Summary

Makes `//examples` testable via `bazel test` so that the demo setup is automatically verified, preventing silent breakage from being merged unnoticed.

## What Changed

- **`examples/integration_test/`** - New integration test package:
  - `test_examples.py` - Starts the launch manager, transitions to `Running` (verifying all example apps are alive), transitions back to `Startup`, and cleanly shuts down via SIGTERM.
  - `lifecycle_demo_test.json` - Launch manager config scoped to the test environment (`/tmp/tests/examples`).
  - `BUILD` - Bazel build file packaging binaries via `pkg_tar` and wiring up the `integration_test` rule.
- **`examples/README.md`** - Documents how to run the new integration tests.
- **Visibility updates** - Extended `//examples:__subpackages__` visibility to `testing_utils`, `environments`, and the x86_64 OCI image target to allow reuse of shared test infrastructure.

## Acceptance Criteria

- `bazel test //examples/... --config=<...>` tests that the demo is working correctly.

## Notes on Approach

The test reuses the existing base image and `integration_test` infra from `tests/utils/environments/x86_64-linux` rather than defining a new one from scratch, as suggested in the issue. Binaries are packaged into a `pkg_tar` archive and deployed to the correct location, replacing the need for manual copy scripts.

The smoke test in `tests/integration/smoke` uses gtest APIs, so sharing test code directly wasn't practical - this test is implemented in Python using the same `setup_test` fixture pattern used elsewhere in the integration test suite.

## How to Run

```
bazel test //examples/... --config=<...>
```

Closes #218


examples/integration_test/BUILD
```diff
@@ -0,0 +1,58 @@
...
+    srcs = [
+        ":examples_etc_files",
+        ":examples_main_files",
+    ],
+)
+
+integration_test(
+    name = "examples",
+    srcs = ["test_examples.py"],
+    tags = ["integration"],
```
MaciejKaszynski: I think we should remove this tag as this test should be treated separately to the tests/integration

examples/README.md
```diff
@@ -29,3 +29,8 @@ There is an interactive mode that walks you through two demo scenarios.
 This mode requires the run.sh script to be executed **in an active tmux** session.
 
 `bazel run //examples:run_examples --config=<...> -- tmux`
+
+## Integration Tests
```
MaciejKaszynski: I think we should separate the language of integration test from this. In score integration tests are the equivalent to binary tests. Maybe just call this a Demo Verification?

MaciejKaszynski (changes requested):

MaciejKaszynski (approved):
Looks good, I re-ran this 300 times to make sure the timing is correct and came across the error from https://github.com/eclipse-score/lifecycle/issues/175 but otherwise works good

examples/demo_verification/test_examples.py
```diff
@@ -0,0 +1,60 @@
...
+@add_test_properties(
+    partially_verifies=[],
+    test_type="interface-test",
+    derivation_technique="explorative-testing",
+)
+def test_examples(target, setup_test, remote_test_dir):
+    """
+    Objective: Verifies the example demo runs end-to-end without crashing.
+
+    Starts the launch manager, transitions to Running (all example apps start), then back to Startup.
```
NicolasFussberger: Would it be possible to extend this to also kill a process and make a supervision fail as in 
https://github.com/eclipse-score/lifecycle/blob/12c2d2f4bf949122632e96c2d44eab0caf310d50/examples/demo.sh#L47
https://github.com/eclipse-score/lifecycle/blob/12c2d2f4bf949122632e96c2d44eab0caf310d50/examples/demo.sh#L60-L62

NicolasFussberger:
Thanks for your contribution @VukPavicRTRK !
I really like that this can now be executed automatically.

I have some ideas for follow-up changes, as now we have some duplication between the old manual example execution and the new automated execution: https://github.com/eclipse-score/lifecycle/issues/226

If you have time, would be awesome if you can take this up.

NicolasFussberger (approved):

PR #220, opened 2026-06-01
Removed local supervision, global supervision and recovery notification

WilliamRoebuck:
Closes #133 
* Removed local, global and recovery notification
* Pass process identifier instead of pg identifier
* Remove deadline and logical leftovers
* Collapse checkpoint supervision into alive supervision
* Simplify the internal config (version updated to 8.1)
* Add Alive_UT, MonitorIfDaemon_UT, exec_error_domain_UT


score/launch_manager/daemon/src/alive_monitor/details/ifexm/ProcessStateReader.cpp
```diff
@@ -15,6 +15,8 @@
 
 #include "score/mw/launch_manager/alive_monitor/details/timers/TimeConversion.hpp"
 
+// LCOV_EXCL_START (slated for removal)
```
NicolasFussberger: Not sure if we should mark the code with LCOV_EXCL_START / LCOV_EXCL_END.

I would prefer removing the coverage check or lowering the threshold.

The coverage calculation is broken and adding unit tests currently might actually lower the coverage as the report does not include all the code.

NicolasFussberger: fixed

NicolasFussberger (approved):

MaciejKaszynski (approved):

PR #225, opened 2026-06-03
Remove hmproc config

WilliamRoebuck:
Previously we would generate a config file for each monitored process. The only value from that actually used was the instance specifier. This is now passed by an environment variable per process.

* Remove hmproc config, pass by env var instead
* Add unit test


NicolasFussberger (changes requested):
After latest merge there is now a new config in examples/demo_verification/lifecycle_demo_test.json which also needs to have the CONFIG_PATH removed.
Otherwise looks good to me.

NicolasFussberger (approved):

PR #228, opened 2026-06-05
Remove manual example demo in favour of demo_verification

### LINKED ISSUE #226: Improvement: Remove manual example test in lifecycle
### What

Predecessor: https://github.com/eclipse-score/lifecycle/issues/218

Within https://github.com/eclipse-score/lifecycle/issues/218 an automatically executable version of the example showcase has been introduced.
The manual version is still in the repo which creates some duplication of files:
* examples/demo.sh mostly identical to examples/demo_verification/test_examples.py
* examples/config/lifecycle_demo.json mostly identical to examples/demo_verification/lifecycle_demo_test.json

Difference:
* examples/demo_verification does not make use of the mw::log configs in examples/config while the manual demo does
* From the logs of examples/demo_verification it is not really clear what steps are executed in this demo

Would be great to make examples/demo_verification the new demo and remove the manual demo.

### Acceptance Criteria (DoD)

The manual demo is removed and fully replaced by what is now examples/demo_verification

### How

The manual demo is removed:
* Remove examples/demo.sh
* Remove examples/run.sh
* Update examples/README.md
* Update examples/BUILD

Extensions to examples/demo_verification:
* Update examples/demo_verification to use mw::log configs
* Make the log output such that it is observable from the log what steps are being performed. This could be some highlighted output of the step that is being executed (E.g. Killing supervised app xy, similar as it was in demo.sh)

VukPavicRTRK:
## Description

Removes the Docker-based manual demo and promotes `examples/demo_verification` as the single demo for the example applications.

### Removed

* `examples/demo.sh`, `examples/run.sh`, `examples/Dockerfile`, and related Docker artifacts
* `examples/config/` (superseded by `demo_verification/`)
* Orphaned targets from `examples/BUILD`

### Extended `examples/demo_verification`

* Added `mw::log` configurations (`logging.json`, `hm_logging.json`, `ecu_logging_config.json`), packaged into `etc/`
* Added step logging to `test_examples.py` to make executed scenarios visible in test output

### Documentation Updates

* Updated `examples/README.md`
* Updated the root `README.md`

Both documents now reflect the new demo structure and reference `examples/demo_verification` as the authoritative example workflow.

Closes #226.


examples/demo_verification/test_examples.py
```diff
@@ -49,6 +49,10 @@ def _lmcontrol(target, lmcontrol_path, run_target):
     assert exit_code == 0, f"lmcontrol {run_target} failed"
 
 
+def _step(description):
+    print(f"\n--- Step: {description} ---")
```
NicolasFussberger, line 53: I wonder if we can make this somehow more prominent in the output. Currently, it is hard to find these due to all the other logs.

Two ideas that come to my mind:

* Currently, the launch_manager logs with Debug log level. I was wondering if we shall decrease this to Info so that we have logs cluttering the output. I tried this shortly by adapting the log json files, but this had no effect. It seems these config files are not being picked up at the moment.
* Making the log colored output
* Adding another newline above/below to separate it from the rest

What do you think?

examples/README.md
```diff
@@ -1,36 +1,25 @@
 # Demo Setup
 
-## Building & Running the demo setup
+## Running the Demo Verification
 
-Execute `bazel run //examples:run_examples --config=<...>`. This will build all dependences and run the run.sh script
+Execute `bazel test //examples/... --config=<...>` to build all dependencies and run the automated demo verification.
```
NicolasFussberger: Maybe we should add `--test_output=all` so that you see the actual test steps. Maybe this is not obvious to everyone

NicolasFussberger (changes requested):
Looks very good!
I only have two small comments

examples/demo_verification/test_examples.py
VukPavicRTRK: I think the main issue is actually pytest's output capturing rather than the formatting of the step markers themselves. The _step() helper uses print(), which pytest captures by default. Unless the test is run with -s, these messages are not streamed to the console and are typically only shown on failures.
Here is the command:
`bazel test //examples/... --config=<...> --test_output=all --test_arg=-s`
I will update the README, and then we can evaluate whether additional formatting is still needed.

NicolasFussberger: You are right, already looks much better though.
I will create a separate ticket to investigate the mw::log config issue (https://github.com/eclipse-score/lifecycle/issues/235)

NicolasFussberger (approved):

PR #229, opened 2026-06-05
Refactor Monitor API to public Alive API

### LINKED ISSUE #149: Rename current "HealthMonitor" API in LaunchManager to "Alive"
As per the [feature architecture](https://eclipse-score.github.io/score/main/features/lifecycle/index.html) the API shall be called "Alive"

NicolasFussberger:
Refactors the Launch Manager "Monitor" interface to the Alive API.

Example usage:

```c++
#include "score/mw/lifecycle/alive.h"

score::mw::lifecycle::Alive alive("instance-name");

while(!shutdownRequested) {
  // Do stuff 
  // ...

  alive.ReportAlive();
}
```

```rust
let alive = alive_rs::Alive::new("my instance").expect("Failed to create alive instance");
alive.report_alive();
```

Public bazel targets:
`//score/launch_manager:alive_cc`
`//score/launch_manager:alive_rust`

Changes:
* Rename Monitor class to Alive
* Remove the possibility to report different checkpoint ids from the public API
* Since this is a public launch manager API, it is moved to new top-level score/launch_manager/alive folder

Open Questions:
* Feature Architecture defines method ["report_health_status"](https://eclipse-score.github.io/score/main/features/lifecycle/architecture/launch_manager.html#logic_arc_int_op__lifecycle__report_health) which seems odd if we are only reporting an alive notification. What was the intention of this interface definition?

Closes: https://github.com/eclipse-score/lifecycle/issues/149


examples/rust_supervised_app/Cargo.toml
MaciejKaszynski, line 1: Would it make sense to add another program to the example with the alive supervision only? Right now the only docs on how to use this is this PRs description. This can be a separate issues also.

score/launch_manager/alive/BUILD
```diff
@@ -0,0 +1,29 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
```
MaciejKaszynski: Does the copyright number matter? should be 2026 but I personally don't care :)

score/launch_manager/alive/src/details/BUILD
```diff
@@ -1,5 +1,5 @@
 # *******************************************************************************
-# Copyright (c) 2026 Contributors to the Eclipse Foundation
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
```
MaciejKaszynski: don't go back
```suggestion
# Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/health_monitor/src/Cargo.toml
```diff
@@ -18,7 +18,7 @@ thread.workspace = true
 score_log.workspace = true
 score_testing_macros.workspace = true
 containers.workspace = true
-monitor_rs = { workspace = true, optional = true }
+alive_rs = { workspace = true, optional = true }
```
MaciejKaszynski, line 21: is `optional` correct here?

MaciejKaszynski (changes requested):

examples/rust_supervised_app/Cargo.toml
NicolasFussberger: Yeah I think that makes sense. I did some check locally, but did not want to extend the PR even further.

I created a separate issue here: https://github.com/eclipse-score/lifecycle/issues/234

score/health_monitor/src/Cargo.toml
NicolasFussberger: I suppose that is optional because there is also a stub implementation for the interface within health_monitor.
But the cargo build is not working for me locally, neither before nor with these changes.

Is the cargo build still in use @pawelrutkaq ?

score/launch_manager/alive/src/details/BUILD
NicolasFussberger: fixed

score/launch_manager/alive/BUILD
NicolasFussberger: fixed

score/launch_manager/src/alive/src/alive.h
```diff
@@ -0,0 +1,82 @@
...
+    /// @brief Destructor of an Alive
+    virtual ~Alive() noexcept;
+
+    /// @brief Reports an alive notification
+    /// @remark Thread safety: This method is NOT thread safe.
+    void ReportAlive() const noexcept;
+
+private:
+    /// @brief Unique pointer to implementation class of Alive
+    std::unique_ptr<AliveImpl> aliveImplPtr;
```
pawelrutkaq, line 72: this pimpl Alive (AliveImpl), does not sound as any benefit really here.

score/launch_manager/src/alive/src/details/AliveImpl.h
```diff
@@ -41,36 +39,36 @@ class MonitorImpl
...
+    AliveImpl& operator=(AliveImpl&&) & noexcept = delete;
 
     /// @brief Destructor of the class
-    virtual ~MonitorImpl() = default;
+    virtual ~AliveImpl() = default;
 
     /// @brief Reports an occurrence of a Checkpoint
     /// @param [in] f_checkpointId   Checkpoint identifier.
-    void ReportCheckpoint(Checkpoint f_checkpointId) const noexcept(true);
+    void ReportCheckpoint(std::uint32_t f_checkpointId) const noexcept(true);
```
pawelrutkaq, line 71: Two things: First i dont see any beenfit having Alive and AliveImpl. Seconds is that Impl provides an extended interface (checkpoint id instead just alive notification),  why?

pawelrutkaq (changes requested):

score/launch_manager/src/alive/src/details/AliveImpl.h
NicolasFussberger: Most of the implementation still deals with checkpoints.
I started removing this from the public API as a first step. The rest of the code is yet to be simplified. I did not want to propagate this change here as it would result in massive amount of changes.
But of course you are right, there numbered checkpoint is already obsolete when there is only an alive notification

score/launch_manager/src/alive/src/alive.h
NicolasFussberger: The benefit of the pimpl is that we can keep the public API minimal.
If I were to remove the AliveImpl class everything that is included there would need to become now part of the public API and those classes are meant to stay internal so that we can freely change them later.

ramceb:
FT-Team meeting: 2026-06-10: The report_health_status api shall also provide the possibility to report "not ok" state to shorten FTTI times in some use-cases.

NicolasFussberger:
> FT-Team meeting: 2026-06-10: The report_health_status api shall also provide the possibility to report "not ok" state to shorten FTTI times in some use-cases.

Okay understood. Then I guess the C++ API would look like this:

```c++
enum class HealthStatus : uint8_t {
  kOk,
  kNotOk
};

class Alive {
  public:
    void ReportHealthStatus(HealthStatus status);
};
```

Though I am also wondering if having a separate method for the failure reporting would be more explicit.
Something like:

```c++
class Alive {
  public:
    void ReportAlive();
    void ReportFailure();
};
```

What do you think @pawelrutkaq @ramceb ?

pawelrutkaq:
For me,no matter if enum or static names. Just lets dont have two interfaces for the same ;)

score/launch_manager/src/alive/src/alive.h
pawelrutkaq: not really in my opinion - public is what is public that is. rest can be private/protected and you change it. pimpl here is just code duplication.

score/launch_manager/src/alive/src/details/AliveImpl.h
pawelrutkaq: so shall we link some ticket for cleanu p ?

score/launch_manager/src/alive/src/alive.h
```diff
@@ -0,0 +1,89 @@
+/********************************************************************************
```
pawelrutkaq, line 1: file name lower case, other upper case ?

score/launch_manager/alive/src/alive.h
```diff
@@ -0,0 +1,89 @@
...
+extern "C"
+{
+#endif
+    void* score_lcm_alive_initialize(const char* instanceSpecifier) noexcept;
+    void score_lcm_alive_deinitialize(void* instance) noexcept;
+    void score_lcm_alive_report_alive(void* instance) noexcept;
+    void score_lcm_alive_report_failure(void* instance) noexcept;
+#ifdef __cplusplus
+}
+#endif
```
pawelrutkaq: this does not need to be part of header or ? can be in cpp ?

score/launch_manager/alive/src/alive.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    if (aliveImplPtr.get() != nullptr)
+    {
+        aliveImplPtr->ReportCheckpoint(kDefaultCheckpointId);
+    }
+}
+
+void Alive::ReportFailure() const noexcept
+{
+    // Not implemented
```
pawelrutkaq: assert

score/launch_manager/src/alive/src/alive.cpp
```diff
@@ -0,0 +1,93 @@
...
+    delete alivePtr;
+}
+
+void score_lcm_alive_report_alive(void* instance) noexcept {
+    static_cast<score::mw::lifecycle::Alive*>(instance)->ReportAlive();
+}
+
+void score_lcm_alive_report_failure(void* instance) noexcept {
+    static_cast<score::mw::lifecycle::Alive*>(instance)->ReportFailure();
+}
```
pawelrutkaq, line 84: missing nullptr checks

pawelrutkaq (changes requested):

score/launch_manager/src/alive/src/alive.h
NicolasFussberger: The other public APIs in launch manager are all with lower case.
So now it is consistent at least to the outside.
We'll have to adapt the private code later

score/launch_manager/src/alive/src/alive.h
NicolasFussberger: I think there are two big benefits here for using pimpl pattern.

### 1. Keeping implementation details private

The AliveImpl header includes the following internal headers:
#include "score/mw/launch_manager/alive_monitor/details/ifappl/DataStructures.hpp"
#include "score/mw/launch_manager/alive_monitor/details/ipc/IpcClient.hpp"
#include "score/mw/launch_manager/alive_monitor/details/logging/PhmLogger.hpp"
which again include other internal headers.

When removing the AliveImpl class, those headers would go to the Alive header and thus be required to be part of the public API (+all their transitive headers).
Those are all implementation details that should stay private, otherwise we risk easily breaking users when we change those classes in the future. Furthermore, users may start relying on these implementation details (e.g. specific shared memory layout as defined in DataStructures.hpp).

### 2. ABI compatibility

Size and layout of Alive class will stay unchanged. This can become important when delivering score stack not as bazel modules but as a set of precompiled binaries.

While both are clear benefits in my opinion I would say that the first benefit is the major one.
I don't see why we would expose all these implementation details as part of the API if we can avoid it.
I would say the cost of code duplication is rather small here, compared to the benefit.

score/launch_manager/src/alive/src/details/AliveImpl.h
NicolasFussberger: I created a ticket here https://github.com/eclipse-score/lifecycle/issues/251

score/launch_manager/alive/src/alive.h
NicolasFussberger: you are right, I removed it from the header. Seems enough to have this in the cpp file.

score/launch_manager/alive/src/alive.cpp
NicolasFussberger: I added an assertion that fails when this method is called.

score/launch_manager/src/alive/src/alive.cpp
NicolasFussberger: Not sure what should be the behavior in case nullptr is passed, this would always be a programming error.
Silently ignoring the nullptr, might hide such programming errors.
I introduced now the baselibs assert macros as nullptr check.

score/launch_manager/alive/src/alive.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+namespace score::mw::lifecycle
+{
+
+Alive::Alive(const std::string_view& instance) noexcept(false) :
+    aliveImplPtr(std::make_unique<AliveImpl>(instance))
+{
+}
+
+Alive::Alive(Alive&& se) noexcept :
```
MaciejKaszynski: I think this move constructor and move assignment can just be `= default`

score/launch_manager/src/alive/src/alive.h
MaciejKaszynski, line 1: Did a grep for Monitor and found the following that might need to be updated:

- ./score/launch_manager/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp:93 - log mentions health monitor
- ./score/launch_manager/alive/src/details/AliveImpl.h:28 - the brief should be updated

MaciejKaszynski (changes requested):

score/launch_manager/alive/src/alive.cpp
NicolasFussberger: done in https://github.com/eclipse-score/lifecycle/pull/229/changes/c7019a4febc77659163d3eb44c3d1d41a1eff9b4

score/launch_manager/src/alive/src/alive.h
NicolasFussberger: done in https://github.com/eclipse-score/lifecycle/pull/229/changes/0fe87604ac8a863f486922f533a1726b255a1e52

MaciejKaszynski (approved):

pawelrutkaq (approved):

PR #231, opened 2026-06-08
Remove obsolete Process Group specific supervision functionality

NicolasFussberger:
# Overview

This PR is part of a sequence of refactoring, in which functionality is being removed that was introduced as part of the initial contribution, but is now obsolete in SCore.

## Background

* In the original implementation, the Alive supervision could be enabled only for certain process group states that the process was running in.
* Furthermore, for Logical and Deadline supervision support there was a complex process state tracking required because a supervision could have been configured across multiple processes.
* There was also process crash handling in which the Alive supervision would expire on process crash, leading to supervision failure notification.

## Changes

The implementation was simplified by removing the functionality that is now obsolete in the Score context:
* In Score, Alive supervision is configured per component. It does not care as part of which RunTarget (i.e. ProcessGroupState) the component runs.
* In Score, Logical and Deadline supervision have already been removed from launch_manager code (they are part of health_monitoring library), so the complex process state tracking in ProcessStateTracker class could be removed. For Alive supervision, this is always fixed to a single process.
* In Score, alive supervision is implemented in the launch_manager daemon which also reacts to process crashes directly. No need for any special logic within supervision for crash detection. The launch_manager already handles process crashes.

Closes: https://github.com/eclipse-score/lifecycle/issues/131


scripts/config_mapping/lifecycle_config.py
```diff
@@ -205,19 +205,11 @@ def get_process_type(application_type):
...
-        states = get_all_process_group_states(run_targets)
-        refProcessGroupStates = []
-        for state in states:
-            refProcessGroupStates.append({"identifier": state})
-        return refProcessGroupStates
+    def is_supervised(application_type):
+        return (
+            application_type == "State_Manager"
+            or application_type == "Reporting_And_Supervised"
+        )
```
WilliamRoebuck: This is a redefinition of is_supervised on line 182 (probably came in from a merge). Should be removed.

WilliamRoebuck (commented):
Looks good, just one nitpick

scripts/config_mapping/lifecycle_config.py
NicolasFussberger: Good catch. I removed the duplicate method here https://github.com/eclipse-score/lifecycle/pull/231/changes/c17fe5fdc6ac558ded3204a8112eecf89b57f4c6

I also noticed when rerunning the mapping tests, that some paths were wrong after the folder restructuring. I fixed it here https://github.com/eclipse-score/lifecycle/pull/231/changes/919f6581cea12a1f8cd142e0424e79c5ea6b7e9a

WilliamRoebuck (approved):

anmittag (approved):

PR #239, opened 2026-06-09
Introducing report_running() API & Cleanup lifecycle_client target

TimoSteuerwaldETAS:
Introduces a low-level `void report_running()` function so applications can report a Running state without requiring usage of the full mw::Lifecycle API.
This was decided in one of the weekly lifecycle meetings and documented in ticket https://github.com/eclipse-score/lifecycle/issues/140

This PR introduces the API.
Follow-up PR will then remove the existing LifecycleClient class from public visibility.

Public targets for using the lifecycle client API:

Implementation Targets:
//score/launch_manager:lifecycle_cc
//score/launch_manager:lifecycle_rust

Mock Targets:
//score/launch_manager:applicationcontext_mock_cc
//score/launch_manager:lifecycle_mock_cc
//score/launch_manager:report_running_mock_cc

Realates To: https://github.com/eclipse-score/lifecycle/issues/140


score/launch_manager/lifecycle_client/src/report_running.cpp
```diff
@@ -0,0 +1,23 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
NicolasFussberger: Maybe new code should use 2026 copyright year

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -182,6 +182,22 @@ cc_library(
...
+    srcs = ["src/report_running.cpp"],
+    hdrs = ["src/report_running.h"],
+    features = COMPILER_WARNING_FEATURES,
+    include_prefix = "score/mw/lifecycle/lifecycle_client",
+    strip_include_prefix = "/score/launch_manager/lifecycle_client/src",
+    visibility = ["//visibility:public"],
+    tags = ["FUSA"],
+    deps = [
+        ":lifecycle_client",
+        "@score_baselibs//score/mw/log",
```
NicolasFussberger: I don't see a direct dependency to mw::log in the report_running impl.

score/launch_manager/lifecycle_client/src/report_running.cpp
```diff
@@ -0,0 +1,23 @@
...
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "score/mw/lifecycle/lifecycle_client/report_running.h"
+
+#include "score/mw/log/logging.h"
```
NicolasFussberger: header to be removed

score/launch_manager/lifecycle_client/src/report_running.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include <score/mw/lifecycle/lifecycle_client.h>
```
NicolasFussberger: Should this be:
```suggestion
#include "score/mw/lifecycle/lifecycle_client.h"
```

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -182,6 +182,22 @@ cc_library(
     ]
 ]
 
+
+cc_library(
+    name = "report_running",
+    srcs = ["src/report_running.cpp"],
+    hdrs = ["src/report_running.h"],
+    features = COMPILER_WARNING_FEATURES,
+    include_prefix = "score/mw/lifecycle/lifecycle_client",
```
NicolasFussberger: Should this be ?
```suggestion
    include_prefix = "score/mw/lifecycle",
```

Consumer should include the file via 
`#include "score/mw/lifecycle/report_running.h"`

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -182,6 +182,22 @@ cc_library(
...
+    strip_include_prefix = "/score/launch_manager/lifecycle_client/src",
+    visibility = ["//visibility:public"],
```
NicolasFussberger: I think this should be restricted visibility

```suggestion
    visibility = ["//score:__subpackages__"],
```
only the targets in score/launch_manager/BUILD are the public targets.

If I remember correctly, the report_running should just become part of the `lifecycle_cc` target

score/launch_manager/lifecycle_client/src/report_running.cpp
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -243,6 +257,23 @@ cc_library(
...
+    deps = [
+        "@googletest//:gtest",
+    ],
+)
+
 filegroup(
```
NicolasFussberger: We should check if this can be removed. I don't see any usage

score/launch_manager/lifecycle_client/BUILD
NicolasFussberger: We should check if this entire thing can be removed. I don't see any usage of it.

NicolasFussberger: I wonder why there is a target that compiles again all these sources that already exist in individual libraries.

Ideally, there are individual private libraries and one target that only lists them as dependencies

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/src/report_running.cpp
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/src/report_running.cpp
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Yes. This was needed in first place to build the rust supervised app, however as the rust supervised app only uses the old internal lifecycle client, I could adapt it accordingly. Now it is no longer needed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: As github can not show me to which line of code this comment belongs, please check yourself if this has been fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Yes, can be removed.

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -32,21 +32,6 @@ cc_library(
 
 cc_library(
```
NicolasFussberger, line 16: This still exposes multiple includes paths.
For client code both include paths are working:

```cpp
#include <score/mw/lifecycle/lifecycle_client/application.h>
#include <score/mw/lifecycle/lifecycle_client/runapplication.h>

#include <score/mw/lifecycle/application.h>
#include <score/mw/lifecycle/runapplication.h>
```

Can we restrict this only to the latter one?

score/launch_manager/daemon/src/alive_monitor/rust/src/monitor.rs
```diff
@@ -15,7 +15,6 @@ use libc::{c_char, c_uint, c_void};
 use std::ffi::CString;
 use std::marker::PhantomData;
 
-#[link(name = "lifecycle_client")]
```
NicolasFussberger, line 18: should this be corrected instead of being removed completely?

score/launch_manager/lifecycle_client/src/runapplication_UT.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+TEST_F(RunTest, AsPosixProcessPassesCorrectApplicationTypeToRun)
+{
+    RecordProperty("Description",
+                   "The Application instance passed to LifeCycleManager::run() is of the type given as "
+                   "the Run template argument.");
+
+    EXPECT_CALL(lifecycle_mock_, run(_, _))
+        .WillOnce([](score::mw::lifecycle::Application& app, const score::mw::lifecycle::ApplicationContext&) {
+            EXPECT_NE(dynamic_cast<StubApplication*>(&app), nullptr);
```
NicolasFussberger: Would it not be cleaner to have the StubApplication use MOCK_METHODs for implementation and having some expectation on the calls to StubApplication?

score/launch_manager/BUILD
```diff
@@ -71,10 +73,30 @@ cc_library(
...
+
+alias(
+    name = "applicationcontext_mock_cc",
+    actual = "//score/launch_manager/lifecycle_client:applicationcontext_mock",
+)
+
+alias(
+    name = "application_mock_cc",
+    actual = "//score/launch_manager/lifecycle_client:application_mock",
+)
```
NicolasFussberger: Do we need to expose mocks for Application and ApplicationContext for users?

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -13,54 +13,22 @@
...
-        "src/runapplication.cpp",
-    ],
-    hdrs = [
-        "src/aasapplicationcontainer.h",
-        "src/application.h",
-        "src/applicationcontext.h",
-        "src/lifecyclemanager.h",
-        "src/runapplication.h",
-    ],
     copts = select({
```
NicolasFussberger: Do we need -pthread on this high level or is it enough if this is at the cc_library that actually uses threads

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -13,54 +13,22 @@
...
     deps = [
-        ":lifecycle_client_interface",
-        "//score/launch_manager/lifecycle_client/src/details:lifecycle_client_impl",
+        ":aasapplicationcontainer",
+        ":application",
+        ":lifecyclemanager",
+        ":lifecycleclient_internal",
+        ":runapplication",
+        ":report_running",
         "@score_baselibs//score/language/futurecpp",
```
NicolasFussberger: I assume these baselibs dependency should be on the target that actually uses them and not on this target

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -120,9 +88,13 @@ cc_library(
         "src/lifecyclemanager.h",
     ],
     features = COMPILER_WARNING_FEATURES,
-    include_prefix = "score/mw/lifecycle/lifecycle_client",
+    include_prefix = "score/mw/lifecycle",
+        linkopts = select({
+        "@platforms//os:qnx": [],
+        "@platforms//os:linux": ["-lpthread"],
```
NicolasFussberger: Does this actually make use of pthread?

score/launch_manager/lifecycle_client/BUILD
NicolasFussberger, line 96: Why do we need application.h header here? Does this not come from :application dependency?

score/launch_manager/lifecycle_client/BUILD
```diff
@@ -138,49 +110,75 @@ cc_library(
...
-]
+cc_library(
+    name = "runapplication",
+    srcs = [
+        "src/runapplication.cpp",
+    ],
+    hdrs = [
+        "src/runapplication.h",
+        "src/lifecyclemanager.h",
+        "src/applicationcontext.h",
```
NicolasFussberger, line 121: Why do we need "src/applicationcontext.h" and "src/lifecyclemanager.h" here? Do they not come from dependencies?

NicolasFussberger (changes requested):

score/launch_manager/lifecycle_client/src/runapplication_UT.cpp
TimoSteuerwaldETAS: No, as these expectations are on the `lifecycle_mock`, but not on `StubApplication`. So of course we could override the API with `MOCK_METHOD `calls. However as there are no expectations needed, I would stick to the current implementation. This would be even more confusing.

But you are partly right: It is currently not easy to understand. According to [Martin Fowler](https://martinfowler.com/articles/mocksArentStubs.html) dummy would be the right prefix for `StubApplication `and `StubApplicationWithIntArg`, because both are only there to fulfill parameter lists, but are actually never called at all.

score/launch_manager/BUILD
TimoSteuerwaldETAS: Depends, `Application` is wrong (there is no need at all for a mock target, as it is a c++ interface anyways), but `ApplicationContext` could be useful to mock the arguments delivered to an app. Furthermore I am unsure how someone should use the `ApplicationContextMock` from the outside if we not export it. I guess for some reason the mock has been added.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Fixed.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: Good point. `jthread.hpp` is something different.

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: If we want to mock `ApplicationContext`, we can not use `application` package as dependency, as this then gets the real `applicationcontext` as a dependency (as `application.h` includes `applicationcontext.h`) . Therefore I've added headers which are referenced in the `lifecyclemanager.h`. (BTW, this was the error with the runapplication_UT, there we had `application` as dependency (see commit 819f3d0).

Same also for other targets.

We could do this also differently by introducing cc_targets for header and cpp files separately. See below:
```
cc_library(
    name = "application_hdr",
    hdrs = [
        "src/application.h",
    ],
    features = COMPILER_WARNING_FEATURES,
    include_prefix = "score/mw/lifecycle",
    strip_include_prefix = "/score/launch_manager/lifecycle_client/src",
    tags = ["FUSA"],
    visibility = ["//score/launch_manager/lifecycle_client:__subpackages__"],
    deps = [
        ":applicationcontext_hdr",
    ],
)

cc_library(
    name = "application_impl",
    srcs = [
        "src/application.cpp",
    ],
    features = COMPILER_WARNING_FEATURES,
    include_prefix = "score/mw/lifecycle",
    strip_include_prefix = "/score/launch_manager/lifecycle_client/src",
    tags = ["FUSA"],
    visibility = ["//score/launch_manager/lifecycle_client:__subpackages__"],
    deps = [
        ":application_hdr",
        ":applicationcontext_impl",
    ],
)
```

I don't like this solution as it makes our build file even more complex. What do you think?

score/launch_manager/lifecycle_client/BUILD
TimoSteuerwaldETAS: See above

score/launch_manager/lifecycle_client/BUILD
NicolasFussberger: For jthread there should be a dependency to baselibs futurecpp

score/launch_manager/lifecycle_client/BUILD
NicolasFussberger: Then lets keep is as is for now.

S-MOHAMD:

NicolasFussberger (approved):

NicolasFussberger:
> Please lnk corresponding issue/bug. see: https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

done

PR #243, opened 2026-06-10
Adding enum return type for SafeProcessMap

### LINKED ISSUE #222: Small Improvement: Improve SafeProcessMap by returning an enum instead of int
### What

Currently, the methods of [SafeProcessMap ](https://github.com/eclipse-score/lifecycle/blob/27353408e7901331d41569005d847cfc168ea3c2/score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp#L44 )class returns various integer values denoting different conditions.
To improve readability, an enum shall be returned instead.

### Acceptance Criteria (DoD)

SafeProcessMap returns an enum.
The code using SafeProcessMap is updated accordingly

### How

_No response_

NemanjaTrifunovicRTRK:
## Description

Extends `SafeProcessMap` with an enum class for the return value of public methods.

## Changes

* **`score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp`**

  * Updated public API with the enum class return value

* **`score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.cpp`**

  * Updated public API with the enum class return value

* **`score/launch_manager/daemon/src/process_group_manager/details/process_info_node.cpp`**

  * Added comparison with the enum class return value

* **`score/launch_manager/daemon/src/process_group_manager/details/os_handler.cpp`**

  * Added comparison with the enum class return value

* **`score/launch_manager/daemon/src/process_group_manager/details/oshandler_UT.cpp`**

  * Updated Unit Tests

* **`score/launch_manager/daemon/src/process_group_manager/details/safeprocessmap_UT.cpp`**

  * Updated Unit Tests

## Notes

Only the public API is changed, the private method API and logic are not changed.

## Verification

Verified with successful build and test execution
```bash
bazel build --config=x86_64-linux  //score/...
bazel test //examples/... --config=x86_64-linux --test_output=all --test_arg=-s
bazel test --config=x86_64-linux  //score/...
```

Closes #222


score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
```diff
@@ -46,22 +49,44 @@ class ITerminationCallback {
...
     uint32_t pid_right_ = 0xFFFFFFFF;  ///< Even branch for binary tree of process IDs (right child).
     osal::ProcessID pid_ = -1;         ///< Process ID associated with this item.
     ProcessInfoData data_;
 };
 
-/// @brief The SafeProcessMap class provides a thread-safe mapping of unique process IDs (ProcessID) to ITerminationCallback pointers.
-/// It ensures safe concurrent access and modification of the mapping, using atomic operations.
-class SafeProcessMap final {
-   public:
+enum class SafeProcessMapReturnTypes : std::int32_t
```
MaciejKaszynski: Nit pick I would make this not plural.
```suggestion
enum class SafeProcessMapReturnType : std::int32_t
```

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
```diff
@@ -46,22 +49,44 @@ class ITerminationCallback {
...
+    kYield = 1,
+
+    /// @brief An error occurred during insertion (e.g., out of memory).
+    kInsertionError = -1,
+
+    /// @brief The provided process ID (`key`) is not valid ( < 0).
+    kInvalidIdError = -2,
+
+    /// @brief The state is not defined.
+    kUndefined = 2,
```
MaciejKaszynski: nit pick I would make the undefined state 0 so that unset values are undefined and can be caught easier

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
MaciejKaszynski, line 146: This should also use the return type I think

also `insertNode` & `removeNode`

then could remove some static_casts from other functions

MaciejKaszynski (changes requested):

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: My goal was to avoid modifying internal methods and to retain the existing return values so there would be no need to change the logic. For that reason, I wouldn’t set an undefined value to 0.
The concept kUndefined = 2 was introduced solely for unit tests, since the value 2 was also used there as the undefined value.

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: Updated.

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
MaciejKaszynski: ok thats fine then

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: These are private method, they arenot used outside that class. Also, these methods' return value logic is not compatible with the methods `insertIfNotTerminated` and `findTerminated`, it is not possible to use the same (unique) enum class for their return value.

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
```diff
@@ -80,28 +105,32 @@ class SafeProcessMap final {
...
-    /// This method is called by a worker thread after starting a process. It attempts to insert the given process ID (key)
-    /// and its associated ITerminationCallback pointer into the map, ensuring that the process is not already marked as terminated.
-    /// In the case of a clash due to PID re-use, this method yields until the situation is resolved.
+    /// This method is called by a worker thread after starting a process. It attempts to insert the given process ID
+    /// (key) and its associated ITerminationCallback pointer into the map, ensuring that the process is not already
+    /// marked as terminated. In the case of a clash due to PID re-use, this method yields until the situation is
+    /// resolved.
     /// @param key The process ID to insert into the map.
     /// @param object A pointer to the ITerminationCallback associated with the process.
     /// @return 0 if the key (Process ID) was not found and a new entry was made,
```
MaciejKaszynski: Can you remove the returns comment for all methods that are now using the new enum.

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
MaciejKaszynski: I've had a better look at the class now and I think there is more cleanup to be done actually.
I think even the name of this function is lying about what its doing.

I made another ticket to followup on this. https://github.com/eclipse-score/lifecycle/issues/245
but leaving this is fine I think

MaciejKaszynski (changes requested):

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: Okay.
#222 can be closed?

MaciejKaszynski: yeah when this PR goes through. Only  thing needed is this I think https://github.com/eclipse-score/lifecycle/pull/243#discussion_r3396151071

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: Done.

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NemanjaTrifunovicRTRK: Can this PR be closed only after closing the issue https://github.com/eclipse-score/lifecycle/issues/245?

MaciejKaszynski (approved):

score/launch_manager/daemon/src/process_group_manager/details/safe_process_map.hpp
NicolasFussberger: Lets do #245 as a separate ticket. I think we can go ahead with the return type changes.

NicolasFussberger (approved):

PR #244, opened 2026-06-10
Fixing asan error on mpmc queue

### LINKED ISSUE #241: mpmc_concurrent_queue_test excluded from ASan/UBSan due to fake-stack alignment false positive
### Description

`MPMCConcurrentQueue` has `alignas(64)` members (`m_head`, `m_tail`, `m_stopped`). ASan's fake stack does not honor over-alignment, so when a test fixture embeds the queue as a value member, UBSan reports a misaligned reference at constructor entry (`mpmc_concurrent_queue.hpp:87`) before any queue logic executes. The test is otherwise clean.

Currently worked around with:

```python
tags = ["no-asan"]
```

in `concurrency/BUILD`.

### Analysis results

_No response_

### Solution
Three options: 
1. Add `alignas(CacheLineSize)` to the class itself - declares the alignment requirement as part of the type.
2. Add a UBSan alignment suppression in `score_cpp_policies` for `score::lcm::internal::MPMCConcurrentQueue`.
3. Heap-allocate the queue in test fixtures (`std::make_unique<...>()`).

### Error Occurrence  Rate

Reproducible

### How to reproduce

Remove `no-asan` from the target, then run:

```bash
bazel test --config=asan_ubsan_lsan --config=x86_64-linux \
  //score/launch_manager/daemon/src/common/concurrency:mpmc_concurrent_queue_test \
  --test_tag_filters= \
  --verbose_failures
```

### Supporting Information

_No response_

### Classification

Minor

### First Affected Release

not released (main)

### Last Affected Release

not released (main)

### Expected Fixed Release

before release (main)

### Category

- [ ] Safety Relevant
- [ ] Security Relevant

MaciejKaszynski:
Fixes the ASAN error with the `MPMCConcurrentQueue`
- Fixes https://github.com/eclipse-score/lifecycle/issues/241

Using manual padding instead of `alignas` so that ASAN's fake stack can correctly.
Setting the alignment on the class itself didn't fix the asan error as it would just set the alignment for the class itself.


VukPavicRTRK (changes requested):
Nice fix for the ASan false positive. One thing to call out: `CacheLinePaddedAtomic<T>` preserves cache-line *size* but not cache-line *alignment*, so it weakens the original false-sharing guarantee provided by `alignas(CacheLineSize)`. At minimum, the comments on `m_head`, `m_tail`, and `m_stopped` should be updated since they're no longer strictly "aligned on their own cache line."

MaciejKaszynski:
> Nice fix for the ASan false positive. One thing to call out: `CacheLinePaddedAtomic<T>` preserves cache-line _size_ but not cache-line _alignment_, so it weakens the original false-sharing guarantee provided by `alignas(CacheLineSize)`. At minimum, the comments on `m_head`, `m_tail`, and `m_stopped` should be updated since they're no longer strictly "aligned on their own cache line."

So I have done a benchmark and seen that false sharing doesn't actually cause any performance drawbacks. I have committed the benchmark and made the members not aligned at all .

S-MOHAMD:

score/launch_manager/src/daemon/src/common/concurrency/mpmc_concurrent_queue_benchmark.cpp
```diff
@@ -0,0 +1,121 @@
+/********************************************************************************
```
pawelrutkaq, line 1: code looks like part of codebase, maybe it shall be bench_test.cpp ?

MaciejKaszynski: [7c36518](https://github.com/eclipse-score/lifecycle/pull/244/commits/7c36518007802206d5ce9c516f4347cd136e9558) I've changed it to `benchmark`, I didn't want to add test as there is no pass or fail so think benchmark is better

NicolasFussberger (approved):

PR #248, opened 2026-06-15
Launch Manager loads the new configuration file

### LINKED ISSUE #209: Migrate Launch Manager from the reading the old configuration file format to the new format
### What

**Background**

* A new user-facing configuration format has been defined (see [here](https://github.com/eclipse-score/lifecycle/blob/main/src/launch_manager_daemon/config/config_schema/launch_manager.schema.json))
* However, the information from this file is still being mapped to the existing legacy configuration file format

**Proposal**

* Create a new flatbuffer format for the new configuration file
* Create the code to read in the new configuration file
* Adapt the launch manager to make use of the new configuration file (requiring initially some mapping of the new terms like RunTarget, Component, etc. to the existing code concepts)

### Acceptance Criteria (DoD)

* Launch Manager can read the new configuration file

### How

_No response_

paulquiring:
* Map user-facing json config to the new flatbuffer config
* Load new launch manager config
* Introduce adapters to make the rest of the code work with the new configuration content
* Changes are avaible using --//config:use_new_configuration=True

Closes: https://github.com/eclipse-score/lifecycle/issues/209

### DIFF
--- a/config/BUILD
+++ b/config/BUILD
@@ -72,7 +72,7 @@ config_setting(
 bool_flag(
     name = "use_new_configuration",
     build_setting_default = False,
-    visibility = ["//:__subpackages__"],
+    visibility = ["//visibility:public"],
 )
 
 config_setting(
@@ -80,5 +80,5 @@ config_setting(
     flag_values = {
         ":use_new_configuration": "True",
     },
-    visibility = ["//:__subpackages__"],
+    visibility = ["//visibility:public"],
 )
--- a/config/BUILD
+++ b/config/BUILD
@@ -68,3 +68,17 @@ config_setting(
     },
     visibility = ["//visibility:public"],
 )
+
+bool_flag(
+    name = "use_new_configuration",
+    build_setting_default = False,
+    visibility = ["//:__subpackages__"],
+)
+
+config_setting(
+    name = "lm_use_new_configuration",
+    flag_values = {
+        ":use_new_configuration": "True",
+    },
+    visibility = ["//:__subpackages__"],
+)
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/daemon/AliveMonitorImpl.cpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/daemon/AliveMonitorImpl.cpp
@@ -30,28 +30,46 @@ namespace saf
 namespace daemon
 {
 
-AliveMonitorImpl::AliveMonitorImpl(std::shared_ptr<score::lcm::IRecoveryClient> recovery_client,
-                                   std::unique_ptr<watchdog::IWatchdogIf> watchdog,
-                                   std::unique_ptr<score::lcm::IProcessStateReceiver> process_state_receiver)
+#ifdef USE_NEW_CONFIGURATION
+AliveMonitorImpl::AliveMonitorImpl(SptrIRecoveryClient recovery_client,
+                                   UptrIWatchdogIf watchdog,
+                                   UptrIProcessStateReceiver process_state_receiver,
+                                   const Config& config)
+    : m_recovery_client(recovery_client),
+      m_watchdog(std::move(watchdog)),
+      m_logger{score::lcm::saf::logging::PhmLogger::getLogger(score::lcm::saf::logging::PhmLogger::EContext::factory)},
+      m_process_state_receiver{std::move(process_state_receiver)},
+      m_config(config)
+{
+}
+#else
+AliveMonitorImpl::AliveMonitorImpl(SptrIRecoveryClient recovery_client,
+                                   UptrIWatchdogIf watchdog,
+                                   UptrIProcessStateReceiver process_state_receiver)
     : m_recovery_client(recovery_client),
       m_watchdog(std::move(watchdog)),
       m_logger{score::lcm::saf::logging::PhmLogger::getLogger(score::lcm::saf::logging::PhmLogger::EContext::factory)},
       m_process_state_receiver{std::move(process_state_receiver)}
 {
 }
+#endif
 
 EInitCode AliveMonitorImpl::init() noexcept
 {
-    score::lcm::saf::daemon::EInitCode initResult{score::lcm::saf::daemon::EInitCode::kGeneralError};
+    EInitCode initResult{EInitCode::kGeneralError};
     try
     {
         m_osClock.startMeasurement();
 
-        m_daemon = std::make_unique<score::lcm::saf::daemon::PhmDaemon>(
-            m_osClock, m_logger, std::move(m_watchdog), std::move(m_process_state_receiver));
+        m_daemon = std::make_unique<PhmDaemon>(m_osClock, m_logger, std::move(m_watchdog),
+                                               std::move(m_process_state_receiver));
+    #ifdef USE_NEW_CONFIGURATION
+        initResult = m_daemon->init(m_recovery_client, m_config);
+    #else
         initResult = m_daemon->init(m_recovery_client);
+    #endif
 
-        if (initResult == score::lcm::saf::daemon::EInitCode::kNoError)
+        if (initResult == EInitCode::kNoError)
         {
             const long ms{m_osClock.endMeasurement()};
             m_logger.LogDebug() << "AliveMonitor: Initialization took " << ms << " ms";
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/daemon/SwClusterHandler.hpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/daemon/SwClusterHandler.hpp
@@ -20,6 +20,9 @@
 #include "score/mw/launch_manager/alive_monitor/details/ifexm/ProcessStateReader.hpp"
 #include "score/mw/launch_manager/alive_monitor/details/logging/PhmLogger.hpp"
 #include "score/mw/launch_manager/alive_monitor/details/timers/Timers_OsClock.hpp"
+#ifdef USE_NEW_CONFIGURATION
+#include "score/mw/launch_manager/configuration/config.hpp"
+#endif
 #include <string>
 #include <vector>
 
@@ -92,6 +95,9 @@ class SwClusterHandler
     /// @param [in] f_bufferConfig_r           Configuration settings for constructing workers
     /// @return                              Construction is successful (true), otherwise failure (false)
     bool constructWorkers(
+#ifdef USE_NEW_CONFIGURATION
+        const score::mw::launch_manager::configuration::Config& config,
+#endif
         std::shared_ptr<score::lcm::IRecoveryClient> f_recoveryClient_r,
         ifexm::ProcessStateReader& f_processStateReader_r,
         const factory::MachineConfigFactory::SupervisionBufferConfig& f_bufferConfig_r) noexcept(false);
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
@@ -38,6 +38,11 @@ namespace saf
 namespace factory
 {
 
+using BufferConfig = MachineConfigFactory::SupervisionBufferConfig;
+using RecoveryClient = score::lcm::IRecoveryClient;
+using NanoSecondType = saf::timers::NanoSecondType;
+using IdentifierHash = score::lcm::IdentifierHash;
+
 namespace
 {
 /// @brief Prefix for all log messages
@@ -67,7 +72,7 @@ std::unique_ptr<char[]> read_flatbuffer_file(const std::string& f_filename_r)
 
 /* RULECHECKER_comment(0, 7, check_incomplete_data_member_construction, "Members processDataBase is not \
 required to be initialized using member initializer list.", true_no_defect) */
-FlatCfgFactory::FlatCfgFactory(const factory::MachineConfigFactory::SupervisionBufferConfig& f_bufferConfig_r)
+FlatCfgFactory::FlatCfgFactory(const BufferConfig& f_bufferConfig_r)
     : IPhmFactory(),
       bufferConfig_r(f_bufferConfig_r),
       flatBuffer_p(nullptr),
@@ -364,7 +369,7 @@ bool FlatCfgFactory::createSupervisionCheckpoints(std::vector<ifappl::Checkpoint
 bool FlatCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_alive_r,
                                              std::vector<ifappl::Checkpoint>& f_checkpoints_r,
                                              std::vector<ifexm::ProcessState>& f_processStates_r,
-                                             std::shared_ptr<score::lcm::IRecoveryClient> f_recoveryClient_r)
+                                             std::shared_ptr<RecoveryClient> f_recoveryClient_r)
 {
     bool isSuccess{true};
 
@@ -381,7 +386,7 @@ bool FlatCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_
             {
                 // Collect Alive Supervision configuration
                 const char* nameCfgAlive_p{hmAliveSupervision_p->ruleContextKey()->c_str()};
-                saf::timers::NanoSecondType aliveReferenceCycleCfg{
+                NanoSecondType aliveReferenceCycleCfg{
                     timers::TimeConversion::convertMilliSecToNanoSec(hmAliveSupervision_p->aliveReferenceCycle())};
                 uint32_t minAliveIndicationsCfg{hmAliveSupervision_p->minAliveIndications()};
                 uint32_t maxAliveIndicationsCfg{hmAliveSupervision_p->maxAliveIndications()};
@@ -409,7 +414,7 @@ bool FlatCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_
                 // coverity[cert_exp34_c_violation] PHM.ecucfgdsl Process.identifier MANDATORY
                 // coverity[dereference] PHM.ecucfgdsl Process.identifier MANDATORY
                 aliveSupCfg.processIdentifier =
-                    score::lcm::IdentifierHash{flatBuffer_p->process()->Get(processIndex)->identifier()->str()};
+                    IdentifierHash{flatBuffer_p->process()->Get(processIndex)->identifier()->str()};
 
                 f_alive_r.emplace_back(aliveSupCfg);
 
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
@@ -30,6 +30,10 @@ namespace saf
 namespace factory
 {
 
+using HMCoreEcuCfg = HMCOREFlatBuffer::HMCOREEcuCfg;
+using DeviceConfigurations = watchdog::IDeviceConfigFactory::DeviceConfigurations;
+using NanoSecondType = timers::NanoSecondType;
+
 namespace
 {
 /// @brief Prefix for all log messages
@@ -95,7 +99,7 @@ bool MachineConfigFactory::init() noexcept(false)
     return loadHmCoreConfig(flatBuffer_p);
 }
 
-bool MachineConfigFactory::loadHmCoreConfig(const HMCOREFlatBuffer::HMCOREEcuCfg* f_cfg_r) noexcept(false)
+bool MachineConfigFactory::loadHmCoreConfig(const HMCoreEcuCfg* f_cfg_r) noexcept(false)
 {
     loadHmSettings(*f_cfg_r);
     loadWatchdogDevices(*f_cfg_r);
@@ -104,7 +108,7 @@ bool MachineConfigFactory::loadHmCoreConfig(const HMCOREFlatBuffer::HMCOREEcuCfg
     return true;
 }
 
-void MachineConfigFactory::loadWatchdogDevices(const HMCOREFlatBuffer::HMCOREEcuCfg& f_flatBuffer_r) noexcept(false)
+void MachineConfigFactory::loadWatchdogDevices(const HMCoreEcuCfg& f_flatBuffer_r) noexcept(false)
 {
     const auto* watchdogs{f_flatBuffer_r.watchdogs()};
     if ((watchdogs == nullptr) || (watchdogs->size() == 0U))
@@ -137,7 +141,7 @@ void MachineConfigFactory::loadWatchdogDevices(const HMCOREFlatBuffer::HMCOREEcu
     }
 }
 
-void MachineConfigFactory::loadHmSettings(const HMCOREFlatBuffer::HMCOREEcuCfg& f_flatBuffer_r) noexcept(true)
+void MachineConfigFactory::loadHmSettings(const HMCoreEcuCfg& f_flatBuffer_r) noexcept(true)
 {
     const auto* configContainer{f_flatBuffer_r.config()};
     if ((configContainer != nullptr) && (configContainer->size() == 1U))
@@ -163,13 +167,12 @@ void MachineConfigFactory::loadHmSettings(const HMCOREFlatBuffer::HMCOREEcuCfg& 
     }
 }
 
-std::optional<watchdog::IDeviceConfigFactory::DeviceConfigurations> MachineConfigFactory::getDeviceConfigurations()
-    const
+std::optional<DeviceConfigurations> MachineConfigFactory::getDeviceConfigurations() const
 {
     return watchdogConfigs;
 }
 
-timers::NanoSecondType MachineConfigFactory::getCycleTimeInNs() const noexcept(true)
+NanoSecondType MachineConfigFactory::getCycleTimeInNs() const noexcept(true)
 {
     return cycleTimeNs;
 }
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
@@ -20,12 +20,15 @@
 #include "score/mw/launch_manager/alive_monitor/details/logging/PhmLogger.hpp"
 #include "score/mw/launch_manager/alive_monitor/details/timers/Timers_OsClock.hpp"
 #include "score/mw/launch_manager/alive_monitor/details/watchdog/IDeviceConfigFactory.hpp"
-
+#ifdef USE_NEW_CONFIGURATION
+#include "score/mw/launch_manager/configuration/config.hpp"
+#else
 namespace HMCOREFlatBuffer
 {
 /* RULECHECKER_comment(1:0,1:0, check_non_pod_struct, "External data type form generated flatbuffer code", true_no_defect) */
 struct HMCOREEcuCfg;
 }  // namespace PHMCOREFlatBuffer
+#endif
 
 namespace score
 {
@@ -78,7 +81,11 @@ public:
     ///         False, if an invalid machine configuration was provided.
     /// @note FlatCfg constructor does not define any exception guarantee and may throw a non specified exception
     /// @throws std::bad_alloc in case of insufficient memory
+#ifdef USE_NEW_CONFIGURATION
+    bool init(const score::mw::launch_manager::configuration::Config& config) noexcept(false);
+#else
     bool init() noexcept(false);
+#endif
 
     /// @copydoc IDeviceConfigFactory::getDeviceConfigurations()
     std::optional<watchdog::IDeviceConfigFactory::DeviceConfigurations> getDeviceConfigurations() const override;
@@ -92,6 +99,7 @@ public:
     const SupervisionBufferConfig& getSupervisionBufferConfig() const noexcept(true);
 
 private:
+#ifndef USE_NEW_CONFIGURATION
     /// @brief Loads the hm machine config
     /// @param [in] f_cfg_r The flatcfg api
     /// @throws std::bad_alloc for string allocation in case of insufficient memory
@@ -105,6 +113,7 @@ private:
     /// @brief Load HM settings from the machine config. I.e. buffer sizes, periodicity, etc.
     /// @param [in] f_flatBuffer_r The flatcfg buffer
     void loadHmSettings(const HMCOREFlatBuffer::HMCOREEcuCfg& f_flatBuffer_r) noexcept(true);
+#endif
 
     /// @brief Log all configuration settings
     void logConfiguration() noexcept(true);
@@ -119,9 +128,11 @@ private:
     /// @brief Configured supervision buffer sizes
     SupervisionBufferConfig supBufferCfg{};
 
+#ifndef USE_NEW_CONFIGURATION
     /// Pointer to HM Flat Buffer for given Software Cluster
     /// Raw pointer is used here because the memory is deallocated by FlatBuffer.
     const HMCOREFlatBuffer::HMCOREEcuCfg* flatBuffer_p;
+#endif
 
     /// Logger object for logging messages
     logging::PhmLogger& logger_r{logging::PhmLogger::getLogger(logging::PhmLogger::EContext::factory)};
--- a/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
+++ b/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
@@ -1,0 +1,462 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "score/mw/launch_manager/configuration/configuration_adapter.hpp"
+#include "score/mw/launch_manager/common/alive_interface_path.hpp"
+#include "score/mw/launch_manager/common/log.hpp"
+#include "score/mw/launch_manager/osal/num_cores.hpp"
+
+#include <algorithm>
+#include <cassert>
+#include <cstring>
+#include <map>
+#include <set>
+
+namespace score::mw::launch_manager::configuration {
+
+namespace {
+constexpr const char* kAliveInterfaceEnvName = "LCM_ALIVE_INTERFACE_PATH";
+constexpr uint32_t kDefaultProcessExecutionError = 1U;
+
+uint32_t defaultProcessorAffinityMask() {
+    return static_cast<uint32_t>((1ULL << score::lcm::internal::osal::getNumCores()) - 1ULL);
+}
+}  // namespace
+
+score::lcm::internal::osal::CommsType ConfigurationAdapter::mapApplicationType(
+    ApplicationType app_type) const {
+    switch (app_type) {
+        case ApplicationType::Reporting:
+        case ApplicationType::ReportingAndSupervised:
+            return score::lcm::internal::osal::CommsType::kReporting;
+        case ApplicationType::StateManager:
+            return score::lcm::internal::osal::CommsType::kControlClient;
+        case ApplicationType::Native:
+        default:
+            return score::lcm::internal::osal::CommsType::kNoComms;
+    }
+}
+
+bool ConfigurationAdapter::initialize(const Config& config) {
+    return buildFromConfig(config);
+}
+
+void ConfigurationAdapter::deinitialize() {
+    for (auto& process_group : process_groups_) {
+        for (auto& process : process_group.processes_) {
+            for (size_t i = 0U;
+                 i < score::lcm::internal::kArgvArraySize && process.startup_config_.argv_[i] != nullptr;
+                 ++i) {
+                free(const_cast<char*>(process.startup_config_.argv_[i]));
+                process.startup_config_.argv_[i] = nullptr;
+            }
+            for (size_t i = 0U; process.startup_config_.envp_[i] != nullptr; ++i) {
+                free(process.startup_config_.envp_[i]);
+                process.startup_config_.envp_[i] = nullptr;
+            }
+        }
+    }
+    component_by_name_.clear();
+    component_to_process_index_.clear();
+}
+
+void ConfigurationAdapter::fillStartupConfigFromDeployment(
+    const ComponentConfig& comp,
+    score::lcm::internal::osal::OsalConfig& startup) const {
+    const auto& deploy = comp.deployment_config;
+    const auto& props = comp.component_properties;
+
+    startup.executable_path_ = deploy.bin_dir + "/" + props.binary_name;
+    startup.short_name_ = comp.name;
+
+    startup.uid_ = deploy.sandbox.uid;
+    startup.gid_ = deploy.sandbox.gid;
+    startup.supplementary_gids_ = deploy.sandbox.supplementary_group_ids;
+    startup.security_policy_ = deploy.sandbox.security_policy.value_or("");
+    startup.scheduling_policy_ = deploy.sandbox.scheduling_policy;
+    startup.scheduling_priority_ = deploy.sandbox.scheduling_priority;
+    startup.cpu_mask_ = defaultProcessorAffinityMask();
+
+    startup.resource_limits_.stack_ = 0U;
+    startup.resource_limits_.cpu_ = 0U;
+    startup.resource_limits_.data_ = 0U;
+    startup.resource_limits_.as_ = 0U;
+    if (deploy.sandbox.max_memory_usage.has_value()) {
+        startup.resource_limits_.as_ = *deploy.sandbox.max_memory_usage;
+    }
+
+    startup.comms_type_ = mapApplicationType(props.application_profile.application_type);
+}
+
+void ConfigurationAdapter::fillStartupArguments(
+    const ComponentProperties& props,
+    score::lcm::internal::osal::OsalConfig& startup) const {
+    // strdup() returns nullptr on OOM. On this embedded target, OOM during daemon
+    // startup is unrecoverable — the OS will terminate the process.
+    size_t arg_index = 0U;
+    startup.argv_[arg_index++] = strdup(startup.executable_path_.c_str());
+
+    assert(props.process_arguments.size() <= score::lcm::internal::kMaxArg
+           && "Too many process arguments for argv array");
+    size_t max_args = std::min(props.process_arguments.size(),
+                               static_cast<size_t>(score::lcm::internal::kMaxArg));
+    for (size_t i = 0U; i < max_args; ++i) {
+        startup.argv_[arg_index++] = strdup(props.process_arguments[i].c_str());
+    }
+}
+
+size_t ConfigurationAdapter::fillStartupEnvironment(
+    const DeploymentConfig& deploy,
+    score::lcm::internal::osal::OsalConfig& startup) const {
+    size_t env_index = 0U;
+
+    assert(deploy.environmental_variables.size() + 1U <= score::lcm::internal::kMaxEnv
+           && "Too many environmental variables for envp array");
+    size_t max_env = std::min(deploy.environmental_variables.size(),
+                              static_cast<size_t>(score::lcm::internal::kMaxEnv));
+    size_t env_count = 0;
+    for (const auto& ev : deploy.environmental_variables) {
+        if (env_count >= max_env) {
+            break;
+        }
+
+        std::string env_str = std::string(ev.key()) + "=" + std::string(ev.value());
+        startup.envp_[env_index++] = strdup(env_str.c_str());
+        ++env_count;
+    }
+
+    return env_index;
+}
+
+void ConfigurationAdapter::appendAliveInterfaceEnvironment(
+    const ComponentConfig& comp,
+    size_t& env_index,
+    score::lcm::internal::osal::OsalConfig& startup) const {
+    bool is_supervised =
+        comp.component_properties.application_profile.application_type == ApplicationType::ReportingAndSupervised ||
+        comp.component_properties.application_profile.application_type == ApplicationType::StateManager;
+    if (!is_supervised || env_index >= static_cast<size_t>(score::lcm::internal::kMaxEnv)) {
+        return;
+    }
+
+    std::string iface_path =
+        std::string(kAliveInterfaceEnvName) + "=" + score::lcm::internal::aliveInterfacePath(comp.name);
+    startup.envp_[env_index++] = strdup(iface_path.c_str());
+}
+
+PgManagerConfig ConfigurationAdapter::buildPgManagerConfig(const ComponentConfig& comp) const {
+    PgManagerConfig pgm{};
+    const auto& deploy = comp.deployment_config;
+    const auto& props = comp.component_properties;
+
+    pgm.is_self_terminating_ = props.application_profile.is_self_terminating;
+    pgm.startup_timeout_ms_ = std::chrono::milliseconds(deploy.ready_timeout_ms);
+    pgm.termination_timeout_ms_ = std::chrono::milliseconds(deploy.shutdown_timeout_ms);
+    pgm.execution_error_code_ = kDefaultProcessExecutionError;
+    pgm.number_of_restart_attempts = 0U;
+    if (deploy.ready_recovery_action.has_value()) {
+        pgm.number_of_restart_attempts = deploy.ready_recovery_action->number_of_attempts;
+    }
+
+    return pgm;
+}
+
+DependencyList ConfigurationAdapter::buildDependencyList(const ComponentProperties& props) const {
+    DependencyList dependencies;
+    dependencies.reserve(props.depends_on.size());
+
+    for (const auto& dep_name : props.depends_on) {
+        Dependency dep{};
+        dep.process_state_ = score::lcm::ProcessState::kRunning;
+
+        auto dep_it = component_by_name_.find(dep_name);
+        if (dep_it != component_by_name_.end()) {
+            const auto& dep_props = dep_it->second->component_properties;
+            if (dep_props.ready_condition.has_value()) {
+                dep.process_state_ = dep_props.ready_condition->process_state == ProcessState::Running
+                                         ? score::lcm::ProcessState::kRunning
+                                         : score::lcm::ProcessState::kTerminated;
+            }
+        }
+
+        dep.target_process_id_ = IdentifierHash{dep_name};
+        dependencies.push_back(dep);
+    }
+
+    return dependencies;
+}
+
+OsProcess ConfigurationAdapter::buildOsProcess(
+    const ComponentConfig& comp,
+    uint32_t process_index) const {
+    OsProcess os_process{};
+    os_process.process_id_ = IdentifierHash{comp.name};
+    os_process.process_number_ = process_index;
+
+    auto& startup = os_process.startup_config_;
+    fillStartupConfigFromDeployment(comp, startup);
+    fillStartupArguments(comp.component_properties, startup);
+
+    size_t env_index = fillStartupEnvironment(comp.deployment_config, startup);
+    appendAliveInterfaceEnvironment(comp, env_index, startup);
+    startup.envp_[env_index] = nullptr;
+
+    os_process.pgm_config_ = buildPgManagerConfig(comp);
+    os_process.dependencies_ = buildDependencyList(comp.component_properties);
+
+    return os_process;
+}
+
+void ConfigurationAdapter::resolveDependsOnEntry(
+    const std::string& dep_name,
+    const DependsOnMap& depends_on_by_name,
+    std::vector<uint32_t>& indexes,
+    std::set<std::string>& visited) const {
+    if (!visited.insert(dep_name).second) {
+        return;
+    }
+
+    bool found = false;
+    auto comp_it = component_to_process_index_.find(dep_name);
+    if (comp_it != component_to_process_index_.end()) {
+        found = true;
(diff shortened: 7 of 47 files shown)


scripts/config_mapping/lifecycle_config.py
paulquiring, line 776: Needed? It's now in score_defaults

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.cpp
```diff
@@ -0,0 +1,289 @@
...
+        f_checkpoints_r.clear();
+        return false;
+    }
+
+    logger_r_.LogDebug() << kLogPrefix << "Supervision checkpoints created:" << f_checkpoints_r.size();
+    return true;
+}
+
+bool ConfigCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_alive_r,
+                                                std::vector<ifappl::Checkpoint>& f_checkpoints_r,
```
NicolasFussberger: Please add Unit Tests for the new file

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,392 @@
+/********************************************************************************
```
NicolasFussberger, line 1: Please add Unit Tests for the new file

NicolasFussberger, line 1: Can the old configuration manager implementation be removed?

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.hpp
```diff
@@ -0,0 +1,96 @@
+/********************************************************************************
```
NicolasFussberger: Can the old FlatCfgFactory and MachineConfigFactory implementations be removed?

score/launch_manager/daemon/src/alive_monitor/details/daemon/PhmDaemon.hpp
NicolasFussberger: I think MachineConfigFactory should get obsolete.
Here it still reads in the legacy hmcore bin file

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,392 @@
...
+        default:
+            return osal::CommsType::kNoComms;
+    }
+}
+
+// Maps SchedulingPolicy string to POSIX constant
+static int32_t mapSchedulingPolicy(int32_t policy_enum)
+{
+    // config.hpp uses int32_t directly from the FBS enum (OTHER=0, FIFO=1, RR=2)
+    switch (policy_enum)
```
NicolasFussberger: Is this mapping already done when creating the Config struct?

score/launch_manager/daemon/src/configuration/configuration_adapter.hpp
```diff
@@ -0,0 +1,92 @@
...
+#include "score/mw/launch_manager/common/constants.hpp"
+#include "score/mw/launch_manager/common/process_group_state_id.hpp"
+#include "score/mw/launch_manager/process_group_manager/iprocess.hpp"
+#include "score/mw/launch_manager/process_state_client/posix_process.hpp"
+#include "score/mw/launch_manager/configuration/configuration_manager.hpp"
+#include "score/mw/launch_manager/configuration/config.hpp"
+
+namespace score {
+namespace lcm {
+namespace internal {
```
NicolasFussberger: I assume this new class should also use the new namespace score::mw::launch_manager::configuration

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include "score/mw/launch_manager/configuration/config.hpp"
+
+namespace score
+{
+namespace lcm
+{
+namespace saf
+{
+namespace factory
+{
```
NicolasFussberger: I assume the new namespace should be score::mw::launch_manager::alive_monitor

score/launch_manager/daemon/src/main.cpp
```diff
@@ -119,12 +116,24 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] const char* argv[])
...
-        //     LM_LOG_FATAL() << "LCM could not daemonize!, error:" << strerror(errno);
-        //     return EXIT_FAILURE;
-        // }
+        score::mw::launch_manager::configuration::FlatbufferConfigLoader config_loader;
+        auto config_result = config_loader.load(score::filesystem::Path(config_path));
+        if (!config_result.has_value())
+        {
+            LM_LOG_FATAL() << "Failed to load configuration from:" << config_path;
+            close(osal::IpcCommsSync::sync_fd);
+            close(osal::IpcCommsSync::control_client_handler_nudge_fd);
```
NicolasFussberger: The code for closing the two fds appears three times in the file. Could we move this to a separate method? Or wrap this in an RAII-style struct together with the reserveFD?

S-MOHAMD:

scripts/config_mapping/config.bzl
```diff
@@ -26,49 +24,22 @@ def _launch_manager_config_impl(ctx):
...
+        executable = flatc,
         mnemonic = "LaunchManagerFlatbufferConfigGeneration",
-        progress_message = "compiling generated Launch Manager configs in {} to flatbuffer files in {}".format(gen_dir_json.short_path, gen_dir_flatbuffer.short_path),
+        progress_message = "compiling Launch Manager config in {} to flatbuffer in {}".format(gen_dir_json.short_path, gen_dir_flatbuffer.short_path),
+        arguments = [
+            "-b",
+            "-o",
+            gen_dir_flatbuffer.path,
+            lm_schema.path,
+            gen_dir_json.path + "/launch_manager_config.json",
```
paulquiring: The name should not be hard coded. The name of the input shall be taken like this <input_name>_gen.json for the output filename.

score/launch_manager/daemon/src/main.cpp
```diff
@@ -104,8 +86,15 @@ void reserveFD(int fd)
 /// @param argv Array of command-line arguments.
 /// @return The exit code. 0 for success, non-zero for failure.
 // coverity[autosar_cpp14_a15_3_3_violation:FALSE] Only logging occurs outside the try-catch enclosing main().
-int main([[maybe_unused]] int argc, [[maybe_unused]] const char* argv[])
+int main(int argc, const char* argv[])
 {
+    const char* config_path = "etc/launch_manager_config.bin";
```
paulquiring, line 137: Add a getting_started.rst in score/launch_manager/docs/user_guide/ 
just having a minimal example also explaining the -c (--config) option.

Add -h (--help), which is also printed if an unknown option was used.
After printing the help terminate.

score/launch_manager/daemon/src/main.cpp
```diff
@@ -127,26 +116,40 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] const char* argv[])
...
         std::unique_ptr<score::lcm::internal::IAliveMonitorThread> aliveMonitorThread{
             std::make_unique<score::lcm::internal::AliveMonitorThread>(std::move(healthMonitor))};
 
         std::unique_ptr<ProcessGroupManager> process_group_manager = std::make_unique<ProcessGroupManager>(
             std::move(aliveMonitorThread), recoveryClient, std::move(process_state_notifier));
 
-        if (initializeLCMDaemon(*process_group_manager))
+        if (process_group_manager->initialize(*config_result))
         {
+            LM_LOG_INFO() << "LCM started successfully";
```
paulquiring: Remove this log it is misleading.

score/launch_manager/daemon/src/main.cpp
```diff
@@ -127,26 +116,40 @@ int main([[maybe_unused]] int argc, [[maybe_unused]] const char* argv[])
...
             if (runLCMDaemon(*process_group_manager))
             {
                 exit_code = EXIT_SUCCESS;
             }
         }
+        else
+        {
+            LM_LOG_FATAL() << "LCM startup failed";
+        }
```
paulquiring, line 218: Remove this log it is to verbose.

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,413 @@
...
+
+        auto& startup = os_process.startup_config_;
+        const auto& deploy = comp.deployment_config;
+        const auto& props = comp.component_properties;
+
+        std::string exec_path = deploy.bin_dir;
+        if (!exec_path.empty() && exec_path.back() != '/') {
+            exec_path += '/';
+        }
+        exec_path += props.binary_name;
```
paulquiring: Instead of constructing the exec_path in code it can be directly read from configuration. binary_name and bin_dir can be combined to executable_path in score/lifecycle/score/launch_manager/daemon/src/configuration/details/new_lm_flatcfg.fbs.

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,413 @@
...
+            }
+            for (size_t i = 0U; process.startup_config_.envp_[i] != nullptr; ++i) {
+                free(process.startup_config_.envp_[i]);
+                process.startup_config_.envp_[i] = nullptr;
+            }
+        }
+    }
+}
+
+bool ConfigurationAdapter::buildFromConfig(const Config& config) {
```
paulquiring: This function is way to long it needs to be simplified.

score/launch_manager/daemon/src/alive_monitor/details/daemon/PhmDaemon.hpp
paulquiring: It's loading now the new configuration. We have to keep it for now.

score/launch_manager/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -40,59 +39,27 @@ namespace factory
 
 namespace
 {
-/// @brief Prefix for all log messages
-// coverity[autosar_cpp14_a2_10_4_violation:FALSE] Empty namespace ensures uniqueness for cpp file scope
 static constexpr const char* kLogPrefix{"Factory for FlatCfg AR24-11:"};
```
paulquiring: Can be deleted.

score/launch_manager/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -205,45 +147,35 @@ bool FlatCfgFactory::initIpcServerWithUidBasedAccess(ifappl::CheckpointIpcServer
...
-    if (flatBuffer_p->hmMonitorInterface() != nullptr)
+    try
     {
-        try
+        f_interfaceIpcs_r.reserve(supervised_components_.size());
+
+        for (const auto* comp : supervised_components_)
         {
-            f_interfaceIpcs_r.reserve(static_cast<size_t>(flatBuffer_p->hmMonitorInterface()->size()));
+            const std::string pathInterface = "lifecycle_health_" + comp->name;
```
paulquiring: Move this to a free function which is also used in the ConfigAdapter class.

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,413 @@
...
+            if (env_count >= max_env) break;
+            std::string env_str = std::string(ev.key()) + "=" + std::string(ev.value());
+            startup.envp_[env_index++] = strdup(env_str.c_str());
+            ++env_count;
+        }
+
+        bool is_sup = (props.application_profile.application_type == ApplicationType::ReportingAndSupervised ||
+                        props.application_profile.application_type == ApplicationType::StateManager);
+        if (is_sup && env_index < static_cast<size_t>(score::lcm::internal::kMaxEnv)) {
+            std::string iface_path = "LCM_ALIVE_INTERFACE_PATH=/lifecycle_health_" + comp.name;
```
paulquiring: The free function shall add the /

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.hpp
paulquiring: Not relevant anymore. The implementation changed.

score/launch_manager/daemon/src/configuration/configuration_adapter.hpp
paulquiring: is fixed.

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
paulquiring: Function is removed.

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.hpp
paulquiring: The implementation changed in a way so that they are still needed.

score/launch_manager/docs/user_guide/getting_started.rst
paulquiring, line 1: move this to the correct location

paulquiring: done

score/launch_manager/daemon/src/alive_monitor/details/factory/ConfigCfgFactory.cpp
paulquiring: added [see](https://github.com/eclipse-score/lifecycle/pull/248/changes#diff-59053c90fe6c7cbe05d27f18232d00c604025baf786e25c967866f5c871b36e3)

score/launch_manager/daemon/src/main.cpp
paulquiring: This changed in the newer version too.

score/launch_manager/daemon/src/main.cpp
paulquiring: This is implemented now.

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
paulquiring: done

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
paulquiring: done

score/launch_manager/daemon/src/configuration/configuration_adapter.cpp
paulquiring: done

score/launch_manager/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
paulquiring: done

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,419 @@
...
+    if (deploy.sandbox.max_memory_usage.has_value()) {
+        startup.resource_limits_.as_ = *deploy.sandbox.max_memory_usage;
+    }
+
+    startup.comms_type_ = mapApplicationType(props.application_profile.application_type);
+
+    // strdup() returns nullptr on OOM. On this embedded target, OOM during daemon
+    // startup is unrecoverable — the OS will terminate the process.
+    size_t arg_index = 0U;
+    startup.argv_[arg_index++] = strdup(startup.executable_path_.c_str());
```
NicolasFussberger, line 107: I think argv[0] should have just the executable name, not the whole path

NicolasFussberger: Ah no..seems like it always had the executable path.

NicolasFussberger, line 107: Du we need to strdup the string here or should we just point to the c_str()?

paulquiring: We could use comp.deployment_config.executable_path.c_str(), but this would create a mixed ownership between the configuration object in main and the adapter. The Adapter copies the configuration so that it owns it. Yes this costs a bit of resources but is the simplest and clearest way currently to bridge the old to new configuration.

score/launch_manager/docs/user_guide/getting_started.rst
```diff
@@ -0,0 +1,154 @@
...
+###############
+
+This guide walks through creating a minimal Launch Manager configuration and running the
+daemon for the first time.
+
+Prerequisites
+*************
+
+- The ``launch_manager`` binary built and available on your target.
+- Python 3 with the ``jsonschema`` package installed (required for config validation).
```
NicolasFussberger: I think the user does not really need to be aware of these things as this is all handled in the background. python and `jsonschema` dependency is automatically handled via bazel. I don't think anyone (besides LaunchManager developers) should try to invoke `lifecycle_config.py` themselves.

All the user needs to be aware of is:
* to use the provided bazel abstraction
* How to write a config file
* start launch_manager with "-c" argument

```starlark
load("//:defs.bzl", "launch_manager_config")

launch_manager_config(
    name = "examples_test_config",
    config = ":lifecycle_demo_test.json",
    flatbuffer_out_dir = "etc",
)
```

score/launch_manager/docs/user_guide/getting_started.rst
```diff
@@ -0,0 +1,154 @@
...
+workflow is:
+
+1. Write a human-readable JSON config file.
+2. Convert it to a FlatBuffers-compatible JSON using ``lifecycle_config.py``.
+3. Compile the FlatBuffers JSON to a binary with ``flatc``.
+
+Step 1 — Write the JSON config
+==============================
+
+Create a file ``my_config.json``:
```
NicolasFussberger: I wonder if we should just link to an example config (or include the example config) from `examples/demo_verification/lifecycle_demo_test.json` so that we do not duplicate a config here.

For example:
```rst
  .. dropdown:: example_launch_manager_config.json

     .. literalinclude:: /../../../examples/demo_verification/lifecycle_demo_test.json
        :language: json
```

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/SwClusterHandler.hpp
```diff
@@ -74,46 +73,26 @@ class SwClusterHandler
...
 
     /// @brief No Move Assignment
     SwClusterHandler& operator=(SwClusterHandler&&) = delete;
 
-    /// @brief Construct required worker objects for the Software Cluster
-    /// @details Construct the interfaces, checkpoints, supervisions and recovery notifications
-    /// @param [in] f_recoveryClient_r       Interface to the launch manager for recovery
-    /// @param [in] f_processStateReader_r   Process state reader object for PHM daemon
-    /// @param [in] f_bufferConfig_r           Configuration settings for constructing workers
-    /// @return                              Construction is successful (true), otherwise failure (false)
```
NicolasFussberger, line 93: Maybe we can bring back some of the doxygen as that seems still valid and to some extent helpful

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -101,78 +68,57 @@ bool FlatCfgFactory::createProcessStates(std::vector<ifexm::ProcessState>& f_pro
 {
     bool isSuccess{true};
 
-    if (flatBuffer_p->process() != nullptr)
+    try
     {
-        try
+        supervised_components_.clear();
```
NicolasFussberger: should we move the setup of supervised_components into the init method?
Other methods also assume this vector being already filled which creates a non-obvious dependency that `createProcessStates` must always be called first.

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -272,29 +202,20 @@ bool FlatCfgFactory::createMonitorIf(std::vector<ifappl::MonitorIfDaemon>& f_int
         for (auto& interfaceIpc : f_interfaceIpcs_r)
         {
             f_interfaces_r.emplace_back(interfaceIpc, interfaceIpc.getPath().data());
+            f_processStates_r.at(static_cast<size_t>(index)).attachObserver(f_interfaces_r.back());
 
-            // Shared Memories (f_interfaceIpcs_r) has also been created from PhmMonitorInterface config
-            // Accessing the same index (vector <-> config vector) is mapping the same configuration.
-            // coverity[cert_exp34_c_violation] f_interfaceIpcs_r is created from PhmMonitorInterface
-            // coverity[dereference] f_interfaceIpcs_r is created from PhmMonitorInterface same size assured
-            auto refProcessIndex{flatBuffer_p->hmMonitorInterface()->Get(index)->refProcessIndex()};
```
NicolasFussberger, line 280: Previously this was getting the processIndex via the config. Now we rely on the assumption that for every component in supervised_components vector, a CheckpointIpcServer and ProcessState have been created with same ordering as in supervised_components vector. This assumption should be fine now, as there is a 1:1 mapping between Process and Alive supervision (this was not true before, as there could be multiple independent checkpoint reporters per process).

I wonder if we should make this more obvious by using the same index for those vectors.

E.g.

```c++
        // before
        uint32_t index{0U};
        f_interfaces_r.reserve(f_interfaceIpcs_r.size());
        for (auto& interfaceIpc : f_interfaceIpcs_r)
        {
            f_interfaces_r.emplace_back(interfaceIpc, interfaceIpc.getPath().data());
            f_processStates_r.at(static_cast<size_t>(index)).attachObserver(f_interfaces_r.back());

            logger_r.LogDebug() << "Successfully created MonitorInterface:" << f_interfaces_r.back().getInterfaceName();
            index++;
        }
        // suggestion
        f_interfaces_r.reserve(supervised_components_.size());
        for(std::size_t compIndex = index; compIndex < supervised_components_.size(); compIndex++) {
            auto& interfaceIpc = f_interfaceIpcs_r.at(compIndex);
            f_interfaces_r.emplace_back(interfaceIpc, interfaceIpc.getPath().data());
            f_processStates_r.at(static_cast<size_t>(compIndex)).attachObserver(f_interfaces_r.back());
            logger_r.LogDebug() << "Successfully created MonitorInterface:" << f_interfaces_r.back().getInterfaceName();
        }
```

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -307,55 +228,40 @@ bool FlatCfgFactory::createSupervisionCheckpoints(std::vector<ifappl::Checkpoint
...
-        }
-        catch (const std::exception& f_exception_r)
+        for (size_t idx = 0; idx < supervised_components_.size(); ++idx)
         {
-            isSuccess = false;
-            logger_r.LogError() << kLogPrefix << "Could not create supervision worker objects, due to exception:"
-                                << std::string_view{f_exception_r.what()};
+            const auto* comp = supervised_components_[idx];
+            const std::string checkpointCfgName = comp->name + "_checkpoint";
+            const uint32_t checkpointId = 1U;
```
NicolasFussberger: Not sure if this is worth it, but the checkpointId has to match this https://github.com/eclipse-score/lifecycle/blob/58d6d1b883eb17c8ff3175937d9905a138ed32b5/score/launch_manager/src/alive/src/alive.cpp#L21

I wonder if we should make this a static constant in score/launch_manager/src/daemon/src/alive_monitor/details/factory/StaticConfig.hpp which is then used in both places.

The background is, when this code was still supporting Deadline and Logical supervisions, then different kinds of checkpoints were required.
But now only Alive supervision support is left, which only needs sending "Alive notificiations" - for which he just use checkpointId 1.

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -368,77 +274,69 @@ bool FlatCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_
...
-                    score::lcm::IdentifierHash{flatBuffer_p->process()->Get(processIndex)->identifier()->str()};
-
-                f_alive_r.emplace_back(aliveSupCfg);
-
-                // Subscribe created Alive Supervision to ProcessState class
-                f_processStates_r.at(static_cast<size_t>(processIndex)).attachObserver(f_alive_r.back());
-
-                logger_r.LogDebug() << kLogPrefix << "Successfully created alive supervision worker object:"
-                                    << f_alive_r.back().getConfigName();
+                continue;
```
NicolasFussberger: I think this case should not really happen. If the component is supervised then alive supervision had to be specified.
Maybe we should have an assert here?

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -368,77 +274,69 @@ bool FlatCfgFactory::createAliveSupervisions(std::vector<supervision::Alive>& f_
...
+            aliveSupCfg.aliveReferenceCycle = aliveReferenceCycleCfg;
+            aliveSupCfg.minAliveIndications = minAliveIndicationsCfg;
+            aliveSupCfg.maxAliveIndications = maxAliveIndicationsCfg;
+            aliveSupCfg.isMinCheckDisabled = isMinCheckDisabledCfg;
+            aliveSupCfg.isMaxCheckDisabled = isMaxCheckDisabledCfg;
+            aliveSupCfg.failedCyclesTolerance = failedCyclesToleranceCfg;
+            aliveSupCfg.checkpointBufferSize = bufferConfig_r.bufferSizeAliveSupervision;
+            aliveSupCfg.recoveryClient = f_recoveryClient_r;
+
+            aliveSupCfg.processIdentifier = score::lcm::IdentifierHash{comp->name};
```
NicolasFussberger: I some other place the function `getProcessId` was used to calculate the process Id. Maybe we should use the same way, to avoid accidentally using a different string as a base.

Maybe change getProcessId to this, then the decision which string is hashed is in a single place:
```
score::lcm::IdentifierHash getProcessId(const score::mw::launch_manager::configuration::ComponentConfig* config) {
    return score::lcm::IdentifierHash{comp->name};
}
```

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
```diff
@@ -73,44 +65,23 @@ class MachineConfigFactory : public watchdog::IDeviceConfigFactory
...
-
-    /// @brief Load HM settings from the machine config. I.e. buffer sizes, periodicity, etc.
-    /// @param [in] f_flatBuffer_r The flatcfg buffer
-    void loadHmSettings(const HMCOREFlatBuffer::HMCOREEcuCfg& f_flatBuffer_r) noexcept(true);
-
     /// @brief Log all configuration settings
     void logConfiguration() noexcept(true);
 
     /// @brief Configured watchdog devices
-    /// By default, no watchdog device is configured
```
NicolasFussberger, line 113: Maybe we should bring back some of this doxygen, thats seems rather useful

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
```diff
@@ -31,136 +29,39 @@ namespace factory
...
-bool MachineConfigFactory::init() noexcept(false)
+bool MachineConfigFactory::init(const score::mw::launch_manager::configuration::Config& config) noexcept(false)
 {
-    std::unique_ptr<char[]> loadBuffer_p = read_flatbuffer_file("hmcore.bin");
-    if(!loadBuffer_p) {
-        logger_r.LogInfo() << kLogPrefix << "No HM Machine Configuration found. Using default configuration.";
-        logConfiguration();
-        return true;
+    const auto& alive_sup = config.aliveSupervision();
+    if (alive_sup.evaluation_cycle_ms != 0U)
```
NicolasFussberger: `alive_sup.evaluation_cycle_ms == 0` should never happen. Maybe we should make this an assert?
Maybe [SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE](https://github.com/eclipse-score/baselibs/blob/f0b79795718aa4d125873e73a7625230ffd0c6a0/score/language/futurecpp/include/score/assert.hpp#L111)

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
```diff
@@ -31,136 +29,39 @@ namespace factory
...
-    if (flatBuffer_p == nullptr)
+    const auto& wd_opt = config.watchdog();
+    if (wd_opt.has_value())
     {
-        logger_r.LogError() << kLogPrefix << "Reading HM configuration from FlatCfg failed.";
-        return false;
+        const auto& wd = *wd_opt;
+        watchdog::DeviceConfig wdConfig{};
+        wdConfig.fileName = wd.device_file_path;
+        assert(wd.max_timeout_ms <= std::numeric_limits<std::uint16_t>::max());
```
NicolasFussberger: Should we make this a validation in score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp ?

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
```diff
@@ -42,14 +37,11 @@ class PhmLogger;
...
 {
 
-/// @brief Factory for loading the HM Machine Configuration
-/// @details Provides methods to retrieve the settings from the HM Machine configuration if a configuration is
-/// provided. If no configuration is provided, the default values are returned.
+/// @brief Factory for loading the HM Machine Configuration from the unified launch_manager_config.bin
 class MachineConfigFactory : public watchdog::IDeviceConfigFactory
 {
 public:
     /// @brief Holds different buffer sizes that may be configured in the HM Machine Config
```
NicolasFussberger, line 54: The comment is no longer valid.
All these buffer sizes are currently not configurable anymore.

This seems like some fine tuning we may enable with e.g. CLI flags in the future. For now the default values should be sufficient

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,419 @@
...
+        if (props.ready_condition.has_value()) {
+            dep.process_state_ = (props.ready_condition->process_state == ProcessState::Running)
+                ? score::lcm::ProcessState::kRunning
+                : score::lcm::ProcessState::kTerminated;
+        } else {
+            dep.process_state_ = score::lcm::ProcessState::kRunning;
+        }
+        dep.target_process_id_ = IdentifierHash{dep_name};
+        os_process.dependencies_.push_back(dep);
+    }
```
NicolasFussberger, line 194: I think here is a bug. We need to lookup the ReadyCondition process_state for the referenced component.

Image

I think it would be good to extend the UTs with these different state combinations to catch this

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,419 @@
...
+
+    {
+        const auto& fallback = config.fallbackRunTarget();
+        ProcessGroupState fallback_state;
+        fallback_state.name_ = IdentifierHash{"MainPG/fallback_run_target"};
+        for (const auto& comp_name : fallback.depends_on) {
+            auto it = component_to_process_index.find(comp_name);
+            if (it != component_to_process_index.end()) {
+                fallback_state.process_indexes_.push_back(it->second);
```
NicolasFussberger: I think here we also need to resolve the dependencies recursively.

E.g. fallback_run_target depends_on [CompA]
and CompA depends_on [CompB] then

fallback_run_target = [CompA, CompB]

I think it would be good to extend the UT with this case

score/launch_manager/docs/user_guide/getting_started.rst
paulquiring: done here https://github.com/eclipse-score/lifecycle/pull/248/changes#diff-cf0cb0e9ec62a01afbed5ad12df6cc113d0beb22a8415bcd5ef51b79e7ba3781

score/launch_manager/docs/user_guide/getting_started.rst
paulquiring: This worked as suggested and makes the getting stared much shorter!

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/SwClusterHandler.hpp
paulquiring: done

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
paulquiring: done

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
paulquiring: Yes this is much better, it's pushed.

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
paulquiring: We would need to reorganize the code a bit 
Looking at:

score/launch_manager/src/daemon/src/alive_monitor/details/factory/StaticConfig.hpp
score/launch_manager/src/alive/src/alive.cpp

we might want to move StaticConfig.hpp into score/launch_manager/src/common or a similar location. We might want to create a ticket out if this comment to work in it later?

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
paulquiring: done

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
paulquiring: Going for [SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE](https://github.com/eclipse-score/baselibs/blob/f0b79795718aa4d125873e73a7625230ffd0c6a0/score/language/futurecpp/include/score/assert.hpp#L111) seems to be a good idea. I chose assert to be consistent with the rest of the code base. We should change this in the code base with a separate PR & Issue.

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.hpp
paulquiring: changed to "Holds buffer sizes for supervision objects (currently using compile-time defaults)"

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
NicolasFussberger: Ticket for removal of this id already exists https://github.com/eclipse-score/lifecycle/issues/251

score/launch_manager/src/daemon/src/alive_monitor/details/factory/FlatCfgFactory.cpp
```diff
@@ -28,429 +27,306 @@
...
+  }
+
+  return isSuccess;
 }
 
-std::optional<common::ProcessId> FlatCfgFactory::getProcessId(const std::string& f_processPath_r) noexcept(
-    true)
-{
-    return score::lcm::IdentifierHash{f_processPath_r}.data();
+std::optional<common::ProcessId> FlatCfgFactory::getProcessId(
```
NicolasFussberger: std::optional no longer required, hashing cannot fail

score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp
NicolasFussberger: Ticket created https://github.com/eclipse-score/lifecycle/issues/263

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    const std::string initial_run_target_name = std::string(config.initialRunTarget());
+
+    ProcessGroup pg;
+    const IdentifierHash pg_name{"MainPG"};
+    pg.name_ = pg_name;
+    pg.sw_cluster_ = IdentifierHash{"DefaultSoftwareCluster"};
+    pg.off_state_ = IdentifierHash{"MainPG/Off"};
+
+    IdentifierHash recovery_state{"Recovery"};
+    for (const auto& rt : config.runTargets()) {
```
NicolasFussberger: Could be changed to just     
`IdentifierHash recovery_state = IdentifierHash{std::string("MainPG/") + "fallback_run_target"};`

without the for loop

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+        }
+    }
+    pg.recovery_state_ = recovery_state;
+
+    std::map<std::string, const ComponentConfig*> component_by_name;
+    for (const auto& comp : config.components()) {
+        component_by_name[comp.name] = &comp;
+    }
+
+    std::map<std::string, uint32_t> component_to_process_index;
```
NicolasFussberger: do we need a separate map just for the index or can we get the index from component_by_name?

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    startup.comms_type_ = mapApplicationType(props.application_profile.application_type);
+
+    // strdup() returns nullptr on OOM. On this embedded target, OOM during daemon
+    // startup is unrecoverable — the OS will terminate the process.
+    size_t arg_index = 0U;
+    startup.argv_[arg_index++] = strdup(startup.executable_path_.c_str());
+    size_t max_args = std::min(props.process_arguments.size(),
+                               static_cast<size_t>(score::lcm::internal::kMaxArg));
```
NicolasFussberger, line 112: should we assert that we have enough space for all configured process_arguments and env variables?
kMaxArg with 20 is rather small, can we bump this to 100?

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+        if (env_count >= max_env) break;
+        std::string env_str = std::string(ev.key()) + "=" + std::string(ev.value());
+        startup.envp_[env_index++] = strdup(env_str.c_str());
+        ++env_count;
+    }
+
+    bool is_sup = (props.application_profile.application_type == ApplicationType::ReportingAndSupervised ||
+                    props.application_profile.application_type == ApplicationType::StateManager);
+    if (is_sup && env_index < static_cast<size_t>(score::lcm::internal::kMaxEnv)) {
+        std::string iface_path = "LCM_ALIVE_INTERFACE_PATH=" + score::lcm::internal::aliveInterfacePath(comp.name);
```
NicolasFussberger: we could also add a constant for the env name

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+        dep.process_state_ = score::lcm::ProcessState::kRunning;
+        auto dep_it = component_by_name.find(dep_name);
+        if (dep_it != component_by_name.end()) {
+            const auto& dep_props = dep_it->second->component_properties;
+            if (dep_props.ready_condition.has_value()) {
+                dep.process_state_ = (dep_props.ready_condition->process_state == ProcessState::Running)
+                    ? score::lcm::ProcessState::kRunning
+                    : score::lcm::ProcessState::kTerminated;
+            }
+        }
```
NicolasFussberger, line 190: here is a possible error case. In case the configured dependency does not actually exist

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    pgm.is_self_terminating_ = props.application_profile.is_self_terminating;
+    pgm.startup_timeout_ms_ = std::chrono::milliseconds(deploy.ready_timeout_ms);
+    pgm.termination_timeout_ms_ = std::chrono::milliseconds(deploy.shutdown_timeout_ms);
+    pgm.execution_error_code_ = kDefaultProcessExecutionError;
+    pgm.number_of_restart_attempts = 0U;
+    if (deploy.ready_recovery_action.has_value()) {
+        pgm.number_of_restart_attempts = deploy.ready_recovery_action->number_of_attempts;
+    }
+
+    for (const auto& dep_name : props.depends_on) {
```
NicolasFussberger, line 178: depends_on might also contain a run_target and not only a component.
Special Case: This may be also the fallback_run_target

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+            }
+            auto comp_cfg = component_by_name.find(dep_name);
+            if (comp_cfg != component_by_name.end()) {
+                for (const auto& sub_dep : comp_cfg->second->component_properties.depends_on) {
+                    resolve_depends(sub_dep, indexes, visited);
+                }
+            }
+        }
+        auto rt_it = run_target_by_name.find(dep_name);
+        if (rt_it != run_target_by_name.end()) {
```
NicolasFussberger: I think depends_on might also refer to fallback_run_target which is currently not in the run_target_by_name map

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    }
+
+    const auto& run_targets = config.runTargets();
+
+    std::map<std::string, const RunTargetConfig*> run_target_by_name;
+    for (const auto& rt : run_targets) {
+        run_target_by_name[rt.name] = &rt;
+    }
+
+    std::map<std::string, const ComponentConfig*> component_by_name;
```
NicolasFussberger: component_by_name exists twice in this class

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+        component_by_name[comp.name] = &comp;
+    }
+
+    std::function<void(const std::string&, std::vector<uint32_t>&, std::set<std::string>&)> resolve_depends;
+    resolve_depends = [&](const std::string& dep_name,
+                          std::vector<uint32_t>& indexes,
+                          std::set<std::string>& visited) {
+        if (!visited.insert(dep_name).second) return;
+        auto comp_it = component_to_process_index.find(dep_name);
+        if (comp_it != component_to_process_index.end()) {
```
NicolasFussberger: There might be some possible error cases when component/runTarget names are not found because they have not been configured

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    for (const auto& rt : run_targets) {
+        run_target_by_name[rt.name] = &rt;
+    }
+
+    std::map<std::string, const ComponentConfig*> component_by_name;
+    for (const auto& comp : config.components()) {
+        component_by_name[comp.name] = &comp;
+    }
+
+    std::function<void(const std::string&, std::vector<uint32_t>&, std::set<std::string>&)> resolve_depends;
```
NicolasFussberger: Maybe it could make sense to have a dedicated Visitor class for this

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
NicolasFussberger: Maybe we keep the current validation in the python script and just put asserts in the code for these error cases

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+            resolve_depends(dep_name, fallback_state.process_indexes_, visited);
+        }
+        states.push_back(std::move(fallback_state));
+    }
+
+    return states;
+}
+
+void ConfigurationAdapter::resolveDependencyIndexes(std::vector<OsProcess>& processes) {
+    for (auto& proc : processes) {
```
NicolasFussberger, line 294: Is this needed?

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
NicolasFussberger: should we build these maps once during initialize and store them as private member?

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -0,0 +1,428 @@
...
+    return nullptr;
+}
+
+std::optional<const ProcessGroup*> ConfigurationAdapter::getProcessGroupByNameAndIndex(
+    const IdentifierHash& pg_name, const uint32_t index) const {
+    auto pg = getProcessGroupByID(pg_name);
+    if (pg) {
+        if (index < pg->processes_.size()) {
+            return pg;
+        }
```
NicolasFussberger, line 454: does not make sense to validate the index here

score/launch_manager/src/daemon/src/configuration/configuration_adapter.hpp
```diff
@@ -0,0 +1,138 @@
...
+    std::optional<const std::vector<uint32_t>*> getProcessIndexesList(
+        const score::lcm::internal::ProcessGroupStateID& process_group_state_id) const;
+    std::optional<const OsProcess*> getOsProcessConfiguration(const IdentifierHash& pg_name_,
+                                                               const uint32_t index) const;
+    std::optional<const DependencyList*> getOsProcessDependencies(const IdentifierHash& process_group_name,
+                                                                   const uint32_t index) const;
+
+    static const uint32_t kDefaultProcessExecutionError;
+    static uint32_t kDefaultProcessorAffinityMask();
+    static const int32_t kDefaultSchedulingPolicy;
```
NicolasFussberger: These defaults seem obsolete now

config/BUILD
```diff
@@ -68,3 +68,17 @@ config_setting(
     },
     visibility = ["//visibility:public"],
 )
+
+bool_flag(
+    name = "use_new_configuration",
+    build_setting_default = False,
+    visibility = ["//visibility:public"],
```
NicolasFussberger, line 75: Should we make this visibility private?
I assume this is an internal flag for development

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.hpp
```diff
@@ -121,9 +138,8 @@ class ProcessGroupManager final
     /// @return Pointer to the OSAL process interface.
     osal::IProcess* getProcessInterface();
 
-    /// @brief Gets the configuration manager.
-    /// @return Pointer to the ConfigurationManager object.
-    ConfigurationManager* getConfigurationManager();
+    /// @brief Gets the configuration.
+    /// @return Pointer to the configuration object.
```
NicolasFussberger: I guess this belongs to     ConfigurationType* getConfiguration(); from above

scripts/config_mapping/lifecycle_config.py
```diff
@@ -170,9 +172,20 @@ def dict_merge_recursive(dict_a, dict_b):
...
     )
 
-    for key in ("initial_run_target", "fallback_run_target"):
-        if key in config:
-            new_config[key] = config[key]
+    if "initial_run_target" in config:
+        new_config["initial_run_target"] = config["initial_run_target"]
+
+    if "fallback_run_target" in config:
+        fallback_defaults = {
```
NicolasFussberger, line 181: what was the reason for not having this in the default dictionary at the top?

scripts/config_mapping/lifecycle_config.py
```diff
@@ -186,6 +199,161 @@ def is_supervised(application_type):
...
+        }
+        if "max_memory_usage" in sandbox:
+            sandbox_out["max_memory_usage"] = sandbox["max_memory_usage"]
+        if "max_cpu_usage" in sandbox:
+            sandbox_out["max_cpu_usage"] = sandbox["max_cpu_usage"]
+
+        deployment = {
+            "ready_timeout": depl_cfg["ready_timeout"],
+            "shutdown_timeout": depl_cfg["shutdown_timeout"],
+            "bin_dir": depl_cfg.get("bin_dir", "/opt"),
```
NicolasFussberger: There are already defaults for bin_dir and working_dir.
I wonder if we should avoid repeating this here.

scripts/config_mapping/lifecycle_config.py
```diff
@@ -186,6 +199,161 @@ def is_supervised(application_type):
...
+    out["fallback_run_target"] = fb_out
+
+    out["alive_supervision"] = {
+        "evaluation_cycle": config.get("alive_supervision", {}).get(
+            "evaluation_cycle", 0.5
+        ),
+    }
+
+    watchdog_config = config.get("watchdog", {})
+    if watchdog_config and "device_file_path" in watchdog_config:
```
NicolasFussberger: I wonder why checking specifically for device_file_path?
Ideally, device_file_path and max_timeout should be mandatory in the schema if a watchdog is defined at all. But currently, it is not.

scripts/config_mapping/lifecycle_config.py
```diff
@@ -186,6 +199,161 @@ def is_supervised(application_type):
...
 
+SCHED_POLICY_MAP = {
+    "SCHED_OTHER": "OTHER",
+    "SCHED_FIFO": "FIFO",
+    "SCHED_RR": "RR",
+}
+
+
+def output_filename(input_filename):
+    return "launch_manager_config_gen.json"
```
NicolasFussberger: Can we make this depending on the input filename?

scripts/config_mapping/config.bzl
```diff
@@ -1,38 +1,109 @@
...
-    """Run a script to convert from the new config structure to the old, this 
-    creates 3 seperate configs.
-    """
-    script = ctx.executable.script
+def _launch_manager_config_impl(ctx):
     config = ctx.file.config
     schema = ctx.file.schema
+    script = ctx.executable.script
+    json_out_dir = ctx.attr.json_out_dir
+    use_new_format = ctx.attr.use_new_format
```
NicolasFussberger: Would it be possible to simplify the variant handling here to just always generate all the configs?
The naming should never conflict as we have fixed names for the old config.

Then we are able to have the test cases for the config generation valid for both old and new config until the old config is removed

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/AliveMonitorImpl.cpp
```diff
@@ -30,6 +30,19 @@ namespace saf
 namespace daemon
 {
 
+#ifdef USE_NEW_CONFIGURATION
+AliveMonitorImpl::AliveMonitorImpl(std::shared_ptr<score::lcm::IRecoveryClient> recovery_client,
```
paulquiring: Just not readable enough.

scripts/config_mapping/config.bzl
```diff
@@ -138,11 +185,39 @@ def launch_manager_config(
...
+    # Single combiner whose sources are selected by --//config:use_new_configuration
+    # Always outputs to `flatbuffer_out_dir` (e.g. "etc") so the launch_manager
+    # binary can find its config at "etc/<file>.bin" regardless of format.
     lm_config_combiner(
         name = name,
-        lm_config = ":" + name + "_lm_bin",
-        hm_config = ":" + name + "_hm_bin",
-        hm_core_config = ":" + name + "_hmcore_bin",
+        srcs = select({
+            "//config:lm_use_new_configuration": [":" + name + "_new_lm_bin"],
```
NicolasFussberger: When calling this from another repo (i.e. score_reference_integration) we will need this change to make it work

```suggestion
            Label("//config:lm_use_new_configuration"): [":" + name + "_new_lm_bin"],

```

Additionally, I realize now that we need the feature flag to have public visibility for this to work. Otherwise, reference_integration fails with visibility error

scripts/config_mapping/config.bzl
```diff
@@ -138,11 +185,39 @@ def launch_manager_config(
...
+
+    config_basename = Label(config).name
+    dot_index = config_basename.rfind(".")
+    config_stem = config_basename[:dot_index] if dot_index != -1 else config_basename
+
+    serialize_buffer(
+        name = name + "_new_lm_bin",
+        data = ":" + new_gen_name,
+        schema = new_lm_schema,
+        output = name + "_serialized_new/" + config_stem + "_gen.bin",
```
NicolasFussberger: Could we remove the _gen suffix?
Otherwise you will have to know that this is added in order to state this in the "-c <config>" CLI command

config/BUILD
paulquiring: public is kept so that it can be tested outside of this module

NicolasFussberger (approved):

PR #264, opened 2026-06-25
Run C++ code formatter in 3 folders

LorantLipcsei-ETAS:
Run C++ code formatter in 3 folders : 

- examples
- score/health_monitor
- tests

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -38,7 +38,7 @@ IncludeCategories:
   - Regex: '^(<|").*(>|")$'
     Priority: 1
 IndentWidth: 4
-InsertNewlineAtEOF: true
+# InsertNewlineAtEOF: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 QualifierAlignment: Left
 CommentPragmas: '^.*A2Lfactory:'
--- a/.clang-format
+++ b/.clang-format
@@ -38,7 +38,7 @@ IncludeCategories:
   - Regex: '^(<|").*(>|")$'
     Priority: 1
 IndentWidth: 4
-# InsertNewlineAtEOF: true
+InsertNewlineAtEOF: true
 KeepEmptyLinesAtTheStartOfBlocks: true
 QualifierAlignment: Left
 CommentPragmas: '^.*A2Lfactory:'
--- a/examples/control_application/control.hpp
+++ b/examples/control_application/control.hpp
@@ -15,11 +15,12 @@
 
 #include <cstddef>
 
-struct RunTargetInfo {
+struct RunTargetInfo
+{
     char runTargetName[1024]{};
 };
 
-static constexpr char const* control_socket_path = "/sm_control";
+static constexpr const char* control_socket_path = "/sm_control";
 static constexpr std::size_t control_socket_capacity = 32;
 
 #endif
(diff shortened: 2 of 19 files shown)


.clang-format
```diff
@@ -38,7 +38,7 @@ IncludeCategories:
   - Regex: '^(<|").*(>|")$'
     Priority: 1
 IndentWidth: 4
-InsertNewlineAtEOF: true
+# InsertNewlineAtEOF: true
```
NicolasFussberger: Why was this commented out?

LorantLipcsei-ETAS: I changed it because that line gave an unknown attribute error and searching for it revealed that this was encountered by others as well due to usually outdated clang-format ( under version 12, but ours is above 14 ).

pawelrutkaq: Shall not be done. Thats not our problem - one need to use correct clang format.

pawelrutkaq (changes requested):
Shall we integrate https://github.com/eclipse-score/tooling/pull/294. since if we have formatting errors now, does CI job detect formatting issues ? If not, we shall fix CI too with this PR.

.clang-format
NicolasFussberger: Devcontainer uses clang-format version 20.1.8 and the .clang-format file works for this version. So I also think this change has to be reverted.

NicolasFussberger:
> Shall we integrate [eclipse-score/tooling#294](https://github.com/eclipse-score/tooling/pull/294). since if we have formatting errors now, does CI job detect formatting issues ? If not, we shall fix CI too with this PR.

I think this would be good to use. Otherwise formatting will diverge again and again if this is not being enforced.

.clang-format
LorantLipcsei-ETAS: I will fix this and revert the change in the .clang-format file.

LorantLipcsei-ETAS: I reverted the change in the .clang-format file and re-ran the formatting using clang-format 20.1.8.

NicolasFussberger:
> > Shall we integrate [eclipse-score/tooling#294](https://github.com/eclipse-score/tooling/pull/294). since if we have formatting errors now, does CI job detect formatting issues ? If not, we shall fix CI too with this PR.
> 
> I think this would be good to use. Otherwise formatting will diverge again and again if this is not being enforced.

@LorantLipcsei-ETAS  can you create a separate ticket for this?

LorantLipcsei-ETAS:
> > > Shall we integrate [eclipse-score/tooling#294](https://github.com/eclipse-score/tooling/pull/294). since if we have formatting errors now, does CI job detect formatting issues ? If not, we shall fix CI too with this PR.
> > 
> > 
> > I think this would be good to use. Otherwise formatting will diverge again and again if this is not being enforced.
> 
> @LorantLipcsei-ETAS can you create a separate ticket for this?

Sure, here is the link to the ticket : https://github.com/eclipse-score/lifecycle/issues/268.

LorantLipcsei-ETAS:
@paulquiring took over the WI ( https://dev.azure.com/bosch-denso-common/VRTE-Adaptive/_workitems/edit/98344 ) so he will finish this PR.

paulquiring:
Closed, required tooling is introduced in https://github.com/eclipse-score/lifecycle/pull/313. After that the code base will be formatted in one go.

PR #267, opened 2026-06-26
QNX8 workflow with unit test execution

paulquiring:
Update qnx8 workflow to run also tests on qnx8


.github/workflows/build_qnx8.yml
```diff
@@ -15,27 +15,38 @@ name: QNX8 Build
...
   push:
     branches:
       - main
+  merge_group:
+    types: [checks_requested]
+  workflow_call:
 jobs:
   qnx-build:
-    uses: eclipse-score/cicd-workflows/.github/workflows/qnx-build.yml@c1c90b1a82a1fab0fc202979dde6686b2162d5a8 # v0.0.0
+    name: Build and Test ${{ matrix.bazel-config }}
```
NicolasFussberger: I am not sure if test execution for qnx is sufficiently setup in the repo. I would assume this requires spawning a qnx qemu machine that would require more setup in the repo?

paulquiring: I'm not aware that we do something special for QNX in baselibs. It should be very simple to just start tests with QEMU, everything is statically linked so no deployment of libraries is required.

.github/workflows/build_qnx8.yml
```diff
@@ -15,27 +15,38 @@ name: QNX8 Build
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - bazel-config: x86_64-qnx
+            bazel-test-target: >-
+              //score/...
```
NicolasFussberger: We should also include //examples/... there is also an automated demo scenario

.github/workflows/build_qnx8.yml
NicolasFussberger: Need to support https://github.com/eclipse-score/qnx_unit_tests in our MODULE.bazel and bazel config file.
We can check how baselibs does it https://github.com/eclipse-score/baselibs/blob/769bd7958690a833e14b6ef50cca56f417478c7d/.bazelrc#L76

.github/workflows/build_qnx8.yml
paulquiring: I tried running this. I see this error
```bash
/proc/boot/run_test.sh[18]: /persistent/unit_tests/cc_test_qnx: No such file or directory
```
The manual points out that this works with unit tests.
https://github.com/eclipse-score/qnx_unit_tests#how-it-works

MODULE.bazel
```diff
@@ -78,8 +78,34 @@ use_repo(
     "score_gcc_aarch64_toolchain",
     "score_gcc_x86_64_toolchain",
     "score_qcc_aarch64_toolchain",
+    "score_qcc_aarch64_toolchain_pkg",
     "score_qcc_x86_64_toolchain",
+    "score_qcc_x86_64_toolchain_pkg",
+)
+
+# UTs on QNX
+
```
paulquiring: remove blank

MODULE.bazel
```diff
@@ -78,8 +78,34 @@ use_repo(
...
+bazel_dep(name = "score_qnx_unit_tests", version = "0.2.0", dev_dependency = True)
+bazel_dep(name = "score_rules_imagefs", version = "0.0.3", dev_dependency = True)
+
+imagefs = use_extension("@score_rules_imagefs//extensions:imagefs.bzl", "imagefs", dev_dependency = True)
+imagefs.toolchain(
+    name = "score_qnx_x86_64_ifs_toolchain",
+    sdp_to_import = "@score_qcc_x86_64_toolchain_pkg",
+    sdp_version = "8.0.0",
```
paulquiring, line 94: 8.0.4?

paulquiring: Leeds to the error:
ERROR: /workspaces/lifecycle/score/launch_manager/src/alive/src/details/BUILD:15:8: While resolving toolchains for target //score/launch_manager/src/alive/src/details:AliveImpl_UT (5b91a44): invalid registered toolchain '@score_qnx_x86_64_ifs_toolchain//:ifs-x86_64-qnx-sdp_8.0.0': no such target '@@score_rules_imagefs++imagefs+score_qnx_x86_64_ifs_toolchain//:ifs-x86_64-qnx-sdp_8.0.0': target 'ifs-x86_64-qnx-sdp_8.0.0' not declared in package '' defined by /var/cache/bazel/7ec3dfcf57789bc5613ff23c5dcaffec/external/score_rules_imagefs++imagefs+score_qnx_x86_64_ifs_toolchain/BUILD (did you mean ifs-x86_64-qnx-sdp_8.0.4?)

Looks like this number has to match the number in the bazelrc e.g.
`aarch64-qnx-sdp_8.0.0`

NicolasFussberger (approved):

PR #269, opened 2026-06-29
deps: use baselibs instead of baselibs_rust

arkjedrz:


pawelrutkaq (approved):
Lock file needs update

Issue #215, opened 2026-05-27
OSAL Return Results Ignored


MaciejKaszynski:
### What

There are many cases where calls in the osal component return an OsalReturn but the caller doesn't check the result.

We should set all return types that can error in the Osal component as `[[nodiscard]]` and fix the warnings.
We could also set the actual `OsalReturnType` as `[[nodiscard]]` either is fine.

### Acceptance Criteria (DoD)

- All methods in the osal with errors are marked with `[[nodiscard]]`
- All warnings for no discard are fixed

thatguysumeet:
Hi I would like to contribute to issue #215

NicolasFussberger:
Great to hear! Let us know if you need any help to get started.

MaciejKaszynski:
OSAL return type is marked as `[[nodiscard]]` https://github.com/eclipse-score/lifecycle/blob/67b4e30906e2f3befa7f5c9bc11c5022ac8db066/score/launch_manager/src/daemon/src/osal/return_types.hpp#L51 No extra warnings in the build

PR #271, opened 2026-06-30
Improve nodiscard adoption for OSAL call sites

thatguysumeet:
Fix the issue : #215

Improvement using [[nodiscrd]] for OSAL return types.


score/launch_manager/src/daemon/src/alive_monitor/details/ifappl/MonitorIfDaemon_UT.cpp
```diff
@@ -74,7 +74,7 @@ struct MonitorIfDaemonFixture
     /// Initialize the IPC server so that peek/pop/hasOverflow use real shared memory.
     void initIpc()
     {
-        ipcServer.init(makeUniqueIpcName());
+        static_cast<void>(ipcServer.init(makeUniqueIpcName()));
```
MaciejKaszynski: I think ASSERT_TRUE would be nice here

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
```diff
@@ -94,7 +94,7 @@ ControlClientImpl::ControlClientImpl(std::function<void(const score::lcm::Execut
 
     ipc_channel_ = score::lcm::internal::ControlClientChannel::initializeControlClientChannel();
 
-    ipc_request_semaphore_.init(1U, false);
+    static_cast<void>(ipc_request_semaphore_.init(1U, false));
```
MaciejKaszynski: Would it make sense to `assert` here? otherwise we are in an undefined state

MaciejKaszynski (changes requested):

score/launch_manager/src/daemon/src/alive_monitor/details/ifappl/MonitorIfDaemon_UT.cpp
thatguysumeet: If I understood correctly  `ASSERT_TRUE(ipcServer.init(makeUniqueIpcName()) == ifappl::CheckpointIpcServer::EIpcInitResult::kOk)
            << "CheckpointIpcServer init failed";`

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
thatguysumeet: If I understood correctly `const auto init_result = ipc_request_semaphore_.init(1U, false);
    assert((score::lcm::internal::osal::OsalReturnType::kSuccess == init_result) &&
           "ControlClient semaphore initialization failed");`

MaciejKaszynski: Actually there are baselibs asserts, think we should be using it `SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE` since this would be useful info for user. https://github.com/eclipse-score/baselibs/blob/866daca52e8d886a04628be138800d103a5c015b/score/language/futurecpp/include/score/assert.hpp

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
```diff
@@ -107,7 +112,7 @@ ControlClientImpl::~ControlClientImpl() noexcept {
         ipc_response_thread_->join();
     }
 
-    ipc_request_semaphore_.deinit();
+    static_cast<void>(ipc_request_semaphore_.deinit());
```
MaciejKaszynski: I think actually doing a warning log in case of failure would be good.

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
```diff
@@ -284,7 +289,7 @@ score::concurrency::InterruptibleFuture<void> ControlClientImpl::SendIpcMessage(
         }
 
         // we definitely shouldn't forget to release semaphore
-        ipc_request_semaphore_.post();
+        static_cast<void>(ipc_request_semaphore_.post());
```
MaciejKaszynski: Looking at this again, I am a bit apprehensive with this one. 

Looking at the possible errors we can have `EINVAL` for which I think we can assert that this is not returned as if an invalid sempahore is used this would be a logical error. And `EOVERFLOW` for which I would want an error log as it probably means that there is some problem with LCM being stuck.

MaciejKaszynski (changes requested):

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
MaciejKaszynski: Same thing with the change on Line 383

thatguysumeet: Address review comments.

MaciejKaszynski (approved):

thatguysumeet:
@MaciejKaszynski Can I run the same pipeline locally ? 
Or It's a CI thing ?

NicolasFussberger:
> @MaciejKaszynski Can I run the same pipeline locally ? Or It's a CI thing ?

I retriggered the pipeline. Most likely this was just a sporadic failure. Unfortunately, the qnx pipeline very often runs into some connection timeout when downloading from qnx servers.

NicolasFussberger (approved):

PR #286, opened 2026-07-06
Extend cpu affinity mask to support up to 64 cores

shegazyy:
The cpu_mask_ field in OsalConfig and the setaffinity() function used uint32_t, limiting process affinity to at most 32 cores. Systems with more than 32 cores would silently use only the lower 32 bits.

Change the mask type to uint64_t throughout the affected chain:
- OsalConfig::cpu_mask_ (iprocess.hpp)
- setaffinity() signature (set_affinity.hpp and both platform impls)
- kDefaultProcessorAffinityMask() return type (configuration_manager)
- kDefaultNumCores cap raised from 32 to 64 (num_cores.hpp)

On Linux the loop variable and shift literal are updated to uint64_t / 1ULL to avoid the undefined behaviour for i >= 32 noted in the original review discussion (PR #113).

On QNX the 32-bit ThreadCtl(_NTO_TCTL_RUNMASK_GET_AND_SET) is replaced with _NTO_TCTL_RUNMASK64_GET_AND_SET which accepts a uint64_t mask.

kDefaultProcessorAffinityMask() guards against the n == 64 edge case (where 1ULL << 64 would itself be UB) by returning UINT64_MAX instead.

The XML configuration reader is updated to use std::stoull so the full 64-bit coremask value is parsed without truncation.

Closes #122


shegazyy:
Hey @NicolasFussberger 🙏 ready for review whenever you have a moment!

NicolasFussberger:
> Hey @NicolasFussberger 🙏 ready for review whenever you have a moment!

Thanks! I requested @SimonKozik to do a review, he already had a closer look at this some time in the past.

SimonKozik (approved):
Looks OK

NicolasFussberger:
The qnx8 pipeline fails with 

```
 score/launch_manager/src/daemon/src/osal/details/qnx/set_affinity.cpp: In function 'int32_t score::lcm::internal::osal::setaffinity(uint64_t)':
score/launch_manager/src/daemon/src/osal/details/qnx/set_affinity.cpp:30:27: error: '_NTO_TCTL_RUNMASK64_GET_AND_SET' was not declared in this scope
   30 |     return 0 == ThreadCtl(_NTO_TCTL_RUNMASK64_GET_AND_SET, &cpumask) ? 0 : -1;
      |  
```

I also could not find a reference to _NTO_TCTL_RUNMASK64_GET_AND_SET.

But I think also the runmask support in qnx8 changed significantly with the introduction of cpu clusters.
Do you have any suggestion how to proceed here?

NicolasFussberger (changes requested):
see above

shegazyy:
Thanks for catching this @NicolasFussberger 🙏 turns out `_NTO_TCTL_RUNMASK64_GET_AND_SET` is a QNX 7.x constant that got removed in QNX 8 when the CPU cluster model changed. Added an `#ifdef` guard so QNX 7 keeps the 64-bit path and QNX 8 falls back to the 32-bit mask (functionally no worse than before this PR). Full 64-core support on QNX 8 with the cluster API can be a follow-up issue. Should fix the QNX 8 build!

score/launch_manager/src/daemon/src/osal/details/qnx/set_affinity.cpp
```diff
@@ -13,18 +13,30 @@
...
+namespace osal
+{
 
-int32_t setaffinity(uint32_t cpumask) noexcept(true) {
-    return 0 == ThreadCtl(_NTO_TCTL_RUNMASK_GET_AND_SET, &cpumask) ? 0 : -1;
+int32_t setaffinity(uint64_t cpumask) noexcept(true)
+{
+#ifdef _NTO_TCTL_RUNMASK64_GET_AND_SET
+    return 0 == ThreadCtl(_NTO_TCTL_RUNMASK64_GET_AND_SET, &cpumask) ? 0 : -1;
+#else
```
NicolasFussberger: Do you require this to work on qnx7?
I am asking because Score does only support qnx8 and we do not have any means to verify this code for qnx7 at the moment

shegazyy:
Hey @NicolasFussberger 🙏 reworked it to use `_NTO_TCTL_RUNMASK_GET_AND_SET_INHERIT` with the multi-cluster runmask struct, so full 64-core support on QNX 8 with no QNX 7 dead code. Ready for another look!

score/launch_manager/src/daemon/src/osal/details/qnx/set_affinity.cpp
```diff
@@ -13,18 +13,42 @@
...
+        int size;
+        unsigned runmask[kSize];
+        unsigned inherit_mask[kSize];
+    } tm{kSize, {}, {}};
+
+    for (uint64_t i = 0U; i < 64U; ++i)
+    {
+        if (cpumask & (1ULL << i))
+        {
+            RMSK_SET(static_cast<int>(i), tm.runmask);
```
NicolasFussberger, line 45: I tried this out and the code was working for > 32 cpus 👍

I wonder if we should also set the inheritmask here, otherwise only the main thread of the child process gets the affinity. Any further threads created by the child process will not get this affinity.

```suggestion
            RMSK_SET(static_cast<int>(i), tm.runmask);
             RMSK_SET(static_cast<int>(i), tm.inherit_mask);
```

This is what I assume is the wanted behaviour from an integrators point or view and I think this is also the behavior of the linux implementation of this method

NicolasFussberger (dismissed):

shegazyy:
Hi @NicolasFussberger rebased to the latest main 🙏

NicolasFussberger:
> Hi @NicolasFussberger rebased to the latest main 🙏

Thank you!
Could you also adapt the cpu mask datatype in score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp

We are in the progress of refactoring the configuration and getting rid of the legacy configuration that is currently in place.
The configuration_adapter.cpp implements a temporary adapter that still provides the old interface to the current code, so that we can do the migration in multiple steps.

NicolasFussberger (approved):

PR #291, opened 2026-07-07
Add safety plan - lifecycle

aschemmel-tech:
Refers: score/#3036


PandaeDo (dismissed):
Compared with module_templates I see some differences. Especially the index.rst file in the root. There are some useful informations in it. But I still don't like the way how it's implemented. It raises some conflicts. So you might place the content to another place and we can enforce a discussion to change the definitions there.

aschemmel-tech:
> Compared with module_templates I see some differences. Especially the index.rst file in the root. There are some useful informations in it. But I still don't like the way how it's implemented. It raises some conflicts. So you might place the content to another place and we can enforce a discussion to change the definitions there.

I see your point, but the task was not to take over the module template (including its "root" documentation) but to create the safety plan. Also this is not really a safety work product. Unifying module entry point documentation would be needed to be discussed in the PL/FT team lead snc on Monday.

NicolasFussberger:
All changes look good to me.
I would just like to wait with merging until we have created the release tag (probably today), to not introduce any last minute changes.

NicolasFussberger (dismissed):

ramceb (approved):
LGTM

PR #293, opened 2026-07-08
Make `OsalReturnType` `nodiscard` and handle remaining warnings

### LINKED ISSUE #282: Fix Remaining OSAL
### What

Set the `OsalReturnType` as `[[nodiscard]]`

```cpp
enum class [[nodiscard]] OsalReturnType
...
```

Then fix all remaining errors. Some can be seen after PR https://github.com/eclipse-score/lifecycle/pull/271

e.g.

- `iprocess.hpp:113`
- `process_group_manager.cpp:369`
- `process_group_manager.cpp:165`
- ...

### Acceptance Criteria (DoD)

No Warnings with `[[nodiscard]]` on `OsalReturnType`

### How

_No response_

danth:
Fixes #282.

Requires #292 to be merged first.

### DIFF
--- a/score/launch_manager/src/daemon/src/control/control_client_channel.cpp
+++ b/score/launch_manager/src/daemon/src/control/control_client_channel.cpp
@@ -32,34 +32,40 @@ void ControlClientChannel::initialize()
 {
     request_.empty_.store(true);
     response_.empty_.store(true);
-    nudge_LM_Handler_.init(0U, true);
+
+    const auto result = nudge_LM_Handler_.init(0U, true);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result == osal::OsalReturnType::kSuccess,
+                                            "ControlClientChannel semaphore init failed");
+
     initial_result_count_ = 0U;
+
     LM_LOG_DEBUG() << "ControlClientChannel initialized";
 }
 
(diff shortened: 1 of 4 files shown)


score/launch_manager/src/daemon/src/control/control_client_channel.cpp
```diff
@@ -32,34 +32,37 @@ void ControlClientChannel::initialize()
...
     {
         LM_LOG_DEBUG() << "Failed to send response: response is not empty.";
+        return false;
     }
 
-    return result;
+    response_.msg_ = msg;
+    response_.empty_ = false;
+
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(nudge_LM_Handler_.post() == osal::OsalReturnType::kSuccess,
```
NicolasFussberger: I think the logic should be outside the assert, so if someone is not compiling with asserts the software still works as before.

```suggestion
const auto result = nudge_LM_Handler_.post();
SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result == osal::OsalReturnType::kSuccess,
```

Same for the other edits

danth: 👍🏼 did not realise the default asserts could be disabled

NicolasFussberger (approved):

PR #294, opened 2026-07-08
Treat warnings as errors

danth:
This ensures that warnings are not ignored, by causing the build to fail.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -23,10 +23,15 @@ build --java_runtime_version=remotejdk_17
 build --tool_java_runtime_version=remotejdk_17
 build --credential_helper=*.qnx.com=%workspace%/scripts/internal/qnx_creds.py
 
+# Treat warnings as errors, except in external code
+build --copt=-Wall
+build --copt=-Wextra
+build --copt=-Werror
+build --per_file_copt=external/.*@-Wno-error
+
 # Common test flags for all platforms
 test --test_output=errors
 test --@score_baselibs//score/json:base_library=nlohmann
-test --cxxopt=-Wno-deprecated-declarations
 
 # Coverage configuration for C++
 coverage --features=coverage
--- a/examples/c_supervised_app/BUILD
+++ b/examples/c_supervised_app/BUILD
@@ -14,9 +14,8 @@ load("@rules_cc//cc:defs.bzl", "cc_binary")
 
 cc_binary(
     name = "c_supervised_app",
-    srcs = [
-        "main.c",
-    ],
+    srcs = ["main.c"],
+    copts = ["-std=gnu11"],
     linkopts = select({
         "@platforms//os:qnx": [
             "-lsocket",
--- a/examples/c_supervised_app/main.c
+++ b/examples/c_supervised_app/main.c
@@ -14,7 +14,7 @@
 #include <signal.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <unistd.h>
+#include <time.h>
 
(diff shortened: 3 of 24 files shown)


danth:
The existing warnings should be fixed by #293

NicolasFussberger:
I wonder if we should add further warnings for compilation, e.g. like COM does https://github.com/eclipse-score/communication/blob/570e34715b2d3156d2ae2cc4686be29a8c92f3ef/MODULE.bazel#L97
Baselibs does similar things.
This would give some additional benefit to treating warnings as errors compared to now when we do not have many warnings active.

But we should also avoid that users will inherit those compiler flags from our targets.
Currently I am not sure what is the behavior.

danth:
> I wonder if we should add further warnings for compilation

I've enabled `-Wall` and `-Wextra`, except for a couple of warnings present in baselibs.

There is also `-Wpedantic`, but it seems like there would be a lot of changes needed to build successfully with that.

> But we should also avoid that users will inherit those compiler flags from our targets. Currently I am not sure what is the behavior.

Flags set in `.bazelrc` are only applied when Bazel runs inside this repository, if that is what you mean: https://bazel.build/run/bazelrc#bazelrc-file-locations

NicolasFussberger (approved):

NicolasFussberger:
QNX build fails with the following error:
```
external/googletest+/googletest/src/gtest-death-test.cc:1283:34: error: missing initializer for member 'inheritance::pgroup' [-Werror=missing-field-initializers]
 1283 |   struct inheritance inherit = {0};
      |
```

I wonder if this will become an issue that new compiler warning in any dependency may break our build in the future. Is there a possibility to limit the warnings only to our own code?

NicolasFussberger (changes requested):
See above

danth:
Just pushed a new implementation which should not apply the flags to our dependencies, nor to anything depending on us.

However it means we have to call a different rule which wraps the standard `cc_library` / `cc_binary` to add our options.

config/common_opts.bzl
```diff
@@ -1,30 +0,0 @@
...
-# See the NOTICE file(s) distributed with this work for additional
-# information regarding copyright ownership.
-#
-# This program and the accompanying materials are made available under the
-# terms of the Apache License Version 2.0 which is available at
-# https://www.apache.org/licenses/LICENSE-2.0
-#
-# SPDX-License-Identifier: Apache-2.0
-# *******************************************************************************
-SANITIZER_CXXOPTS = select({
```
NicolasFussberger, line 13: Where those sanitizer flags now longer needed due to having this in the bazel rc now?

danth: I think they were only used by the old `cc_library_with_common_opts` which was not actually called from anywhere.

NicolasFussberger:
> Just pushed a new implementation which should not apply the flags to our dependencies, nor to anything depending on us.
> 
> However it means we have to call a different rule which wraps the standard `cc_library` / `cc_binary` to add our options.

I came across --per_file_copt in .bazelrc
Seems like it could be used to set the copts only for files in the current repo.
`build --per_file_copt=//.*\.(cc|cpp|cxx|h|hpp)$@-Wall,-Wextra,-Werror`
(external files start with @repo//...., and so they are not matched).
I have not tried it yet, but if that works I think it would be more lightweight than wrapping all binary/library targets.

danth:
Yes that would be cleaner if it works, I will try

NicolasFussberger (dismissed):

NicolasFussberger (approved):

PR #297, opened 2026-07-08
hmon: use single Arc for heartbeat monitor

arkjedrz:
Replace previous approach with two Arcs.

### DIFF
--- a/score/health_monitor/src/rust/heartbeat/heartbeat_monitor.rs
+++ b/score/health_monitor/src/rust/heartbeat/heartbeat_monitor.rs
@@ -616,6 +616,42 @@ mod tests {
     }
 
     #[test]
+    #[cfg_attr(miri, ignore)]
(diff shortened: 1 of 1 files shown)


danth (commented):
Previously, calling `get_eval_handle` a second time for the same monitor would return a new handle with a separate timestamp. After this change, the same timestamp is shared between all handles.

I'm not sure what the expected behaviour should be, but it would be good to add a test for this either way.

arkjedrz:
@danth, thanks for taking a look.
Each monitor should have `get_eval_handle` used once - by `HealthMonitor::get_monitor`, so the issue would never be observed using public APIs. This method is returning handles to the same monitor, so it was not a right choice to have a stateful handle in the first place. I added a test case for that, it is not passing for `main` branch.
BTW the context of this change is to enable protected memory allocator, having a double `Arc` unnecessarily complicates things.

pawelrutkaq (approved):

PR #299, opened 2026-07-08
Register assertion handler to print failure details to stderr

shegazyy:
Without a registered handler, SCORE_LANGUAGE_FUTURECPP_ASSERT failures call std::abort silently, giving no indication of what failed or where.

Register a handler at daemon startup that writes the failed condition, optional message, file, line, and function to stderr before abort.

Closes: #296

PR #303, opened 2026-07-09
Avoid deadlocks after fork

danth:
Fork only copies the current thread, so from that point on we could deadlock if we access anything that was locked by another thread at the time of the fork.

According to `man 2 fork` we should only use async signal safe functions between `fork` and `execve`.

### DIFF
--- a/score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
+++ b/score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
@@ -1,0 +1,130 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include <charconv>
+#include <unistd.h>
+#include <array>
+#include <cstdlib>
+#include <string_view>
+
+#if (_GNU_SOURCE && __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 32) || __QNXNTO__
+#include <cstring>
+#endif
+
+#ifndef SIGNAL_SAFE_LOG_HPP_INCLUDED
+#define SIGNAL_SAFE_LOG_HPP_INCLUDED
+
+namespace
+{
+
+/// @brief Utility for building log messages in an async signal safe context.
+template <std::size_t C>
+class signal_safe_buffer
+{
+  public:
+    /// @brief Append the given string to the buffer.
+    template <typename T, std::enable_if_t<std::is_convertible_v<T, std::string_view>, bool> = true>
+    void append(const T& value)
+    {
+        const auto string = std::string_view(value);
+
+        std::size_t length = std::min(string.length(), buffer_.size() - end_);
+
+        for (std::size_t index = 0; index < length; ++index)
+        {
+            buffer_[end_++] = string[index];
+        }
+    }
+
+    /// @brief Append the given integer to the buffer.
+    template <typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
+    void append(const T& value)
+    {
+        const auto [pointer, error] = std::to_chars(&buffer_[end_], &buffer_.back(), value);
+
+        if (error == std::errc{})
+        {
+            end_ = pointer - &buffer_.front();
+        }
+    }
+
+    /// @brief Write out the buffer to standard error.
+    /// @return Whether the write succeeded. True means the whole message was
+    ///         written. False means nothing, or only part was written.
+    [[nodiscard]] bool flush()
+    {
+        while (start_ < end_)
+        {
+            const auto written = write(STDERR_FILENO, &buffer_[start_], end_ - start_);
+
+            if (written >= 0)
+            {
+                start_ += written;
+            }
+            else if (errno != EINTR)
+            {
+                return false;
+            }
+        }
+
+        // Rewind the indices in case the buffer is used again.
+        start_ = 0;
+        end_ = 0;
+
+        return true;
+    }
+
+  private:
+    std::array<char, C> buffer_ = {};
+    std::size_t start_ = 0;
+    std::size_t end_ = 0;
+};
+
+}  // namespace
+
+namespace score::lcm::internal
+{
+
+template <typename... T>
+[[nodiscard]] bool signal_safe_log(const T&... values)
+{
+    const int original_errno = errno;  // For reentrancy
+
+    signal_safe_buffer<1024U> buffer = {};
+    (buffer.append(values), ...);
+    buffer.append("\n");
+    const bool written = buffer.flush();
+
+    errno = original_errno;  // For reentrancy
+
+    return written;
+}
+
+template <typename... T>
+[[nodiscard]] bool signal_safe_log_errno(int log_errno, const T&... values)
+{
+#if _GNU_SOURCE && __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 32
+    // strerrordesc_np is documented as async signal safe.
+    return signal_safe_log(values..., " (", strerrordesc_np(log_errno), ")");
+#elif __QNXNTO__
+    // QNX strerror is documented as async signal safe.
+    return signal_safe_log(values..., " (", strerror(log_errno), ")");
+#else
+    // POSIX strerror is not async signal safe. Log the raw number instead.
+    return signal_safe_log(values..., " (errno ", log_errno, ")");
+#endif
+}
+
+}  // namespace score::lcm::internal
+
+#endif
--- a/score/launch_manager/src/daemon/src/common/signal_safe_log_UT.cpp
+++ b/score/launch_manager/src/daemon/src/common/signal_safe_log_UT.cpp
@@ -1,0 +1,111 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include <gtest/gtest.h>
+#include <gmock/gmock.h>
+
+#include "score/mw/launch_manager/common/signal_safe_log.hpp"
+
+using namespace testing;
+using score::lcm::internal::signal_safe_log;
+using score::lcm::internal::signal_safe_log_errno;
+
+class signal_safe_log_test : public ::testing::Test
+{
+  protected:
+    void SetUp() override
+    {
+        RecordProperty("TestType", "interface-test");
+        RecordProperty("DerivationTechnique", "explorative-testing");
+    }
+};
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_simple)
+{
+    RecordProperty(
+        "Description", "This test verifies that signal_safe_log successfully writes a simple message to stderr.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log("Hello world"));
+
+    EXPECT_EQ(testing::internal::GetCapturedStderr(), "Hello world\n");
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_simple_with_errno)
+{
+    RecordProperty(
+        "Description",
+        "This test verifies that signal_safe_log successfully writes a simple message and errno to stderr.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log_errno(1, "Hello world"));
+
+    EXPECT_THAT(testing::internal::GetCapturedStderr(), ContainsRegex("^Hello world \\(.*\\)"));
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_formatted)
+{
+    RecordProperty(
+        "Description", "This test verifies that signal_safe_log successfully writes a formatted message to stderr.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log(12, " + ", 76, " = ", 88));
+    EXPECT_EQ(testing::internal::GetCapturedStderr(), "12 + 76 = 88\n");
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_formatted_with_errno)
+{
+    RecordProperty(
+        "Description",
+        "This test verifies that signal_safe_log successfully writes a formatted message and errno to stderr.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log_errno(1, 12, " + ", 76, " = ", 88));
+
+    EXPECT_THAT(testing::internal::GetCapturedStderr(), ContainsRegex("12 \\+ 76 = 88 \\(.*\\)"));
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_minimum_length)
+{
+    RecordProperty("Description", "This test verifies that signal_safe_log can output at least 1024 characters.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log(std::string(1023, 'A')));
+
+    const auto stderr = testing::internal::GetCapturedStderr();
+
+#if __QNXNTO__
+    EXPECT_EQ(stderr->size(), 1024);
+    EXPECT_EQ(stderr->back(), '\n');
+#else
+    EXPECT_EQ(stderr.size(), 1024);
+    EXPECT_EQ(stderr.back(), '\n');
+#endif
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_maximum_length)
+{
+    RecordProperty("Description", "This test verifies that signal_safe_log can output at most 1024 characters.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log(std::string(2048, 'A')));
+
+    const auto stderr = testing::internal::GetCapturedStderr();
+
+#if __QNXNTO__
+    EXPECT_EQ(stderr->size(), 1024);
+    EXPECT_EQ(stderr->back(), 'A');
+#else
+    EXPECT_EQ(stderr.size(), 1024);
+    EXPECT_EQ(stderr.back(), 'A');
+#endif
+}
--- a/score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
+++ b/score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
@@ -118,6 +118,7 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
             {
                 LM_LOG_ERROR() << "[New process] fcntl() at line" << __LINE__
                                << "failed:" << score::lcm::internal::errno_message(errno);
+                sysexit(EXIT_FAILURE);
             }
             break;
         case CommsType::kLaunchManager:
--- a/score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
+++ b/score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
@@ -22,6 +22,7 @@
 #include <limits.h>
 #include <signal.h>
 
+#include "score/mw/launch_manager/common/signal_safe_log.hpp"
 #include "score/mw/launch_manager/process_group_manager/iprocess.hpp"
 #include "score/mw/launch_manager/control/control_client_channel.hpp"
 #include "score/mw/launch_manager/common/log.hpp"
@@ -42,6 +43,8 @@ constexpr int kPosixSuccess = 0;
 namespace
 {
 
+using score::lcm::internal::signal_safe_log;
+using score::lcm::internal::signal_safe_log_errno;
 using score::lcm::internal::osal::CommsType;
 using score::lcm::internal::osal::IpcCommsSync;
 using score::lcm::internal::osal::sysexit;
@@ -52,8 +55,7 @@ void applyLimitOrDie(const int resource, const rlimit& limit, const std::string_
 {
     if (::setrlimit(resource, &limit) == -1)
     {
-        LM_LOG_FATAL() << "[New process] Failed to set rlimit " << rlimit_name << " "
-                       << score::lcm::internal::errno_message(errno);
+        static_cast<void>(signal_safe_log_errno(errno, "Failed to apply rlimit ", rlimit_name));
         sysexit(EXIT_FAILURE);
     }
 }
@@ -101,8 +103,8 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
         case CommsType::kReporting:
             if (-1 == fcntl(IpcCommsSync::sync_fd, F_SETFD, 0))
             {
-                LM_LOG_ERROR() << "[New process] fcntl() at line" << __LINE__
-                               << "failed:" << score::lcm::internal::errno_message(errno);
+                static_cast<void>(signal_safe_log_errno(errno,
+                    "fcntl at line ", __LINE__, " failed"));
                 sysexit(EXIT_FAILURE);
             }
             close(IpcCommsSync::control_client_handler_nudge_fd);
@@ -110,14 +112,14 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
         case CommsType::kControlClient:
             if (-1 == fcntl(IpcCommsSync::sync_fd, F_SETFD, 0))
             {
-                LM_LOG_ERROR() << "[New process] fcntl() at line" << __LINE__
-                               << "failed:" << score::lcm::internal::errno_message(errno);
+                static_cast<void>(signal_safe_log_errno(errno,
+                    "fcntl at line ", __LINE__, " failed"));
                 sysexit(EXIT_FAILURE);
             }
             if (-1 == fcntl(IpcCommsSync::control_client_handler_nudge_fd, F_SETFD, 0))
             {
-                LM_LOG_ERROR() << "[New process] fcntl() at line" << __LINE__
-                               << "failed:" << score::lcm::internal::errno_message(errno);
+                static_cast<void>(signal_safe_log_errno(errno,
+                    "fcntl at line ", __LINE__, " failed"));
                 sysexit(EXIT_FAILURE);
             }
             break;
@@ -125,8 +127,8 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
             // nothing to do here
             break;
         default:
-            LM_LOG_ERROR() << "[New process] at line" << __LINE__ << "unknown CommsType"
-                           << static_cast<std::int32_t>(param.shared_block->comms_type_);
+            static_cast<void>(signal_safe_log("at line ", __LINE__, " unknown CommsType ",
+                static_cast<std::int32_t>(param.shared_block->comms_type_)));
             sysexit(EXIT_FAILURE);
             break;
     }
@@ -142,15 +144,15 @@ void changeCurrentWorkingDirectory(const score::lcm::internal::osal::OsalConfig&
 
     if (config.executable_path_.size() >= string_size)
     {
-        LM_LOG_ERROR() << "[New process] executable path longer than" << string_size
-                       << "chars:" << config.executable_path_;
+        static_cast<void>(signal_safe_log("executable path longer than ",
+            string_size, " chars: ", config.executable_path_));
         sysexit(EXIT_FAILURE);
     }
 
     if (-1 == chdir(dirname(strncpy(path_copy, config.executable_path_.c_str(), string_size))))
     {
-        LM_LOG_ERROR() << "[New process] chdir(" << config.executable_path_
-                       << ") failed:" << score::lcm::internal::errno_message(errno);
+        static_cast<void>(signal_safe_log_errno(errno,
+            "chdir(", config.executable_path_, ") failed"));
         sysexit(EXIT_FAILURE);
     }
 }
@@ -174,8 +176,8 @@ void changeSecurityPolicy(const score::lcm::internal::osal::OsalConfig& config)
     {
         if (score::lcm::internal::osal::setSecurityPolicy(config.security_policy_.c_str()) != 0)
         {
-            LM_LOG_ERROR() << "[New process] changeSecurityPolicy(" << config.security_policy_
-                           << ") failed:" << score::lcm::internal::errno_message(errno);
+            static_cast<void>(signal_safe_log_errno(errno,
+                "changeSecurityPolicy(", config.security_policy_, ") failed"));
             sysexit(EXIT_FAILURE);
         }
     }
@@ -203,7 +205,8 @@ OsalReturnType IProcess::startProcess(ProcessID* pid, IpcCommsP* block, const Os
     {
         if (access(config->executable_path_.c_str(), X_OK) != 0)
         {
-            LM_LOG_ERROR() << "File does not exist or is not executable:" << config->executable_path_;
+            static_cast<void>(signal_safe_log(
+                "File does not exist or is not executable: ", config->executable_path_));
             return result;
         }
 
@@ -273,9 +276,9 @@ inline bool IProcess::setupComms(IpcCommsP& block, int& fd, const OsalConfig& co
     }
 
     static_cast<void>(snprintf(shm_name,
-                               static_cast<uint32_t>(score::lcm::internal::ProcessLimits::maxLocalBuffSize),
-                               "/ipc_shared_mem%u",
-                               shm_name_counter++));
+        static_cast<uint32_t>(score::lcm::internal::ProcessLimits::maxLocalBuffSize),
+        "/ipc_shared_mem%u",
+        shm_name_counter++));
 
     fd = shm_open(shm_name, O_CREAT | O_EXCL | O_RDWR, 0U);
 
@@ -362,7 +365,7 @@ OsalReturnType IProcess::setSchedulingAndSecurity(const OsalConfig& config)
     // setpgid will fail if called by a session lader (which LCMd is), so skip
     if (config.comms_type_ != osal::CommsType::kLaunchManager && 0 != setpgid(0, getpid()))
     {
-        LM_LOG_ERROR() << "setpgid() failed:" << score::lcm::internal::errno_message(errno);
+        static_cast<void>(signal_safe_log_errno(errno, "setpgid() failed"));
         retval = OsalReturnType::kFail;
     }
     // Set scheduling policy with sched_setscheduler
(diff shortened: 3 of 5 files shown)


danth:
Failure is due to https://www.githubstatus.com/incidents/cstx3v63mklm

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,149 @@
...
+    // Set process group id to be equal to the pid
+    // setpgid will fail if called by a session leader (which LCMd is), so skip
+    if (config.comms_type_ != osal::CommsType::kLaunchManager)
+    {
+        result = setpgid(0, getpid());
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result == 0, std::strerror(errno));
+    }
+
+    // Set scheduling policy with sched_setscheduler
+    /* RULECHECKER_comment(1, 1, check_union_object, "Union type defined in external library is used.", true) */
```
MaciejKaszynski:
```suggestion
```

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,149 @@
...
+    sched_param sch_param{};
+
+    sch_param.sched_priority = config.scheduling_priority_;
+
+    // TODO: Clamping should be done when the config is loaded
```
MaciejKaszynski: Can you make issues with all of these TODOs to keep track better, would be really cool if you also appended the link to the comment

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    result = sched_setscheduler(0, config.scheduling_policy_, &sch_param);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result != -1, std::strerror(errno));
+
+    result = osal::setaffinity(config.cpu_mask_);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result != -1, std::strerror(errno));
+
+    result = setgid(config.gid_);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result != -1, std::strerror(errno));
```
MaciejKaszynski: I think for all of the asserts it might not be too nice to just print the errno. User would probably not get any feedback on what happened.

Also would it make sense to use `SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE` so this assert will show up in a production build?

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    result = setuid(config.uid_);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(result != -1, std::strerror(errno));
+}
+
+void changeCurrentWorkingDirectory(const osal::OsalConfig& config)
+{
+    // Change current working directory to the same as the executable
+    constexpr size_t string_size = static_cast<size_t>(PATH_MAX);
```
MaciejKaszynski: just use PATH_MAX, I know it's copied and pasted but would be nice to fix :)

score/launch_manager/src/daemon/src/process_group_manager/details/process_group_manager.cpp
```diff
@@ -104,10 +105,9 @@ bool ProcessGroupManager::initialize()
         return false;
     }
 
-    if (launch_manager_config_ &&
-        OsalReturnType::kFail == IProcess::setSchedulingAndSecurity(launch_manager_config_->startup_config_))
+    if (launch_manager_config_)
```
MaciejKaszynski: Can you add a comment because I think it's hard to see that this is setting lcms own policies

MaciejKaszynski (changes requested):

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: Depends if we want to prioritize performance or error checking.

In the assertion message, the macro adds the file, line number, etc itself so we are just passing in `std::strerror(errno)` to get a description of the specific failure on top of that.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: We already have an issue #304, I will add a link to it here. Maybe we should also link back to these lines from the issue once it's merged into main.

MaciejKaszynski (dismissed):

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: Looking at the code again we have some things like the security policy, where it would not be ideal to simply ignore the fact that it was not set correctly. So I'm going to go for production asserts.

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -99,85 +69,53 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
...
     {
-        LM_LOG_ERROR() << "[New process] executable path longer than" << string_size
-                       << "chars:" << config.executable_path_;
-        sysexit(EXIT_FAILURE);
+        LM_LOG_ERROR() << "File does not exist or is not executable:" << config->executable_path_;
+        return false;
     }
 
-    if (-1 == chdir(dirname(strncpy(path_copy, config.executable_path_.c_str(), string_size))))
+    if (config->scheduling_priority_ < sched_get_priority_min(config->scheduling_policy_))
```
NicolasFussberger: Do you already want to move this to score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp?

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -99,85 +69,53 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
...
-            sysexit(EXIT_FAILURE);
+            SCORE_LANGUAGE_FUTURECPP_UNREACHABLE_MESSAGE(
+                ("unknown CommsType " + std::to_string(static_cast<std::int32_t>(param.shared_block->comms_type_)))
+                    .c_str());
             break;
     }
 }
 
-void changeCurrentWorkingDirectory(const score::lcm::internal::osal::OsalConfig& config)
+// TODO: Move this into the config loader
```
NicolasFussberger: I think only the scheduling priority range can be checked during config loading.
The executable may not be accessible yet at that time, in case the filesystem where the executable resides is only mounted later. So I think this access check has to be done at the time when we want to start the process.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+namespace score
+{
+
+namespace lcm
+{
+
+namespace internal
+{
+
```
NicolasFussberger: Are all of these methods only used btw. fork() and execve()?

If so I think it would be helpful to document this here at the top of the file and pointing out that only async signal safe functions must be called in this file.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,146 @@
...
+void setSchedulingAndSecurity(const osal::OsalConfig& config)
+{
+    int result;  // Used for return value from system calls
+
+    // Set process group id to be equal to the pid
+    // setpgid will fail if called by a session leader (which LCMd is), so skip
+    if (config.comms_type_ != osal::CommsType::kLaunchManager)
+    {
+        result = setpgid(0, getpid());
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(result == 0, std::strerror(errno));
```
NicolasFussberger: I think the strategy to use asserts here is now conflicting with this https://github.com/eclipse-score/lifecycle/pull/299
where asserts actually log to std::cerr which I guess is not safe to call btw. fork() and execve() either. Also the heap allocation from std::ostringstream is probably then a problem.

Do you have any suggestion how to resolve this?

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,146 @@
...
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(result != -1, std::strerror(errno));
+}
+
+void changeCurrentWorkingDirectory(const osal::OsalConfig& config)
+{
+    // Notice that this next static variable is duplicated by the fork() and so does not need
+    // any protection by a mutex although at first sight you may think it could need one.
+    static char path_copy[PATH_MAX + 1U] = {0};
+
+    // TODO: This should be validated when the config is loaded
```
NicolasFussberger: Would you like to move this already to the new config loading code or handle this separately?

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: We could allocate a fixed buffer for the assertion handler and call `write` directly, bypassing `std::cerr`? `write` is async signal safe.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.hpp
danth: That is the intention and would be worth documenting I think.

~~Although there only needs to be no locking, there are additional requirements to be fully async signal safe which I do not think are relevant here.~~ Edit- the man page for `fork` specifies that all requirements must be met

One of the functions is also used [here](https://github.com/etas-contrib/score_lifecycle/blob/e249b48458543ee79a759cfb78e44a1be372cc34/score/launch_manager/src/daemon/src/process_group_manager/details/process_group_manager.cpp#L119) which is a bit out of place

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: I've opened an issue #304, I would be happy to work on it but I think it should be a separate PR

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
danth: For the executable we could not check in advance and just detect it when `execve` fails

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
NicolasFussberger: It sounds like a good solution to me. What do you think @paulquiring ?

paulquiring: We need something like this as assertion handler. 

```c++
void my_signal_handler(int sig) {
    char buffer[128];
    // Format your string (snprintf is generally safe in practice though technically not guaranteed)
    int len = snprintf(buffer, sizeof(buffer), "Caught signal %d\n", sig);
    
    // write to stdout (file descriptor 1) is async-signal-safe
    write(STDOUT_FILENO, buffer, len);
}
```

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,157 @@
...
+    if (amount == 0)
+        return;
+
+    const struct rlimit limit{
+        .rlim_cur = amount,
+        .rlim_max = amount,
+    };
+
+    const auto result = ::setrlimit(resource, &limit);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
```
NicolasFussberger: I think we have the same async signal safe issue with the assert message, as the string concatenation may cause heap allocation which is not async signal safe

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -225,6 +160,18 @@ OsalReturnType IProcess::startProcess(ProcessID* pid, IpcCommsP* block, const Os
 
             if (*pid == kPosixSuccess)
             {
+                /*
+                 * From this point on no locks can be used.
```
paulquiring:
```suggestion
                 * From this point on only async signal safe functions can be     used.
```

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
NicolasFussberger: bazel assort macros terminate the process using std::abort(). Is this safe to terminate the process this way btw. fork() and execve()?

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
NicolasFussberger: snprintf is also not async signal safe 😅

paulquiring:
One thing we need to consider: 

```
fork
// only AS safe functions are allowed  
excev
```
the assert handler implementation as is looks quite okay for all non AS safe use cases. So we might want write an own function which does AS safe prints with write(STDERR_FILENO, msg, msg_len), msg is a fixed string to avoid allocations (e.g long string + std::string("something")).  Our own function does sysexit() after the print.

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
paulquiring: The comments below should also change accordingly.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
paulquiring: @NicolasFussberger and I discussed this a bit more see https://github.com/eclipse-score/lifecycle/pull/303#issuecomment-4989167100

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: `abort` is async signal safe in POSIX, not sure if `std::abort` is doing anything extra?

NicolasFussberger: I think here it is not so much about the signal safety, but other considerations what is being cleaned up: https://www.unixguide.net/unix/programming/1.1.3.shtml

Not sure though exactly what is the difference btw. _exit and std::abort, but If I remember correctly using _exit() is the recommended way to terminate a process btw. fork and execve in case of error

danth: We could remove the extra info and just use `strerror` as the message

danth: Some more info- `std::abort` and `abort` are one and the same. `abort` does no more cleanup than `_exit`. The main difference is that `abort` sends `SIGABRT` which triggers a core dump whereas `_exit` ends the process normally

danth: Update: `strerror` is not async signal safe either, there is `strerrordesc_np` but it's not standard

score/launch_manager/src/daemon/src/common/assertion_handler.cpp
```diff
@@ -0,0 +1,102 @@
...
+ */
+
+void stderr_print(const char* message)
+{
+    auto offset = 0;
+    const auto length = strlen(message);
+
+    while (offset < length)
+    {
+        const auto result = write(2, &message[offset], length - offset);
```
NicolasFussberger: Please use the constant for stderr. It may not be obvious to everyone what the magic number 2 means.

score/launch_manager/src/daemon/src/common/assertion_handler.cpp
```diff
@@ -0,0 +1,102 @@
...
+        else
+        {
+            offset += result;
+        }
+    }
+}
+
+void assertion_handler(const cpp::handler_parameters& params)
+{
+    stderr_print("Assertion failed:");
```
NicolasFussberger: I wonder if with this piece-by-piece writing we would end up with messages looking like this, as we are logging from multiple threads:

```
2026-07-09 05:59:18.088] [INFO] [launch_manager] 2026/07/09 05:59:18.6758083 62925824 000 ECU1 LM LM log debug verbose 1 Request sent. Waiting for acknowledgment...
[2026-07-09 05:59:18.088] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925828 000 ECU1 LM LM log debug verbose 6 Got kRunning for pid 68 ( control_client_mock ) process 0
Assertion failed:
[2026-07-09 05:59:18.089] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925828 000 ECU1 LM LM log debug verbose 5 startProcess for process 0 ( control_client_mock ) done
In
[2026-07-09 05:59:18.089] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925829 000 ECU1 
score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
LM LM log debug verbose 8 ProcessGroupManager::ControlClientHandler: got request kSetStateRequest ( 16 ) re state MainPG/run_target_complex_monitoring of PG MainPG 
[2026-07-09 05:59:18.089] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925830 000 ECU1 LM LM log debug verbose 6 Pending state for process group MainPG changed from to MainPG/run_target_complex_monitoring
[2026-07-09 05:59:18.089] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925830 000 ECU1 LM LM log debug verbose 9 Graph::setState changes from kInTransition to kCancelled for PG 0 ( MainPG ) 
[2026-07-09 05:59:18.090] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925830 000 ECU1 LM LM log debug verbose 1 Control Client handler nudged 
[2026-07-09 05:59:18.090] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925831 000 ECU1 LM LM log debug verbose 1 Request acknowledged. 
[2026-07-09 05:59:18.090] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925831 000 ECU1 LM LM log debug verbose 1 Control Client handler nudged
[2026-07-09 05:59:18.090] [INFO] [launch_manager] 2026/07/09 05:59:18.6758084 62925831 000 ECU1 LM LM log debug verbose 1 2026/07/09 05:59:18.6758084 62925831 000 ECU1 LM LM log fatal verbose 3 clock() at failed initial state transition: 29.370000 ms 
```

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
NicolasFussberger: I think, the additional information may be quite useful.
For example, having the errno to see if something was a permission issue or something else. Not getting this information in the error message can be quite frustrating.
For that reason, it seems to me a dynamic message would be better suited that fixed string in an assert.

After all, some of these failures are not really programming errors (as assertion would suggests), but they are actual failures that can happen at runtime, e.g. due to misconfiguration or integration issues.

That is why I would tend more towards a function, something like `safe_log_and_abort(..)` or something like that. The benefits I see there are:
* The code btw. fork() and execve() gets self-contained. No dependency on an assertion handler that's defined somewhere else.
* It fits better to the type of error. These are not programming errors that must never appear, but they are valid failures that may happen at runtime due to misconfiguration / integration issues
* Easier to support dynamic information in the log. Assert expects an already formatted message (which is difficult without heap usage).

I was thinking of something like this:
```
void safe_log_and_exit(const std::initializer_list<std::string_view> message_parts, const int error_code = 0);
...
safe_log_and_exit({"Failed to set rlimit ", rlimit_name}, errno);
safe_log_and_exit({"Failed to set CPU affinity"}, errno);
safe_log_and_exit({"Failed to set security policy ", config.security_policy_}, errno);
```

What do you think about that?

danth: This seems like a good direction.

Should we keep the async-signal-safe assertion handler anyways, or revert it back to the simpler implementation?

danth: With a separate function instead of assertions I think `_exit` with a failure status would be best

NicolasFussberger: I guess if we already have it we could keep it. Then it is also usable from a signal handler context.

But if there is more outstanding work, then I would tend to revert it. Using it from a signal handler is more of a nice-to-have in my opinion and in the current code I think we do not require it.

anmittag:
@danth there are checks failing, are you planning to repair or delete the PR?

danth:
@anmittag Yes, I'm working on the remaining review comments, should have an up to date version pushed soon.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,205 @@
...
+
+void safe_log_errno_and_exit(std::initializer_list<std::string_view> parts)
+{
+    for (auto part : parts)
+        stderr_print(part);
+
+#if _GNU_SOURCE && __GLIBC__ >= 2 && __GLIBC_MINOR__ >= 32
+    // strerrordesc_np is documented as async signal safe.
+    stderr_print(": ");
+    stderr_print(strerrordesc_np(errno));
```
danth: This is not available in our current glibc version but it's the only guaranteed async signal safe implementation of `strerror` I could find for Linux

paulquiring:
```c++
#include <unistd.h>
#include <string.h>
#include <errno.h>

// A completely safe way to write errors to stderr after a fork
void safe_log_error(const char *msg) {
    if (msg) {
        write(STDERR_FILENO, msg, strlen(msg));
    }
    
    // Safely get the text representation of the current errno
    // strerror_r (POSIX version) is async-signal-safe, unlike standard strerror
    char err_buf[1024];
    if (strerror_r(errno, err_buf, sizeof(err_buf)) == 0) {
        write(STDERR_FILENO, ": ", 2);
        write(STDERR_FILENO, err_buf, strlen(err_buf));
    }
    
    write(STDERR_FILENO, "\n", 1);
}

// Example usage inside the child process:
// if (execve(...) < 0) {
//     safe_log_error("[CHILD] execve failed");
//     _exit(1);
// }
```

danth: According to https://stackoverflow.com/a/18199867/32911257 and https://www.man7.org/linux/man-pages/man7/signal-safety.7.html, `strerror_r` is not async signal safe.

The best solution may be to simply not log this extra information when a suitable function is not available.

score/launch_manager/src/daemon/src/common/assertion_handler.cpp
danth: Yes, this is possible. I think even if we prepared the whole string and made a single call, it could still get split if the write is interrupted.

paulquiring: @danth @NicolasFussberger Is this still relevant?

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -42,39 +38,12 @@ constexpr int kPosixSuccess = 0;
...
-    const struct rlimit limit{
-        .rlim_cur = amount,
-        .rlim_max = amount,
-    };
-
-    applyLimitOrDie(resource, limit, rlimit_name);
-}
+using score::lcm::internal::osal::OsalConfig;
 
+/// @warning This will abort if not successful.
```
paulquiring:
```suggestion
/// @warning This will exit with failure if not successful.
```

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
paulquiring: all other solutions would be to much effort.

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -99,85 +68,46 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
...
-            if (-1 == fcntl(IpcCommsSync::control_client_handler_nudge_fd, F_SETFD, 0))
-            {
-                LM_LOG_ERROR() << "[New process] fcntl() at line" << __LINE__
-                               << "failed:" << score::lcm::internal::errno_message(errno);
-            }
+            if (fcntl(IpcCommsSync::sync_fd, F_SETFD, 0) == -1)
+                safe_log_and_exit({"Failed to set sync_fd flags"});
+
+            if (fcntl(IpcCommsSync::control_client_handler_nudge_fd, F_SETFD, 0) == -1)
+                safe_log_and_exit({"Failed to set control_client_handler_nudge_fd flags: ", std::strerror(errno)});
```
paulquiring: std::strerror is not async-signal-safe; it seems they forgot to delete it here.

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -99,85 +68,46 @@ void handleComms(score::lcm::internal::osal::ChildProcessConfig& param)
...
             break;
         case CommsType::kLaunchManager:
             // nothing to do here
             break;
         default:
-            LM_LOG_ERROR() << "[New process] at line" << __LINE__ << "unknown CommsType"
-                           << static_cast<std::int32_t>(param.shared_block->comms_type_);
-            sysexit(EXIT_FAILURE);
+            safe_log_and_exit(
+                {"Unknown CommsType: ", std::to_string(static_cast<int>(param.shared_block->comms_type_))});
```
paulquiring: std::to_string is not async-signal-safe

score/launch_manager/src/daemon/src/common/assertion_handler.cpp
danth: This applies to the new function too

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.hpp
```diff
@@ -0,0 +1,54 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
```
paulquiring: The include guard is missing; have a look at any other header file to see what it should look like.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.hpp
```diff
@@ -0,0 +1,54 @@
...
+#include <string_view>
+
+#include "score/mw/launch_manager/osal/security_policy.hpp"
```
paulquiring: Is this header used?

danth: Removed

danth:
Tests failed due to a network error.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,210 @@
...
+    const size_t length = message.length();
+
+    while (offset < length)
+    {
+        const auto result = write(STDERR_FILENO, &message.at(offset), length - offset);
+
+        if (result == -1)
+        {
+            if (errno != EINTR)
+                _exit(EXIT_FAILURE);
```
NicolasFussberger: I think we should probably stick to the sysexit function we have in score/launch_manager/src/daemon/src/osal/sys_exit.hpp

This will make it easier to unit test this without interfering with other library code (e.g. googletest)

NicolasFussberger: I wonder if we should just return an error here and not directly exit.
Doing the exit in safe_log_and_exit / safe_log_errno_and_exit might be more obvious.

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
```diff
@@ -0,0 +1,210 @@
...
+{
+    for (auto part : parts)
+        stderr_print(part);
+
+    _exit(EXIT_FAILURE);
+}
+
+void safe_log_errno_and_exit(std::initializer_list<std::string_view> parts)
+{
+    for (auto part : parts)
```
NicolasFussberger: Should we first assemble the message in a buffer, to avoid having this split across multiple lines?
The alternative would be we accept this in favour of keeping the code as simple as possible.

What do you think @danth @paulquiring ?

danth: I think we should keep it simple

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.cpp
danth: I think it would be more "correct" but it also means we would have to add a lot of error checking which makes the later functions less clear. Since it's in a private namespace personally I would just leave it as is

score/launch_manager/src/daemon/src/process_group_manager/details/configure_process.hpp
paulquiring: Why is this comment resolved?

danth: Wait, I added it to the `.cpp` rather than the `.hpp` 😆

danth: Should be fixed now

danth:
New implementation considering previous feedback:

- Use a separate function, not the assertion handler, so it cannot be swapped out for something unsafe
- Use signal safe implementations of `strerror`, otherwise fall back to printing the raw error number
- Build the error message in a fixed size buffer to avoid allocations *and* multiple writes
- Comments on all affected functions to avoid later changes making this unsafe again
- Minimal diff to aid review

MaciejKaszynski (changes requested):
All looks good, think we should however also include a unit test just so coverage doesn't go down.

Look at https://stackoverflow.com/questions/3803465/how-to-capture-stdout-stderr-with-googletest for info how to replace the stderr and it should be quite simple to cover this all.

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
```diff
@@ -0,0 +1,125 @@
...
+        {
+            buffer[end++] = string[index];
+        }
+    }
+
+    /// @brief Append the given integer to the buffer.
+    template <typename T, std::enable_if_t<std::is_integral_v<T>, bool> = true>
+    void append(const T& value)
+    {
+        const auto [pointer, error] = std::to_chars(&buffer[end], &buffer.back(), value);
```
NicolasFussberger: What was the conclusion on using std::to_chars in a signal-safe context?

danth: https://gcc.gnu.org/onlinedocs/gcc-13.1.0/libstdc++/api/a00062_source.html (line 130)

It only accesses local variables and the buffer it was given. I don't see anything which stands out to me as unsafe

NicolasFussberger: If its not guaranteed to be async signal safe it may change depending on the compiler version or platform.
But maybe we can leave it here and replace it with a signal safe function in a separate PR? I cannot say how easy or hard this is.

score/launch_manager/src/daemon/src/common/signal_safe_log_UT.cpp
```diff
@@ -0,0 +1,81 @@
...
+{
+    RecordProperty(
+        "Description", "This test verifies that signal_safe_log successfully writes a simple message to stderr.");
+
+    testing::internal::CaptureStderr();
+    ASSERT_TRUE(signal_safe_log("Hello world"));
+    ASSERT_EQ(testing::internal::GetCapturedStderr(), "Hello world\n");
+}
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_simple_with_errno)
```
NicolasFussberger, line 43: would be good to have some boundary tests as well to make sure it works for the complete buffer size.

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
danth: Yes that's correct. Implementing it ourself would look similar to the definition I linked, maybe with some unneeded functionality stripped out.

MaciejKaszynski (dismissed):

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
```diff
@@ -0,0 +1,129 @@
...
+#endif
+
+#ifndef SIGNAL_SAFE_LOG_HPP_INCLUDED
+#define SIGNAL_SAFE_LOG_HPP_INCLUDED
+
+namespace
+{
+
+/// @brief Utility for building log messages in an async signal safe context.
+template <int C>
```
paulquiring:
```suggestion
template <std::size_t C>
```

Image

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
```diff
@@ -0,0 +1,129 @@
...
+class signal_safe_buffer
+{
+  public:
+    std::array<char, C> buffer = {};
```
paulquiring: We should use the google style guide 
```c+
class MyClass {
 private:
  int instance_variable_;      // Correct: snake_case with a trailing underscore
  std::string user_name_;      // Correct

 public:
  static constexpr int kMaxCount = 100; // Constant: kCamelCase, no trailing underscore
};

struct MyStruct {
  int instance_variable;       // Correct for structs: snake_case WITHOUT a trailing underscore
  std::string user_name;       // Correct
};
```

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
paulquiring: issue created https://github.com/eclipse-score/lifecycle/issues/385

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
```diff
@@ -0,0 +1,129 @@
...
+
+namespace score::lcm::internal
+{
+
+template <typename... T>
+[[nodiscard]] bool signal_safe_log(const T&... values)
+{
+    const int original_errno = errno;  // For reentrancy
+
+    signal_safe_buffer<1024> buffer = {};
```
paulquiring:
```suggestion
    signal_safe_buffer<1024U> buffer = {};
```

danth: Shouldn't this be inferred based on the type of the template argument? Now it has been changed to `std::size_t`

score/launch_manager/src/daemon/src/common/signal_safe_log_UT.cpp
```diff
@@ -0,0 +1,111 @@
...
+    const auto stderr = testing::internal::GetCapturedStderr();
+
+#if __QNXNTO__
+    EXPECT_EQ(stderr->size(), 1024);
+    EXPECT_EQ(stderr->back(), '\n');
+#else
+    EXPECT_EQ(stderr.size(), 1024);
+    EXPECT_EQ(stderr.back(), '\n');
+#endif
+}
```
paulquiring, line 93: why the if ... else ... endif?

danth: It needs `->` on QNX and `.` on Linux because the type is different.

score/launch_manager/src/daemon/src/common/signal_safe_log.hpp
paulquiring: Rule 5.13.4 Unsigned integer literals shall be appropriately suffixed. See MISRA C++:2 023 Guidelines for the use of C++17 in critical systems. Some SCA tool, not active currently, will complain about this in the future.

MaciejKaszynski (approved):

score/launch_manager/src/daemon/src/common/signal_safe_log_UT.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+TEST_F(signal_safe_log_test, signal_safe_log_test_maximum_length)
```
paulquiring, line 95: :)

paulquiring (approved):

PR #308, opened 2026-07-10
Simplify Bazel files for integration tests

### LINKED ISSUE #295: Simplify Test Bazel
### What

Our Bazel BUILD files for each tests are pretty long and making a macro that would wrap them all would be nice.

### Acceptance Criteria (DoD)

- Bazel macro / rule created for integration tests.

### How

_No response_

danth:
Fixes #295

### DIFF
--- a/tests/utils/bazel/integration.bzl
+++ b/tests/utils/bazel/integration.bzl
@@ -10,35 +10,82 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
+load("@rules_pkg//pkg:mappings.bzl", "pkg_attributes", "pkg_files")
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 load("@score_itf//:defs.bzl", "py_itf_test")
 load("@score_lifecycle_pip//:requirements.bzl", "all_requirements")
+load("//:defs.bzl", "launch_manager_config")
 load("//tests/utils/bazel:constants.bzl", "SCORE_TEST_INSTALL_PREFIX")
 
-def integration_test(name, srcs, test_binaries, args = [], deps = [], data = [], install_prefix = SCORE_TEST_INSTALL_PREFIX, **kwargs):
-    """Creates an integration test with test binaries available, also adds all
-    the required dependencies.
+def integration_test(
+        name,
+        srcs,
+        binaries,
+        files = [],
+        config = None,
+        install_prefix = SCORE_TEST_INSTALL_PREFIX,
+        **kwargs):
+    """Creates an integration test.
+
+    Test binaries and all required dependencies are made available.
+
+    This macro will create a few build targets which you may find useful:
+        :environment: A tarball of all the files in the test environment
+        :config: The launch manager configuration
+        :(name): Runs the test script
 
     Args:
-        test_binaries: Label of the `package_test_binaries` target
+        name: Name of the test
+        srcs: The test script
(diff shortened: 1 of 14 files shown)


tests/utils/bazel/integration.bzl
```diff
@@ -10,35 +10,77 @@
...
+
+    Test binaries and all required dependencies are made available.
 
     Args:
-        test_binaries: Label of the `package_test_binaries` target
+        name: Name of the test
+        srcs: Build targets for the test script
+        binaries: Build targets the binaries being tested
+        files: Build targets for additional files to install
+        config: Build target for the launch manager configuration file
```
MaciejKaszynski: Can you add a mention that the bazel target will be `config` as users might find it useful to test a config build by just building this target.

tests/utils/bazel/integration.bzl
```diff
@@ -10,35 +10,77 @@
...
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 load("@score_itf//:defs.bzl", "py_itf_test")
 load("@score_lifecycle_pip//:requirements.bzl", "all_requirements")
+load("//:defs.bzl", "launch_manager_config")
 load("//tests/utils/bazel:constants.bzl", "SCORE_TEST_INSTALL_PREFIX")
 
-def integration_test(name, srcs, test_binaries, args = [], deps = [], data = [], install_prefix = SCORE_TEST_INSTALL_PREFIX, **kwargs):
-    """Creates an integration test with test binaries available, also adds all
-    the required dependencies.
+def integration_test(
```
MaciejKaszynski, line 20: I think I would rather make a separate bazle rule to wrap everything and not replace this I think for the most part we would never actually use the original in the tests but might be possible in the future? What do you think?

MaciejKaszynski (changes requested):

tests/utils/bazel/integration.bzl
danth: Currently we generate these targets per test:

```
//tests/integration/process_wrong_binary_failure:binaries
//tests/integration/process_wrong_binary_failure:config
//tests/integration/process_wrong_binary_failure:files
//tests/integration/process_wrong_binary_failure:process_wrong_binary_failure
//tests/integration/process_wrong_binary_failure:process_wrong_binary_failure.test_binary
//tests/integration/process_wrong_binary_failure:tar
```

And also some test-specific ones.

I would also be inclined to rename the main target to something like `test`, rather than duplicating the folder name. Then we can document all of them.

MaciejKaszynski: actually duplicating the name is done so that the bazel target can be shortened

if you have
//tests/integration/my_test:my_test
then if the target and the parent directory is the same you can do
//tests/integration/my_test

so I would keep the test name duplication.

tests/utils/bazel/integration.bzl
danth: The previous macro already made a lot of assumptions, for example that binaries would be under `tests/name/`, so it would be hard to use it differently I think.

Perhaps the `launch_manager_config` call could be separate, but our current tests all use it. Doing it this way makes sure it has the same target name under every test which is convenient for debugging as you said above

MaciejKaszynski: ok we can leave it as one then

MaciejKaszynski (approved):

NicolasFussberger (approved):

PR #313, opened 2026-07-14
Add pre-commit config

paulquiring:
Introudces .pre-commit-config.yaml with the following checks:

* Run existing bazel Copyright check
* Run existing bazel Format check
* Run clang-format on changes (currently disabled, to be enabled once codebase is formatted)
* Run clang-tidy on changes (currently disabled, to be enabled once list of checks are aligned)


externals/ipc_dropin/include/ipc_dropin/ringbuffer.hpp
```diff
@@ -1,9 +1,23 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+ #ifndef IPC_DROPIN_RINGBUFFER_HPP_
```
MaciejKaszynski:
```suggestion
#ifndef IPC_DROPIN_RINGBUFFER_HPP_
```

externals/ipc_dropin/include/ipc_dropin/socket.hpp
```diff
@@ -1,4 +1,17 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+ #ifndef IPC_DROPIN_SOCKET_HPP_
```
MaciejKaszynski:
```suggestion
#ifndef IPC_DROPIN_SOCKET_HPP_
```

MaciejKaszynski (changes requested):

.pre-commit-config.yaml
```diff
@@ -29,22 +29,24 @@ repos:
...
           )
       - id: clang-tidy
-        entry: bazel build //... --config=clang_tidy
+        entry: bazel build --config=clang_tidy //...
         language: system
         files: '\.(c|cc|cpp|h|hpp)$'
         # all relevant files are currently excluded
         exclude: |
           (?x)^(
+           externals/|
```
paulquiring: indention

.clang-format
MaciejKaszynski, line 1: Is it intended to deviate from the convention of `.clang-format`?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,62 @@
...
+        files: (BUILD|BUILD\.bazel|WORKSPACE|\.bazelrc|[^/]+\.(py|bzl|bazel|yaml|yml|rs))$
+        pass_filenames: false
+  - repo: https://github.com/pocc/pre-commit-hooks
+    rev: v1.3.5
+    hooks:
+      - id: clang-format
+        name: clang-format
+        # Run the tool binary directly so pre-commit's staged relative paths work.
+        # If clang-format cannot be found, the toolchain folder name likely changed.
+        entry: bazel-bin/external/toolchains_llvm++llvm+llvm_toolchain/clang-format -i
```
MaciejKaszynski: I see that you added a script for the clang-tidy to wrap some stuff for the compile-commands, would it make sense to add the same for format even if it doesn't use the compile-commands might be nice to add the `bazel run @llvm_toolchain//:clang-tidy` as a backup so that if the toolchain moves like the comment says it would still work.

MaciejKaszynski (changes requested):

.clang-format
paulquiring: it's now .clang-format

MaciejKaszynski (approved):

.pre-commit-config.yaml
paulquiring: clang-format and clang-tidy paths are now resolved using cquery, scrips handle that now.

.clang-format
```diff
@@ -1,48 +1,4 @@
 BasedOnStyle: Google
-AccessModifierOffset: -2
```
NicolasFussberger, line 2: Is all the content no longer required?

cr_checker_exclusion
NicolasFussberger, line 1: What is this used for?

paulquiring: To exclude files from being formatted by the copyright checker, like [here](https://github.com/eclipse-score/baselibs/blob/main/cr_checker_exclusion). I'll delete this empty file if we have no use for it.

.clang-format
paulquiring: Running clang-format took quite a long time due to the contents of this file. I trimmed it down for testing purposes. I’ll have a look at what that should look like to check what MISRA C++ 2023 stipulates.

paulquiring: I've tidied up the main version a bit. The content remains unchanged.

cr_checker_exclusion
paulquiring: cr_checker_exclusion needs to be present, its a required parameter of copyright_checker. It can be empty.

MODULE.bazel
```diff
@@ -39,6 +40,20 @@ bazel_dep(name = "score_rust_policies", version = "0.0.5", dev_dependency = True
 bazel_dep(name = "score_cpp_policies", version = "0.0.1", dev_dependency = True)
 bazel_dep(name = "score_process", version = "1.6.0", dev_dependency = True)
 bazel_dep(name = "score_platform", version = "0.6.0", dev_dependency = True)
+bazel_dep(name = "score_bazel_tools_cc", version = "0.1.0", dev_dependency = True)
```
paulquiring: needed?

paulquiring: No, was removed in last commit.

NicolasFussberger (dismissed):

MODULE.bazel
```diff
@@ -84,6 +85,18 @@ use_repo(
...
+bazel_dep(name = "toolchains_llvm", version = "1.4.0")
+
+llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
+llvm.toolchain(
+    llvm_version = "15.0.2",
+    stdlib = {"linux-x86_64": "stdc++"},
+)
+use_repo(llvm, "llvm_toolchain")
+
+register_toolchains("@llvm_toolchain//:all")
```
pawelrutkaq: shall not be registered.

paulquiring: Can you provide a reason why the `llvm_toolchain` shall not be registered? This how `bazel-tools-cc`
does it [here](https://github.com/eclipse-score/bazel-tools-cc/blob/main/test/MODULE.bazel). The registration is need so that everybody uses the same `clang-tidy` and `clang-format` version. Tidy and format are part of the tool-chain.

pawelrutkaq: it.s a test. Now llvm will be part of toolchian resolution for build and order of calls will mater when it picks toolchain. See that the Time module or the Comm module does not register the toolchain ;)

paulquiring: Thank you for the explanation I removed the registration and marked `llvm_toolchain` as `dev_dependency`

NicolasFussberger (approved):

PR #314, opened 2026-07-14
Replace `PhmLogger` with `LM_LOG_` macros

### LINKED ISSUE #151: Improvement: Merge the two logging wrappers in launch_manager
> [!IMPORTANT]
> Make sure to link this issue with the PR for your improvement.

**Goal:** 
**- Lifecycle code uses a single log wrapper, which can log to stdout or mw::log and supports multiple log contexts**

For historical reasons, we currently have two log wrappers in use in the launch manager code.
- The former Health Monitor-thread uses a separate log wrapper
- The former LM code uses its own log wrapper

The goal is to merge both approaches into a single log wrapper following the approach used in LM 

Implementation details:
- Predefine only one context for "LM" and keep the rust lib context "HMON" 
- In lifecycle lib logs need to be updated to use the new wrapper same context "LM"

danth:
Closes #151


MaciejKaszynski (approved):
Everything looks good, this would merge the loggers into the same context.

@NicolasFussberger Shall we have one context or split alive monitor stuff into a separate context?

NicolasFussberger:
> ?

My impression is its best to keep it simple and we can later still split the context if that will be needed

NicolasFussberger (approved):

Issue #141, opened 2026-03-31
Decide on equivalent rust API for mw::lifecycle


NicolasFussberger:
Currently, the mw::lifecycle API is only available for C++.

We shall decide about an API we want to provide for rust applications.
Currently, rust applications must fallback to the rust binding of the [LaunchManager internal API](https://github.com/eclipse-score/lifecycle/tree/main/src/launch_manager_daemon/lifecycle_client_lib/rust_bindings) in order to report Running state.

PR #318, opened 2026-07-15
Rename the C and Rust report running APIs to match the C++ API

cameron-craig-etas:
This PR updates the C and Rust "report running" API names to match the C++ API name.
We would like the API names to be consistent to they are easier to work with.


NicolasFussberger (approved):

S-MOHAMD:
related to #141

PR #324, opened 2026-07-16
Clarify Application::Run() documentation

### LINKED ISSUE #158: Improvement: Documentation how method Application::Run() works
The documentation in [class score::mw::lifecycle::Application](https://github.com/eclipse-score/lifecycle/blob/main/src/lifecycle_client_lib/include/application.h) for method `Run()` is somewhat misleading/unprecise.

Regarding the return code it states:
> \return 0 on successful initialization, non-zero otherwise.

This seems to be a copy&pasto from the `Initialize` method. But it also leads to a more important question from an app developer's point of perspective:

1. May/shall the implementation of the `Run` method of an application "consume"/"take hold" of the calling thread until the application is stopped? The return code would represent the overall outcome of the 
2. Or shall the calling thread just be used to change the state to "running", i.e. if required spawning separate threads to do the actual work, and return asap; reporting with the return code if the app did change to running state successfully.

My guess is the first alternative, but it would be helpful to clearly state this in the documentation of this class.

shegazyy:
The previous doc had two problems: the \return tag was a copy-paste from Initialize() ("0 on successful initialization"), and there was no mention that Run() blocks the calling thread until the stop_token fires.

Fixes #158


shegazyy:
Hey @BjoernAtBosch 👋ready for review whenever you have the time.

anmittag:
@shegazyy some checks are failing, are you adapting or deleting the PR?

NicolasFussberger (approved):

PR #337, opened 2026-07-21
fix(lifecycle_client): link lifecycle_mock against lifecyclemanager t…

Chahult:
…o resolve vtable

lifecycle_mock provides a mock implementation of LifeCycleManager, which inherits from the concrete class defined in lifecyclemanager.cpp. Without linking against the :lifecyclemanager target, the vtable symbol is undefined when building lifecycle_mock as a shared library (--no-allow-shlib-undefined).

Added :lifecyclemanager to lifecycle_mock's deps to provide the vtable.


anmittag:
@Chahult some checks are failing,are you planning to adapt or delete the PR?

Chahult:
@NicolasFussberger Could you please review

Chahult:
> @Chahult some checks are failing,are you planning to adapt or delete the PR?

Adapt the changes for config management
Please assign this to me and also add Nicolas Fußberger as a reviewer

score/launch_manager/src/lifecycle_client/BUILD
```diff
@@ -39,7 +39,7 @@ cc_library(
     include_prefix = "score/mw/lifecycle",
     strip_include_prefix = "/score/launch_manager/src/lifecycle_client/src",
     tags = ["FUSA"],
-    visibility = ["//score/launch_manager:__subpackages__"],
+    visibility = ["//visibility:public"],
```
NicolasFussberger: Is this necessary for the current mocking to work?
In the score meeting we had discussed that ideally we would like to just expose a single bazel target for the lifecycle API

NicolasFussberger:
@Chahult In the lifecycle repo we have a bit the problem that these mocks were contributed but we have no tests for them at the moment and its also not really obvious how to use them.
This is why currently, there is a high chance we accidentally break them.

Would it be possible for you to contribute some tests based on your usage of these mocks? E.g. for some example application we have in the examples folder?

score/launch_manager/src/lifecycle_client/BUILD
Chahult: No, now instead of making individual targets public, I've added a single lifecycle_test_mocks facade target with //visibility:public, reverting all internal targets back to //score/launch_manager:__subpackages__. External packages should now depend only on :lifecycle_test_mocks.

Chahult:
> @Chahult In the lifecycle repo we have a bit the problem that these mocks were contributed but we have no tests for them at the moment and its also not really obvious how to use them. This is why currently, there is a high chance we accidentally break them.
> 
> Would it be possible for you to contribute some tests based on your usage of these mocks? E.g. for some example application we have in the examples folder?

Sure, Done — added examples/cpp_lifecycle_app/lifecycle_mocks_UT.cpp with tests that mirror how config_management uses the mocks (argument injection via ApplicationContextMock, lifecycle flow via LifeCycleManagerMock). These also serve as usage documentation for future contributors.

examples/cpp_lifecycle_app/BUILD
```diff
@@ -10,6 +10,18 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
+load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_test")
+
+cc_test(
```
NicolasFussberger: This looks like a self contained test, so not testing the existing sample application.
In that case I think it belongs better in score/launch_manager/src/lifecycle_client/src

score/launch_manager/src/lifecycle_client/BUILD
```diff
@@ -209,6 +210,19 @@ cc_library(
     ],
 )
 
+# Single public entry point for external packages that need lifecycle mocks in their tests.
+# Prefer depending on this target rather than on individual mock targets.
+cc_library(
+    name = "lifecycle_test_mocks",
```
NicolasFussberger: There is already the public target lifecycle_mock_cc which was supposed to fulfil this purpose.
It looks like your target does not include the lifecyclemanagermock. Is this not needed?
Could you adapt the lifecycle_mock_cc instead of creating this new target?

score/launch_manager/src/lifecycle_client/BUILD
```diff
@@ -60,7 +60,7 @@ cc_library(
     include_prefix = "score/mw/lifecycle",
     strip_include_prefix = "/score/launch_manager/src/lifecycle_client/src",
     tags = ["FUSA"],
-    visibility = ["//score/launch_manager:__subpackages__"],
+    visibility = ["//visibility:public"],
```
NicolasFussberger: Please change back the visibility as you mentioned in https://github.com/eclipse-score/lifecycle/pull/337/changes#r3642991228 for the other targets as well

NicolasFussberger (changes requested):

examples/cpp_lifecycle_app/BUILD
Chahult: moved from examples/cpp_lifecycle_app/ to score/launch_manager/src/lifecycle_client/src/

score/launch_manager/src/lifecycle_client/BUILD
Chahult: removed. Instead, adapted lifecycle_mock_cc in //score/launch_manager from a single-target alias into a cc_library aggregating all three mocks (applicationcontext_mock, lifecycle_mock, report_running_mock). The moved test now depends on //score/launch_manager:lifecycle_mock_cc.

score/launch_manager/src/lifecycle_client/BUILD
Chahult: reverted applicationcontext, application, and runapplication back to //score/launch_manager:__subpackages__

score/launch_manager/src/lifecycle_client/BUILD
```diff
@@ -186,6 +186,7 @@ cc_library(
     visibility = ["//score/launch_manager:__subpackages__"],
     deps = [
         ":applicationcontext_mock",
+        ":lifecyclemanager",
```
NicolasFussberger, line 189: Is it really correct to add the lifecylemanager here?
Now this target contains both, the lifecyclemanagermock.cpp and the actual lifecyclemanager implementation

score/launch_manager/BUILD
```diff
@@ -86,9 +86,14 @@ alias(
     actual = "//score/launch_manager/src/lifecycle_client:applicationcontext_mock",
 )
 
-alias(
+cc_library(
     name = "lifecycle_mock_cc",
-    actual = "//score/launch_manager/src/lifecycle_client:lifecycle_mock",
+    testonly = True,
+    deps = [
+        "//score/launch_manager/src/lifecycle_client:applicationcontext_mock",
```
NicolasFussberger: I think applicationcontext mock is already part of //score/launch_manager/src/lifecycle_client:lifecycle_mock

score/launch_manager/src/lifecycle_client/BUILD
Chahult: Yes, it's intentional. LifeCycleManagerMock inherits from the concrete LifeCycleManager class. Without :lifecyclemanager in its deps, the vtable symbols for the base class are undefined when building lifecycle_mock as a shared library under --no-allow-shlib-undefined.
 Adding :lifecyclemanager provides the vtable — this is the minimum fix to make the linker happy. The mock .cpp itself remains unchanged; it just needs the base class compiled in.

score/launch_manager/BUILD
Chahult: You're right , lifecycle_mock already depends on :applicationcontext_mock transitively. Removed the explicit dep from lifecycle_mock_cc

Chahult:
@NicolasFussberger Could you please review again.

NicolasFussberger (approved):

NicolasFussberger:
@Chahult I hope you don't mind, I updated the bazel lockfile because the PR check was failing after merging in latest main.

PR #340, opened 2026-07-21
hmon: Define configuration schema

eduard-moskalchuk:
Example JSON configuration:
```json
{
  "schema_version": 1,
  "supervisor_api_cycle_ms": 100,
  "internal_processing_cycle_ms": 50,
  "thread_parameters": {
    "policy": "POSIX_FIFO",
    "priority": 20,
    "affinity": [0, 1],
    "stack_size": 262144
  },
  "deadline_monitors": [
    {
      "monitor_tag": "control_loop_deadlines",
      "deadlines": [
        {
          "deadline_tag": "read_inputs",
          "range": {
            "min_ms": 1,
            "max_ms": 5
          }
        },
        {
          "deadline_tag": "publish_outputs",
          "range": {
            "min_ms": 2,
            "max_ms": 10
          }
        }
      ]
    }
  ],
  "heartbeat_monitors": [
    {
      "monitor_tag": "planner_heartbeat",
      "range": {
        "min_ms": 90,
        "max_ms": 110
      }
    }
  ],
  "logic_monitors": [
    {
      "monitor_tag": "operation_mode_flow",
      "initial_state": "initializing",
      "states": [
        {
          "state": "initializing",
          "allowed_transitions": ["running", "fault"]
        },
        {
          "state": "running",
          "allowed_transitions": ["degraded", "fault"]
        },
        {
          "state": "degraded",
          "allowed_transitions": ["running", "fault"]
        },
        {
          "state": "fault",
          "allowed_transitions": []
        }
      ]
    }
  ]
}
```

### DIFF
--- a/score/health_monitor/configuration/health_monitor_config.fbs
+++ b/score/health_monitor/configuration/health_monitor_config.fbs
@@ -1,0 +1,82 @@
+// ********************************************************************************
+// Copyright (c) 2026 Contributors to the Eclipse Foundation
+//
+// See the NOTICE file(s) distributed with this work for additional
+// information regarding copyright ownership.
+//
+// This program and the accompanying materials are made available under the
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+// ********************************************************************************
+
+namespace score.mw.health.config;
+
+file_identifier "HMCF";
+file_extension "bin";
+
+/// Scheduler policy for the monitoring thread.
+enum SchedulerPolicy : byte {
+  POSIX_OTHER = 0,
+  POSIX_FIFO = 1,
+  POSIX_RR = 2
+}
+
+/// Thread parameters for the monitoring thread.
+table ThreadParameters {
+  policy: SchedulerPolicy = null;
+  priority: int32 = null;
+  affinity: [uint32];
+  stack_size: uint32 = null;
+}
+
+/// A time range with minimum and maximum durations in milliseconds.
+table TimeRange {
+  min_ms: uint32;
+  max_ms: uint32;
+}
+
+/// A single deadline entry within a deadline monitor.
+table DeadlineEntry {
+  deadline_tag: string (required);
+  range: TimeRange (required);
+}
+
+/// Configuration for a deadline monitor.
+table DeadlineMonitor {
+  monitor_tag: string (required);
+  deadlines: [DeadlineEntry] (required);
+}
+
+/// Configuration for a heartbeat monitor.
+table HeartbeatMonitor {
+  monitor_tag: string (required);
+  range: TimeRange (required);
+}
+
+/// A state and its allowed transitions in a logic monitor.
+table StateNode {
+  state: string (required);
+  allowed_transitions: [string] (required);
+}
+
+/// Configuration for a logic monitor.
+table LogicMonitor {
+  monitor_tag: string (required);
+  initial_state: string (required);
+  states: [StateNode] (required);
+}
+
+/// Root configuration for a HealthMonitor instance.
+table HealthMonitorConfig {
+  schema_version: uint32 (required);
+  supervisor_api_cycle_ms: uint32 (required);
+  internal_processing_cycle_ms: uint32 (required);
+  thread_parameters: ThreadParameters;
+  deadline_monitors: [DeadlineMonitor];
+  heartbeat_monitors: [HeartbeatMonitor];
+  logic_monitors: [LogicMonitor];
+}
+
+root_type HealthMonitorConfig;
--- a/score/health_monitor/configuration/health_monitor_config.schema.json
+++ b/score/health_monitor/configuration/health_monitor_config.schema.json
@@ -1,0 +1,202 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "health_monitor_config.schema.json",
+  "title": "HealthMonitorConfig",
+  "description": "Configuration schema for HealthMonitor (compatible with health_monitor_config.fbs)",
+  "type": "object",
+  "required": [
+    "schema_version",
+    "supervisor_api_cycle_ms",
+    "internal_processing_cycle_ms"
+  ],
+  "properties": {
+    "schema_version": {
+      "type": "integer",
+      "description": "Schema version",
+      "enum": [
+        1
+      ]
+    },
+    "supervisor_api_cycle_ms": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "Cycle duration in ms for supervisor API notifications"
+    },
+    "internal_processing_cycle_ms": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "Cycle duration in ms for internal deadline checks"
+    },
+    "thread_parameters": {
+      "$ref": "#/$defs/ThreadParameters"
+    },
+    "deadline_monitors": {
+      "$ref": "#/$defs/DeadlineMonitors",
+      "description": "Map of deadline monitor groups"
+    },
+    "heartbeat_monitors": {
+      "$ref": "#/$defs/HeartbeatMonitors",
+      "description": "Map of heartbeat monitors"
+    },
+    "logic_monitors": {
+      "$ref": "#/$defs/LogicMonitors",
+      "description": "Map of logic monitors"
+    }
+  },
+  "additionalProperties": false,
+  "$defs": {
+    "SchedulerPolicy": {
+      "type": "string",
+      "enum": [
+        "POSIX_OTHER",
+        "POSIX_FIFO",
+        "POSIX_RR"
+      ],
+      "description": "Scheduler policy for the monitoring thread"
+    },
+    "ThreadParameters": {
+      "type": "object",
+      "properties": {
+        "policy": {
+          "$ref": "#/$defs/SchedulerPolicy",
+          "description": "Scheduler policy for the monitoring thread"
+        },
+        "priority": {
+          "type": "integer",
+          "description": "Thread priority"
+        },
+        "affinity": {
+          "type": "array",
+          "items": {
+            "type": "integer",
+            "minimum": 0
+          },
+          "description": "CPU core IDs the thread can run on"
+        },
+        "stack_size_bytes": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Stack size in bytes"
+        }
+      },
+      "additionalProperties": false
+    },
+    "TimeRange": {
+      "type": "object",
+      "required": [
+        "min_ms",
+        "max_ms"
+      ],
+      "properties": {
+        "min_ms": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Minimum duration in milliseconds"
+        },
+        "max_ms": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Maximum duration in milliseconds"
+        }
+      },
+      "additionalProperties": false
+    },
+    "DeadlineEntry": {
+      "type": "object",
+      "required": [
+        "min_ms",
+        "max_ms"
+      ],
+      "properties": {
+        "min_ms": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Minimum duration in milliseconds"
+        },
+        "max_ms": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Maximum duration in milliseconds"
+        }
+      },
+      "additionalProperties": false
+    },
+    "DeadlineMonitorEntryMap": {
+      "type": "object",
+      "propertyNames": {
+        "minLength": 1
+      },
+      "additionalProperties": {
+        "$ref": "#/$defs/TimeRange"
+      }
+    },
+    "DeadlineMonitors": {
+      "type": "object",
+      "propertyNames": {
+        "minLength": 1
+      },
+      "additionalProperties": {
+        "$ref": "#/$defs/DeadlineMonitorEntryMap"
+      }
+    },
+    "HeartbeatMonitors": {
+      "type": "object",
+      "propertyNames": {
+        "minLength": 1
+      },
+      "additionalProperties": {
+        "$ref": "#/$defs/TimeRange"
+      }
+    },
+    "StateTransitions": {
+      "type": "array",
+      "items": {
+        "type": "string",
+        "minLength": 1
+      },
+      "description": "List of states reachable from this state"
+    },
+    "StateNode": {
+      "type": "object",
+      "minProperties": 1,
+      "maxProperties": 1,
+      "propertyNames": {
+        "minLength": 1
+      },
+      "additionalProperties": {
+        "$ref": "#/$defs/StateTransitions"
+      }
+    },
+    "LogicMonitor": {
+      "type": "object",
+      "required": [
+        "initial_state",
+        "states"
+      ],
+      "properties": {
+        "initial_state": {
+          "type": "string",
+          "minLength": 1,
+          "description": "Starting state of the logic monitor"
+        },
+        "states": {
+          "type": "array",
+          "items": {
+            "$ref": "#/$defs/StateNode"
+          },
+          "minItems": 1
+        }
+      },
+      "additionalProperties": false
+    },
+    "LogicMonitors": {
+      "type": "object",
+      "propertyNames": {
+        "minLength": 1
+      },
+      "additionalProperties": {
+        "$ref": "#/$defs/LogicMonitor"
+      }
+    }
+  }
+}
--- a/score/health_monitor/configuration/health_monitor_config.example.json
+++ b/score/health_monitor/configuration/health_monitor_config.example.json
@@ -1,0 +1,67 @@
+{
+  "schema_version": 1,
+  "supervisor_api_cycle_ms": 100,
+  "internal_processing_cycle_ms": 50,
+  "thread_parameters": {
+    "policy": "POSIX_FIFO",
+    "priority": 20,
+    "affinity": [
+      0,
+      1
+    ],
+    "stack_size_bytes": 262144
+  },
+  "deadline_monitors": {
+    "control_loop_deadlines": {
+      "read_inputs": {
+        "min_ms": 1,
+        "max_ms": 5
+      },
+      "publish_outputs": {
+        "min_ms": 2,
+        "max_ms": 10
+      }
+    },
+    "communication_deadlines": {
+      "receive": {
+        "min_ms": 1,
+        "max_ms": 3
+      },
+      "transmit": {
+        "min_ms": 2,
+        "max_ms": 8
+      }
+    }
+  },
+  "heartbeat_monitors": {
+    "planner_heartbeat": {
+      "min_ms": 90,
+      "max_ms": 110
+    },
+    "worker_heartbeat": {
+      "min_ms": 40,
+      "max_ms": 60
+    }
+  },
+  "logic_monitors": {
+    "operation_mode_flow": {
+      "initial_state": "init",
+      "states": [
+        {
+          "init": [
+            "run",
+            "deinit"
+          ]
+        },
+        {
+          "run": [
+            "deinit"
+          ]
+        },
+        {
+          "deinit": []
+        }
+      ]
+    }
+  }
+}


score/health_monitor/src/cpp/health_monitor_loader.h
```diff
@@ -10,27 +10,26 @@
...
-/// Returns the IPC socket path for the alive monitoring interface of a component.
-inline std::string aliveInterfacePath(const std::string& component_name)
+class HealthMonitorLoader
 {
-    return "/lifecycle_health_" + component_name;
-}
+  public:
+    virtual ~HealthMonitorLoader() = default;
+
+    static std::unique_ptr<HealthMonitorLoader> Create();
```
arkjedrz: This is unnecessary, it enforces creation with factory method and allocation on a heap. Remove it.

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Configuration for a heartbeat monitor.
+table HeartbeatMonitor {
+  monitor_tag: string (required, id:0);
+  range: TimeRange (required, id:1);
+}
+
+/// A state and its allowed transitions in a logic monitor.
+table StateTransition {
```
arkjedrz: -> `StateNode`?
This is not really a state transition, this is a node of a state graph.

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,88 @@
...
+}
+
+/// Scheduler parameters: policy and priority.
+table SchedulerParameters {
+  policy: SchedulerPolicy = Other (id:0);
+  priority: int32 (id:1);
+}
+
+/// Thread parameters for the monitoring thread.
+table ThreadParameters {
```
arkjedrz, line 27: What if certain parameters are optional?

eduard-moskalchuk: Everything is optional unless explicitly marked as required, as I understand.

NicolasFussberger: For scalar types flatbuffer will use a default value (e.g. 0) and you will not know whether it was actually configured or not.
If that is not suitable, you can use " = null;" to truly make it optional.

For non-scalar types you get nullptr from flatbuffer api in case it is not configured.

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,88 @@
...
+  scheduler_parameters: SchedulerParameters (id:0);
+  affinity: [uint32] (id:1);
+  stack_size: uint32 (id:2);
+}
+
+/// A time range with minimum and maximum durations in milliseconds.
+table TimeRange {
+  min_ms: uint32 (id:0);
```
NicolasFussberger: I think you do not need to manually state these ids and can just remove them

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,88 @@
...
+  states: [StateNode] (required, id:2);
+}
+
+/// Root configuration for a HealthMonitor instance.
+table HealthMonitorConfig {
+  version_major: uint32 = 1 (id:0);
+  version_minor: uint32 = 0 (id:1);
+  supervisor_api_cycle_ms: uint32 (id:2);
+  internal_processing_cycle_ms: uint32 (id:3);
+  thread_parameters: ThreadParameters (id:4);
```
NicolasFussberger: Not related to these changed, but I would be curious @eduard-moskalchuk if this thread configuration is working in your context.
I would assume in production environment application process will not have the privileges to changes the affinity or scheduling policy / priority for any of its threads.

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,88 @@
...
+table LogicMonitor {
+  monitor_tag: string (required, id:0);
+  initial_state: string (required, id:1);
+  states: [StateNode] (required, id:2);
+}
+
+/// Root configuration for a HealthMonitor instance.
+table HealthMonitorConfig {
+  version_major: uint32 = 1 (id:0);
+  version_minor: uint32 = 0 (id:1);
```
NicolasFussberger: Probably we should align some versioning concept. In the launch manager config we only have a single schema version [here](https://github.com/eclipse-score/lifecycle/blob/78ddf713d85398ab3680f2150c50cbd5639dc956/score/launch_manager/src/daemon/src/configuration/details/new_lm_flatcfg.fbs#L189)

score/health_monitor/configuration/health_monitor_config.fbs
eduard-moskalchuk: Probably not. In QNX some secpol abilities would be required for that, and most applications don't have them.

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
SimonKozik: The LM schema is based on 2020-12 version. Maybe we can use newer version to keep this in sync?

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
...
+  "description": "Configuration schema for HealthMonitor (compatible with health_monitor_config.fbs)",
+  "type": "object",
+  "required": ["supervisor_api_cycle_ms", "internal_processing_cycle_ms"],
+  "properties": {
+    "schema_version": {
+      "type": "integer",
+      "minimum": 1,
+      "default": 1,
+      "description": "Schema version"
+    },
```
SimonKozik, line 19: LM is using following construct:

``` json
    "schema_version": {
      "type": "integer",
      "description": "Specifies the schema version number that the Launch Manager uses to determine how to parse and validate this configuration file.",
      "enum": [
        1
      ]
    }
```

With your current proposal, integrator can configure 12 as `schema_version` and configuration will still pass validation.
Additionally I would be careful with using `default` in schema. They are not applied during validation and need to be added during configuration read. Populating not configured parameters with default values, during configuration load, is also a bit of a hassle.

I would suggest to go for enum here, and keep defined values always to a single number. Additionally I would define `schema_version` as required at the object level.
This way you will force integrator to declare which version of schema was used for configuration and if configuration validates against that schema, you will know that this schema can be used for reading this configuration file.

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "health_monitor_config.schema.json",
+  "title": "HealthMonitorConfig",
+  "description": "Configuration schema for HealthMonitor (compatible with health_monitor_config.fbs)",
+  "type": "object",
+  "required": ["supervisor_api_cycle_ms", "internal_processing_cycle_ms"],
```
SimonKozik: It may be worth to add `schema_version` here as well.

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
...
+      "type": "integer",
+      "minimum": 1,
+      "default": 1,
+      "description": "Schema version"
+    },
+    "supervisor_api_cycle_ms": {
+      "type": "integer",
+      "minimum": 1,
+      "description": "Cycle duration in ms for supervisor API notifications"
+    },
```
SimonKozik, line 24: LM is using seconds as the measurement unit of time (e.g., '0.5' for 500 milliseconds). I would be tempted to suggest that we should use the same approach for health monitoring.

This way integrator will be not restricted with precision they would like to use. Encoding precision in parameter name will force configuration change, when we would allow different precision.

Additionally it may be useful to define time as a reusable type. This type can be then used by supervisor_api_cycle, internal_processing_cycle and TimeRange

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
...
+      "description": "List of logic monitors"
+    }
+  },
+  "additionalProperties": false,
+  "definitions": {
+    "SchedulerPolicy": {
+      "type": "string",
+      "enum": ["Other", "Fifo", "RoundRobin"],
+      "description": "Scheduler policy for the monitoring thread"
+    },
```
SimonKozik, line 56: LM is using this construct to define scheduling policy:

```json
            "scheduling_policy": {
              "type": "string",
              "description": "Specifies the scheduling policy applied to the component's initial thread. Supported values correspond to OS-defined policies (e.g., FIFO, RR, OTHER).",
              "anyOf": [
                {
                  "enum": [
                    "SCHED_FIFO",
                    "SCHED_RR",
                    "SCHED_OTHER"
                  ]
                },
                {
                  "type": "string"
                }
              ]
            }
```

This way you can get autocomplete, in some text editors, for POSIX defined schedulers, but you also allow configuration of OS specific schedulers.

Current version will restrict integrator to ["Other", "Fifo", "RoundRobin"] and force implementation to do translation from Fifo to SCHED_FIFO when reading config.

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
...
+    "SchedulerParameters": {
+      "type": "object",
+      "properties": {
+        "policy": { "$ref": "#/definitions/SchedulerPolicy" },
+        "priority": {
+          "type": "integer",
+          "description": "Thread priority"
+        }
+      },
+      "required": ["policy", "priority"],
```
SimonKozik: Are we sure that we would like to force integrator to always configure policy and priority? I would prefer to keep required configuration to bare minimum.

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,180 @@
...
+      "additionalProperties": false
+    },
+    "ThreadParameters": {
```
SimonKozik: I would suggest to merge SchedulerPolicy, SchedulerParameters and ThreadParameters into a single type.

I'm not sure if I see benefits of keeping SchedulerPolicy and SchedulerParameters as a separate type.

SimonKozik (commented):
It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.

I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.

score/health_monitor/configuration/health_monitor_config.schema.json
eduard-moskalchuk: If we use floating point numbers, we may have to prove that we do it in a safe way. An activity that can be avoided. Please see the issue https://github.com/eclipse-score/lifecycle/issues/345

score/health_monitor/configuration/health_monitor_config.schema.json
eduard-moskalchuk: But does it make sense to support in the config something that is not supported in the code? I looked at LM implementation and have doubts about this. Also, there is anyway translation from string "SCHED_FIFO" to its actual value in the implementation.

NicolasFussberger: I remember in the flatbuffer file we could not use "SCHED_FIFO" in launch manager, because the generated code would conflict with the original identically-named posix definition of these values.

In the launch manager json file is not really an issue, because there is a script that translates this to the flatbuffer schema format so the name SCHED_FIFO does not actually end up in the code.

NicolasFussberger: Lifecycle Meeting:

For POSIX scheduling policies, use name POSIX_<scheduling_policy>, for os-specific scheduling policies use <OS>_<scheduling_policy>.

The list should be fixed and not user extendable via free-form string.

eduard-moskalchuk:
> It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> 
> I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.

@SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?

SimonKozik:
> > It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> > I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.
> 
> @SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?

@eduard-moskalchuk Thanks for the example. I would definitely recommend to store it alongside schema (this is what LM also does).

Does it make sense to get rid of the tags? For me it is just extra typing for the user, without added value to the information that is stored. An example could looks like this:
```json
{
  "deadline_monitors": {
    "control_loop_deadlines": {
      "read_inputs": {
        "min_ms": 1,
        "max_ms": 5
      },
      "publish_outputs": {
        "min_ms": 2,
        "max_ms": 10
      }
    },
    "communication_deadlines": {
      "receive": {
        "min_ms": 1,
        "max_ms": 3
      },
      "transmit": {
        "min_ms": 2,
        "max_ms": 8
      }
    }
  }
}
```

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,223 @@
...
+    "logic_monitors": {
+      "type": "array",
+      "items": {
+        "$ref": "#/definitions/LogicMonitor"
+      },
+      "description": "List of logic monitors"
+    }
+  },
+  "additionalProperties": false,
+  "definitions": {
```
SimonKozik:
```suggestion
  "$defs": {
```

In 2020-12 schema, the recommended location for reusable schema fragments is `$defs`

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,82 @@
...
+// ********************************************************************************
+
+namespace score.mw.health.config;
+
+file_identifier "HMCF";
+file_extension "bin";
+
+/// Scheduler policy for the monitoring thread.
+enum SchedulerPolicy : byte {
+  Other = 0,
```
NicolasFussberger: Does this need to be adapted as well to the new names POSIX_OTHER?
I assume you want to directly compile the json to flatbuffer without any conversion script

score/health_monitor/configuration/health_monitor_config.schema.json
```diff
@@ -0,0 +1,223 @@
...
+        },
+        "affinity": {
+          "type": "array",
+          "items": {
+            "type": "integer",
+            "minimum": 0
+          },
+          "description": "CPU core IDs the thread can run on"
+        },
+        "stack_size": {
```
NicolasFussberger: Should we put the unit into the name, as we did with the timing values (_ms)?
e.g. stack_size_bytes?

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,82 @@
...
+  monitor_tag: string (required);
+  initial_state: string (required);
+  states: [StateNode] (required);
+}
+
+/// Root configuration for a HealthMonitor instance.
+table HealthMonitorConfig {
+  schema_version: uint32 = null;
+  supervisor_api_cycle_ms: uint32;
+  internal_processing_cycle_ms: uint32;
```
NicolasFussberger: Are supervisor_api_cycle_ms and internal_processing_cycle_ms optional values?
Currently, flatbuffer will return 0 here if not configured.
This might be OK since 0 will be invalid.
Though using = null may be more explicit if its really optional

score/health_monitor/configuration/health_monitor_config.fbs
```diff
@@ -0,0 +1,82 @@
...
+/// Configuration for a logic monitor.
+table LogicMonitor {
+  monitor_tag: string (required);
+  initial_state: string (required);
+  states: [StateNode] (required);
+}
+
+/// Root configuration for a HealthMonitor instance.
+table HealthMonitorConfig {
+  schema_version: uint32 = null;
```
NicolasFussberger: Is the schema version really optional?

eduard-moskalchuk:
> > > It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> > > I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.
> > 
> > 
> > @SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?
> 
> @eduard-moskalchuk Thanks for the example. I would definitely recommend to store it alongside schema (this is what LM also does).
> 
> Does it make sense to get rid of the tags? For me it is just extra typing for the user, without added value to the information that is stored. An example could looks like this:
> 
> ```json
> {
>   "deadline_monitors": {
>     "control_loop_deadlines": {
>       "read_inputs": {
>         "min_ms": 1,
>         "max_ms": 5
>       },
>       "publish_outputs": {
>         "min_ms": 2,
>         "max_ms": 10
>       }
>     },
>     "communication_deadlines": {
>       "receive": {
>         "min_ms": 1,
>         "max_ms": 3
>       },
>       "transmit": {
>         "min_ms": 2,
>         "max_ms": 8
>       }
>     }
>   }
> }
> ```

@SimonKozik your example looks better in JSON, but it's not directly convertible to flatbuffers, is it?

SimonKozik:
> > > > It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> > > > I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.
> > > 
> > > 
> > > @SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?
> > 
> > 
> > @eduard-moskalchuk Thanks for the example. I would definitely recommend to store it alongside schema (this is what LM also does).
> > Does it make sense to get rid of the tags? For me it is just extra typing for the user, without added value to the information that is stored. An example could looks like this:
> > ```json
> > {
> >   "deadline_monitors": {
> >     "control_loop_deadlines": {
> >       "read_inputs": {
> >         "min_ms": 1,
> >         "max_ms": 5
> >       },
> >       "publish_outputs": {
> >         "min_ms": 2,
> >         "max_ms": 10
> >       }
> >     },
> >     "communication_deadlines": {
> >       "receive": {
> >         "min_ms": 1,
> >         "max_ms": 3
> >       },
> >       "transmit": {
> >         "min_ms": 2,
> >         "max_ms": 8
> >       }
> >     }
> >   }
> > }
> > ```
> 
> @SimonKozik your example looks better in JSON, but it's not directly convertible to flatbuffers, is it?

@eduard-moskalchuk This is true, but we could also generate the tags automatically before invoking flatc in Bazel. Just a suggestion though, so I'll leave it up to you to decide which approach provides the best user experience.

NicolasFussberger:
> > > > > It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> > > > > I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.
> > > > 
> > > > 
> > > > @SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?
> > > 
> > > 
> > > @eduard-moskalchuk Thanks for the example. I would definitely recommend to store it alongside schema (this is what LM also does).
> > > Does it make sense to get rid of the tags? For me it is just extra typing for the user, without added value to the information that is stored. An example could looks like this:
> > > ```json
> > > {
> > >   "deadline_monitors": {
> > >     "control_loop_deadlines": {
> > >       "read_inputs": {
> > >         "min_ms": 1,
> > >         "max_ms": 5
> > >       },
> > >       "publish_outputs": {
> > >         "min_ms": 2,
> > >         "max_ms": 10
> > >       }
> > >     },
> > >     "communication_deadlines": {
> > >       "receive": {
> > >         "min_ms": 1,
> > >         "max_ms": 3
> > >       },
> > >       "transmit": {
> > >         "min_ms": 2,
> > >         "max_ms": 8
> > >       }
> > >     }
> > >   }
> > > }
> > > ```
> > 
> > 
> > @SimonKozik your example looks better in JSON, but it's not directly convertible to flatbuffers, is it?
> 
> @eduard-moskalchuk This is true, but we could also generate the tags automatically before invoking flatc in Bazel. Just a suggestion though, so I'll leave it up to you to decide which approach provides the best user experience.



> > > > > It is difficult for me to comment on the LogicMonitor, StateNode, HeartbeatMonitor, DeadlineMonitor and DeadlineEntry definitions without seeing an example configuration.
> > > > > I would appreciate if we could have an example configuration (for all possible configuration parameters), so we could better asses the schema.
> > > > 
> > > > 
> > > > @SimonKozik I added an example to the PR description. Would it make sense to push it as a file as well?
> > > 
> > > 
> > > @eduard-moskalchuk Thanks for the example. I would definitely recommend to store it alongside schema (this is what LM also does).
> > > Does it make sense to get rid of the tags? For me it is just extra typing for the user, without added value to the information that is stored. An example could looks like this:
> > > ```json
> > > {
> > >   "deadline_monitors": {
> > >     "control_loop_deadlines": {
> > >       "read_inputs": {
> > >         "min_ms": 1,
> > >         "max_ms": 5
> > >       },
> > >       "publish_outputs": {
> > >         "min_ms": 2,
> > >         "max_ms": 10
> > >       }
> > >     },
> > >     "communication_deadlines": {
> > >       "receive": {
> > >         "min_ms": 1,
> > >         "max_ms": 3
> > >       },
> > >       "transmit": {
> > >         "min_ms": 2,
> > >         "max_ms": 8
> > >       }
> > >     }
> > >   }
> > > }
> > > ```
> > 
> > 
> > @SimonKozik your example looks better in JSON, but it's not directly convertible to flatbuffers, is it?
> 
> @eduard-moskalchuk This is true, but we could also generate the tags automatically before invoking flatc in Bazel. Just a suggestion though, so I'll leave it up to you to decide which approach provides the best user experience.

I think we should weigh the pros/cons for introducing such a mapping script.
We had different reasons for doing this for the launch manager daemon.
I think such script will require tool qualification / classification, testing, special bazel macro for the user to run the script, etc.
So I would recommend to only do this if this is worth the additional effort.

anmittag:
@eduard-moskalchuk a) branch is outdated and b) some review comments are open

Can you please adapt accordingly

PR #341, opened 2026-07-21
Redesign Health Monitor C++ API with abstract interfaces

eduard-moskalchuk:
Replace concrete class hierarchy with abstract interface design to decouple user code from the Rust FFI implementation:

- Extract pure virtual interfaces: HealthMonitor, HealthMonitorBuilder, DeadlineMonitor, Deadline, HeartbeatMonitor, LogicMonitor
- Move FFI-backed implementations into details/ namespace (*Impl classes)
- Introduce HealthMonitorBuilder::Create() factory method returning std::unique_ptr<HealthMonitorBuilder> for polymorphic construction
- Add GMock-based mock classes (health_monitor_mocks.h) enabling unit testing without the Rust FFI library
- Flatten per-monitor subdirectories (deadline/, heartbeat/, logic/) into top-level headers with details/ impl split
- Remove obsolete tag.h, thread.h/cpp in favor of common.h Tag class and ThreadParameters struct in builder header
- Update example app and integration tests to new API


pawelrutkaq (changes requested):
I will have a look overall, but my first look tells me i dont like need of allocations -> let me look into it closer ;)

pawelrutkaq:
@eduard-moskalchuk  this is what i meant (no skip need of allocations and virtual calls in production code) -> https://github.com/qorix-group/inc_lifecycle/pull/12/changes#diff-f3c426bacc8936a54221d843f347b0e592ce56aad4df991e1e514c0caf78cabe   You can find me on slack as Pawel Rutka and we can dicuss if this would also fill your needs (i directly used your example, now kind of fast & maybe dirty but shows the idea)

eduard-moskalchuk:
Please consider redesigning the API now. Delaying it until users depend on concrete types will make migration far more expensive.

The approach proposed in this PR offers several benefits:

- Standard C++ idiom — virtual interfaces are the canonical pattern for testability. Any C++ developer will immediately understand them; there is no need to explain FFI-stub replacement or friend-template workarounds.
- Cleaner production headers — mocking follows naturally from the design rather than being bolted on.
- No hidden coupling to the Rust FFI surface — interfaces and mocks are completely decoupled from FFI.
- Negligible performance impact — the cost of virtual calls is insignificant compared to the existing FFI + Rust overhead on this path.

pawelrutkaq:
We talked internally, and we agreed to move on with a virtual interface, but:

1. Please revert all API renames in this PR
2. Please revert all other not needed movment of code that was done with this interface proposal

Then we can continue; hope this helps

eduard-moskalchuk:
Great, thanks @pawelrutkaq !
Regarding the renames: the project looks inconsistent when naming conventions aren't applied uniformly. According to .clang-format, we follow the Google style, so method names should use PascalCase. Do any existing dependencies rely on the current names?

pawelrutkaq:
Unless we don't align whole codebase I want to keep it consistent. So if all other components of this repo would do this HealtMon shall also do it (ie report_running)).

Anyway, a separate PR shall be done only for renames.

PR #352, opened 2026-07-23
Preparing to format the code base 

paulquiring:
* Enables clang-format pre-commit hook
* Update in code coverage workflow
* Update in MODULE.bazel


.github/workflows/code_coverage.yml
```diff
@@ -24,7 +24,7 @@ on:
 
 jobs:
   cpp:
-    uses: eclipse-score/cicd-workflows/.github/workflows/cpp-coverage.yml@ea19fcae9aeeb4ac678b750c6c197eaf75414f39 # v0.0.0
+    uses: eclipse-score/cicd-workflows/.github/workflows/cpp-coverage.yml@main
```
MaciejKaszynski: I dont think we should use the branch. I think in general people use the commit tag so that CI is stable even when template is changed.
So if we need newer stuff I would just get the commit tag of main that we need.

MaciejKaszynski (changes requested):

.github/workflows/code_coverage.yml
paulquiring: I understand, but I’d prefer an automatic update. In the example main is used:
[C++ Coverage Workflow](https://github.com/eclipse-score/cicd-workflows/blob/4ce43f957a0e164827ad2906a2ec69fd7f420b5d/README.md#7-c-coverage-workflow)
Fail early fail often.

MaciejKaszynski: @NicolasFussberger Deciding vote?

Think that failing early is good for **our** code, a major change in the template would mean we are blocked from PRs on our repo until we fix the CI

danth: Pinning to a specific commit is also [recommended to avoid supply chain attacks](https://cheatsheetseries.owasp.org/cheatsheets/GitHub_Actions_Security_Cheat_Sheet.html#use-third-party-github-actions-and-reusable-workflows-securely), although debatable whether that matters between our own repos. That page is a good read either way 😄

NicolasFussberger: I guess I would also favor the fixed commits, with either a bot that updates this regularly or we have to update this manually regularly.

With directly using main, when they roll out breaking changes it would just break our pipelines and we cannot distinguish between minor changes and major changes.
Would be great if we could pin to version 1.x.x or something like that so that we get the latest as long as its not a breaking change.

paulquiring: All right then, I’ll pin that version. I think we should check every now and then to see if there are any newer versions.

paulquiring: @NicolasFussberger to be able to pin only the major e.g. v1, releases of minor and bug fixes need to be force pushed to v1 as well. We are still in the major 0 phase, pinning that would mean to use main as it is right now. For now I will simply use the hash of v0.0.2.

.pre-commit-config.yaml
```diff
@@ -30,18 +30,20 @@ repos:
...
-        # all relevant files are currently excluded the format is not 
-        # clarified yet, so we don't want to run clang-format on any files
         exclude: |
           (?x)^(
             externals/|
             examples/|
             tests/|
             score/
           )
+
```
paulquiring: remove blank line

NicolasFussberger (approved):

MaciejKaszynski (approved):

Issue #330, opened 2026-07-17
Move the watchdog code to launch manager


cameron-craig-etas:
#### Scope

Introduce watchdog initialization, servicing, and deinitialization into `ProcessGroupManager::initialize()`, `run()`, and `deinitialize()`. Only unconditional servicing (`serviceWatchdog()`) is in scope. Conditions for firing the watchdog (`fireWatchdogReaction()`) are out of scope and will be addressed separately.

#### Background

The watchdog is currently owned and operated by `PhmDaemon` (score/launch_manager/src/daemon/src/alive_monitor/details/daemon/PhmDaemon.hpp). The lifecycle is:

1. `PhmDaemon::init()`: loads config via `MachineConfigFactory`, calls `watchdog->init(cycleTimeNs, configFactory)` then `watchdog->enable()`

2. `PhmDaemon::startCyclicExec()`: calls `watchdog->serviceWatchdog()` each cycle (or `fireWatchdogReaction()` on critical failure)

3. On termination signal: calls `watchdog->disable()`

Since `health_monitor` and `launch_manager` now run in the same process, watchdog ownership should move to `ProcessGroupManager`. The watchdog will be initialized and enabled in `ProcessGroupManager::initialize()`, serviced every iteration of `ProcessGroupManager::run()`, and disabled in `ProcessGroupManager::deinitialize()`.

#### Future watchdog trigger conditions (out of scope)

For context, the following conditions will eventually trigger `fireWatchdogReaction()` instead of `serviceWatchdog()`:

* `recovery_client_->hasOverflow()` — recovery request queue overflow
* `queuePosixProcess()` failure — process state notification buffer full
* Other unrecoverable LaunchManager internal errors

#### Ownership and instantiation

* `WatchdogImpl` is instantiated in `score/launch_manager/src/daemon/src/main.cpp` (as it is today).
* `ProcessGroupManager` receives a `std::unique_ptr<IWatchdogIf>` via its constructor.
* The watchdog is only used from the main thread. No thread safety concerns.

#### Watchdog configuration

The watchdog requires two pieces of configuration for `init()`:

* `cycleTimeInNs` (int64_t) — the expected servicing cycle time in nanoseconds
* An `IDeviceConfigFactory` — provides the list of watchdog device configurations (device path, timeout range, etc.)

A new configuration will be introduced within `ProcessGroupManager` that contains the relevant watchdog configuration data.

[...]

PR #355, opened 2026-07-23
Introduce watchdog into `ProcessGroupManager` [3/N]: Move ownership to `ProcessGroupManager`

cameron-craig-etas:
#330 

This is part 3 of a number of PRs, to move the watchdog ownership to `ProcessGroupManager`. Now that we have prepared the watchdog sources in previous PRs, we are now doing the actual owner change here.

I have performed the following steps:
- We now create a watchdog object using the `createWatchdog()` factory.
- The watchdog has been removed from the alive monitor and `PhmDaemon`.
- We now call the watchdog from the main loop in `process_group_manager.cpp`
- The watchdog is only enabled for the new config.

Note that unit tests are to follow in a separate PR.


score/launch_manager/src/daemon/src/watchdog/details/WatchdogImpl.cpp
```diff
@@ -55,38 +56,39 @@ T secToMs(const T f_timeout)
...
     bool isSuccess{true};
     try
     {
-        const auto configurations{f_configFactory.getDeviceConfigurations()};
-        if (!configurations)
+        if (watchdog_config.max_timeout_ms > std::numeric_limits<std::uint16_t>::max())
         {
-            LM_LOG_ERROR() << "Watchdog: Invalid watchdog device configuration. Watchdog initialization failed.";
-            isSuccess = false;
+            LM_LOG_ERROR() << "Watchdog: Invalid watchdog timeout value " << watchdog_config.max_timeout_ms
```
NicolasFussberger:
```suggestion
            LM_LOG_ERROR() << "Watchdog: Invalid watchdog timeout value" << watchdog_config.max_timeout_ms
```

I think the "<<" operator in logging implementation adds an additional whitespace automatically. If we add an extra whitespace, we'll have 2 whitespace in the resulting log line

NicolasFussberger:
I think score/launch_manager/src/daemon/src/watchdog/IDeviceConfigFactory.hpp can now be moved to details folder, as the public interface does not require this anymore.

score/launch_manager/src/daemon/src/watchdog/details/BUILD
NicolasFussberger: One idea: We could set the visibility for the library targets in this file to `"//score/launch_manager/src/watchdog:__subpackages__"` so that bazel also enforced that these classes are internal to the watchdog component.

NicolasFussberger:
If I see it correctly, I think score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp and score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory_new.cpp can be removed / trimmed down as the watchdog is no longer loaded from config there.

I think the only value still required is the periodicity in which supervisions are evaluated.
For the other values we anyway only use the defaults as this is no longer configurable in the json.

NicolasFussberger:
> If I see it correctly, I think score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory.cpp and score/launch_manager/src/daemon/src/alive_monitor/details/factory/MachineConfigFactory_new.cpp can be removed / trimmed down as the watchdog is no longer loaded from config there.
> 
> I think the only value still required is the periodicity in which supervisions are evaluated. For the other values we anyway only use the defaults as this is no longer configurable in the json.

Though this could also be done in a separate PR if you like to keep the changes here limited.

NicolasFussberger (changes requested):

score/launch_manager/src/daemon/src/watchdog/BUILD
```diff
@@ -26,5 +26,8 @@ cc_library(
     include_prefix = "score/mw/launch_manager/watchdog",
     strip_include_prefix = "/score/launch_manager/src/daemon/src/watchdog",
     visibility = ["//score:__subpackages__"],
-    deps = [":i_device_config_factory"],
+    deps = [
+        ":i_device_config_factory",
```
NicolasFussberger:
```suggestion

```

NicolasFussberger:
> I think score/launch_manager/src/daemon/src/watchdog/IDeviceConfigFactory.hpp can now be moved to details folder, as the public interface does not require this anymore.

Done. I removed the loading of watchdog settings from MachineConfigFactory.hpp/cpp and remove entirely the MachineConfigFactory_new.cpp

score/launch_manager/src/daemon/src/watchdog/details/BUILD
NicolasFussberger: done

NicolasFussberger (dismissed):

MaciejKaszynski (dismissed):

MaciejKaszynski (approved):

PR #356, opened 2026-07-23
Support deployment_config/working_dir value

### LINKED ISSUE #343: [launch_manager] Honor deployment_config.working_dir as the spawned process cwd
### What

The Launch Manager never applies a component's configured working_dir. The child process cwd is hard-coded to the executable's directory in changeCurrentWorkingDirectory() (chdir(dirname(executable_path))), and the working_dir value is dropped entirely on the way to the daemon.
As a result, applications that access files relative to their cwd (e.g. ./etc/file.txt) fail, because the cwd was <bin_dir> instead of the configured working_dir.

### Acceptance Criteria (DoD)

Working directory specified in the deployment config is applied.

### How

_No response_

TimoSteuerwaldETAS:
This fixes #343 .
Furthermore the config mapping tests have been fixed and are executed on the pipeline.
A feature integration test has been also added which is testing various options, not only working dir.


patches/score_itf_docker_cap_add.patch
```diff
@@ -0,0 +1,11 @@
+--- a/score/itf/plugins/docker.py
++++ b/score/itf/plugins/docker.py
+@@ -331,6 +331,8 @@ def target_init(request, _docker_configuration):
+         environment=_docker_configuration["environment"],
+         volumes=_docker_configuration["volumes"],
+         shm_size=_docker_configuration["shm_size"],
++        privileged=_docker_configuration.get("privileged", False),
```
NicolasFussberger: Any plan to bring these changes to ITF repo?

scripts/config_mapping/lifecycle_config.py
```diff
@@ -185,9 +184,13 @@ def dict_merge_recursive(dict_a, dict_b):
...
-    new_config["initial_run_target"] = config.get(
-        "initial_run_target", merged_defaults["initial_run_target"]
-    )
+    # Only emit initial_run_target when it can be derived from the input config
+    # or the defaults. This keeps preprocess_defaults usable with minimal
+    # defaults that don't define run-target keys.
+    if "initial_run_target" in config or "initial_run_target" in merged_defaults:
+        new_config["initial_run_target"] = config.get(
+            "initial_run_target", merged_defaults.get("initial_run_target")
+        )
```
NicolasFussberger, line 200: I think this change it redundant. There is a validation that ensures that initial_run_target must always be defined (to Startup state)

scripts/config_mapping/lifecycle_config.py
```diff
@@ -289,7 +292,9 @@ def gen_config(output_dir, config, input_filename, schema_version=None):
             "ready_timeout": depl_cfg["ready_timeout"],
             "shutdown_timeout": depl_cfg["shutdown_timeout"],
             "bin_dir": depl_cfg["bin_dir"],
-            "working_dir": depl_cfg["working_dir"],
+            # Default the working directory to bin_dir (the directory the
+            # executable lives in) when not set explicitly.
+            "working_dir": depl_cfg.get("working_dir", depl_cfg["bin_dir"]),
```
NicolasFussberger: This decision to use bin_dir as default exists in two different places, can this be moved to a single function (e.g. get_working_dir)?

scripts/config_mapping/tests/basic_test/expected_output/lm_demo.json
```diff
@@ -20,6 +20,7 @@
             ],
             "security_policy_details": "policy_name",
             "number_of_restart_attempts": 1,
+            "working_dir": "/tmp",
```
NicolasFussberger: Could you adapt one of the configs so that we have all cases covered?
1. Taking the default working dir (as bin dir)
2. Setting the working_dir in the defaults section
3. Setting the working dir explicitly on the component (overwriting the default)

tests/utils/testing_utils/run_until_file_deployed.py
```diff
@@ -35,6 +36,8 @@ def run_until_file_deployed(
     :param file_path: path of the file to wait for on the target.
     :param timeout_s: maximum seconds to wait for the file (default: 30).
     :param poll_interval_s: seconds between file-existence checks (default: 0.5).
```
NicolasFussberger: Maybe update the documentation that poll_interval_s is also used to poll process state

tests/utils/testing_utils/run_until_file_deployed.py
```diff
@@ -64,12 +67,21 @@ def run_until_file_deployed(
...
+
+            # Poll for the process to actually terminate instead of assuming a
+            # single fixed grace period is enough. Termination timing varies
+            # with system load, so a fixed sleep is flaky.
+            stop_deadline = time.monotonic() + stop_timeout_s
+            while proc.is_running() and time.monotonic() < stop_deadline:
+                time.sleep(poll_interval_s)
+
+            assert not proc.is_running(), (
+                f"LCM still running {stop_timeout_s}s after SIGTERM"
```
NicolasFussberger: What happens to other tests when the Process does not shutdown on receiving SIGTERM?
Is the process then killed with -9 to force termination so the next tests are not affected by this?

tests/integration/sandbox_options/BUILD
```diff
@@ -0,0 +1,53 @@
...
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+load("@rules_cc//cc:defs.bzl", "cc_binary")
+load("//tests/utils/bazel:integration.bzl", "integration_test")
+
+exports_files(
+    ["lm_demo.json"],
```
NicolasFussberger: I don't understand where is this file and why is it visible for examples folder?

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,178 @@
...
+constexpr int expected_priority = 10;
+
+const char* policy_name(const int policy)
+{
+    switch (policy)
+    {
+        case SCHED_FIFO:
+            return "SCHED_FIFO";
+        case SCHED_RR:
+            return "SCHED_RR";
```
NicolasFussberger: I think SCHED_RR was not tested.

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,178 @@
...
+    return pass;
+}
+
+bool verify_sandbox_options()
+{
+    bool all_pass = true;
+
+    const uid_t expected_uid = 666;
+    const gid_t expected_gid = 666;
+    const std::array<gid_t, 3> expected_supp_groups = {123, 321, 456};
```
NicolasFussberger: I think we should cover at least the cases:
* No supplementary group ids configured
* Some supplementary group ids configured

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,178 @@
...
+    const std::string expected_cwd = "/tmp";
```
NicolasFussberger: I think we should at least cover the cases:
* A specific working_dir configured
* The default working_dir is used

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return pass;
+}
+
+bool verify_sandbox_options()
+{
+    bool all_pass = true;
+
+    const uid_t expected_uid = 666;
```
NicolasFussberger: Can this be a reusable process, where you pass the expected values via CLI flags?

tests/integration/sandbox_options/control_daemon_mock.cpp
```diff
@@ -0,0 +1,52 @@
...
+    TEST_STEP("Activate RunTarget Running")
+    {
+        score::cpp::stop_token stop_token;
+        auto result = client.ActivateRunTarget("Running").Get(stop_token);
+        EXPECT_TRUE(result.has_value()) << "Activating target Running failed: " << result.error().Message();
+    }
+    TEST_STEP("Activate RunTarget Startup")
+    {
+        score::cpp::stop_token stop_token;
+        auto result = client.ActivateRunTarget("Startup").Get(stop_token);
```
NicolasFussberger: Launch Manager goes to Startup by default. I think this is redundant.
Looks like we can remove the whole control_daemon_mock from this test

tests/integration/sandbox_options/sandbox_options.json
```diff
@@ -0,0 +1,124 @@
...
+            "deployment_config": {
+                "environmental_variables": {
+                    "PROCESSIDENTIFIER": "DefaultPG_app0"
+                },
+                "sandbox": {
+                    "uid": 666,
+                    "gid": 666,
+                    "supplementary_group_ids": [123, 321, 456],
+                    "scheduling_policy": "SCHED_FIFO",
+                    "scheduling_priority": 10
```
NicolasFussberger, line 73: I think we should also assert some default priority if not configured

NicolasFussberger (changes requested):

patches/score_itf_docker_cap_add.patch
TimoSteuerwaldETAS: Something similar is already there. Because today the itf_score version has been increased to 0.5.0, I reverted the patch.

scripts/config_mapping/tests/basic_test/expected_output/lm_demo.json
TimoSteuerwaldETAS: Done.

tests/integration/sandbox_options/BUILD
TimoSteuerwaldETAS: Me neither. Removed.

tests/utils/testing_utils/run_until_file_deployed.py
TimoSteuerwaldETAS: I have added a `kill -9` to not influence other tests.

tests/integration/sandbox_options/sandbox_options_process.cpp
TimoSteuerwaldETAS: Sure. Fixed.

scripts/config_mapping/lifecycle_config.py
TimoSteuerwaldETAS: As the preprocessing of the config (this code segment) takes place before calling `custom_validations(..)` (with the returned `new_config`), I will keep this.

scripts/config_mapping/lifecycle_config.py
TimoSteuerwaldETAS: Fixed.

tests/utils/testing_utils/run_until_file_deployed.py
TimoSteuerwaldETAS: Fixed.

tests/integration/sandbox_options/sandbox_options_process.cpp
TimoSteuerwaldETAS: Add another process to this test which tests SCHED_RR.

tests/integration/sandbox_options/sandbox_options_process.cpp
TimoSteuerwaldETAS: Fixed.

tests/integration/sandbox_options/sandbox_options_process.cpp
TimoSteuerwaldETAS: Fixed. See sandbox_options_process_a and sandbox_options_process_b.

tests/integration/sandbox_options/control_daemon_mock.cpp
TimoSteuerwaldETAS: As this test now uses two processes, I would rather stick to control_daemon_mock, because the control_daemon_mock is writing the test_end file. If we drop that process it is unclear when this test is finished, as now the two processes under test need to write the file instead.

score/launch_manager/src/daemon/src/configuration/config_schema/lm_flatcfg.fbs
```diff
@@ -43,6 +43,7 @@ table Process {
   uid: uint (id:4);
   gid: uint (id:5);
   path: string (id:6);
+  working_dir: string (id:11);
```
MaciejKaszynski, line 46: This is adding this to the old `lm_flatcfg.fbs` however we are converting to the new schema in `score/launch_manager/src/daemon/src/configuration/details/new_lm_flatcfg.fbs` I think we can make this change just for the new config and not edit the old.

MaciejKaszynski: Same thing for the loader code in `score/launch_manager/src/daemon/src/configuration/`

MaciejKaszynski (changes requested):

score/launch_manager/src/daemon/src/configuration/config_schema/lm_flatcfg.fbs
MaciejKaszynski: Actually you do support it with the [configuration_adapter.cpp](https://github.com/eclipse-score/lifecycle/pull/356/changes#diff-ad9383f7b5762399fbcc3789223d1d08e4ab8ed369de1c7ee198cb2956fe878b)

scripts/config_mapping/integration_tests.py
```diff
@@ -28,6 +28,7 @@
     script_dir.parent.parent
```
MaciejKaszynski: I think because this file is referenced like this I wonder if the bazel test might not recognise cache misses correctly 🤔. I also don't like storing all of the location of these files as these variables.

I think we can use pytest args and make a custom conftest.py so that the location of this is taken from the args of the pytest command.

I tested this idea out with this patch and the test passes, do you think it would make sense to do it this way?
[staged-changes.patch](https://github.com/user-attachments/files/30462182/staged-changes.patch)

scripts/config_mapping/lifecycle_config.py
```diff
@@ -289,7 +300,9 @@ def gen_config(output_dir, config, input_filename, schema_version=None):
             "ready_timeout": depl_cfg["ready_timeout"],
             "shutdown_timeout": depl_cfg["shutdown_timeout"],
             "bin_dir": depl_cfg["bin_dir"],
-            "working_dir": depl_cfg["working_dir"],
+            # Default the working directory to bin_dir (the directory the
+            # executable lives in) when not set explicitly.
```
MaciejKaszynski, line 303: Can you update the user docs to inform of the default value https://eclipse-score.github.io/lifecycle/main/score/launch_manager/docs/user_guide/configuration.html#deployment-config-object

tests/integration/sandbox_options/control_daemon_mock.cpp
MaciejKaszynski: Think this is actually causing the error in https://github.com/eclipse-score/lifecycle/actions/runs/30273517385/job/90001693926?pr=356#step:4:1286

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,342 @@
...
+        << (expected.priority.has_value() ? "" : " (default)") << " but got " << param.sched_priority;
+    if (param.sched_priority != expected_priority)
+    {
+        pass = false;
+    }
+
+    return pass;
+}
+
+bool verify_sandbox_options()
```
MaciejKaszynski: I think this is a bit long for thet test, can you separate this into another header file like `verify_sandbox.hpp`

Then would be nice if the API was 
```cpp
::testing::AssertionResult verifySandbox(ExpectedValues)
```

And the test logic would simply be

```cpp
EXPECT_TRUE(verifySandbox(ExpectedValues {
    .uid = xxx,
    .gid = xxx
     ...
}));

```

tests/utils/testing_utils/run_until_file_deployed.py
```diff
@@ -64,12 +67,27 @@ def run_until_file_deployed(
...
+            assert res == 0, "Couldn't kill process with SIGTERM"
+
+            # Poll for the process to actually terminate instead of assuming a
+            # single fixed grace period is enough. Termination timing varies
+            # with system load, so a fixed sleep is flaky.
+            stop_deadline = time.monotonic() + stop_timeout_s
+            while proc.is_running() and time.monotonic() < stop_deadline:
+                time.sleep(poll_interval_s)
+
+            if proc.is_running():
```
MaciejKaszynski: I think there is a `proc.stop()` method that handles the SIGTERM -> SIGKILL escalation. This is also being changed in https://github.com/eclipse-score/lifecycle/pull/328/changes#diff-5a2398a3ecf6f7c0d805b90182ce8c998bee6de87256a9379052fb919e55c202

would it be breaking to not do this change for now, and let the subsequent PR handle that?

tests/integration/sandbox_options/sandbox_options_process.cpp
```diff
@@ -0,0 +1,342 @@
...
+        value = arg.substr(prefix.size());
+        return true;
+    }
+    return false;
+}
+
+// Populate 'out' from the command line arguments. Only the options that are actually present are
+// set; any option left out stays unset and is therefore not verified. Returns false if an
+// unrecognized option is encountered.
+bool parse_arguments(int argc, char** argv, ExpectedValues& out)
```
MaciejKaszynski, line 90: Are you sure you want to have the process sandbox configs parsed from args?

Feel like making the ExpectedValue struct hard coded in the test code basically does the same thing, otherwise the values are just hard coded in the json config and you also need logic to parse the args.

MaciejKaszynski (changes requested):

tests/integration/sandbox_options/sandbox_options.json
```diff
@@ -0,0 +1,158 @@
...
+        "sandbox_options_process_b": {
+            "component_properties": {
+                "binary_name": "sandbox_options_process_b",
+                "application_profile": {
+                    "application_type": "Reporting"
+                },
+                "process_arguments": [
+                    "--uid=0",
+                    "--gid=0",
+                    "--supplementary-groups=",
```
NicolasFussberger:
```suggestion

```

tests/integration/sandbox_options/sandbox_options_process.cpp
NicolasFussberger: The reason was that we'll likely need multiple combinations of values to be tested (e.g. empty supplementary group list, non-empty supplementary group list, etc.), which would be difficult if this is not configurable but hard coded

tests/integration/sandbox_options/sandbox_options.json
TimoSteuerwaldETAS: Fixed in sandbox_options_process.cpp

tests/integration/sandbox_options/sandbox_options.json
TimoSteuerwaldETAS: Fixed.

tests/integration/sandbox_options/control_daemon_mock.cpp
TimoSteuerwaldETAS: Most probably this was related to the missing `"is_self_terminating": true` option in the `application_profile`.

It is also passing now in the pipeline: https://github.com/eclipse-score/lifecycle/actions/runs/30453507338/job/90581144811#step:4:4319

tests/utils/testing_utils/run_until_file_deployed.py
TimoSteuerwaldETAS: After discussion with @danth I rebased with main, as there is since yesterday already an updated version of this file. However I added some assertions regarding timeout during shutting down lcm to be more explicit.

TimoSteuerwaldETAS:
During repeated testing via
```
bazel test //tests/integration/sandbox_options:sandbox_options \
  --config=x86_64-linux --verbose_failures --nocache_test_results --runs_per_test=100
```
lcm crashed during shutdown. As this is unrelated to the test itself, I have opened this issue #397

tests/integration/sandbox_options/verify_sandbox.hpp
```diff
@@ -0,0 +1,213 @@
...
+/// @param[in] expected The expected sandbox option values; unset options are not verified.
+/// @return AssertionSuccess if every set expectation matches, otherwise AssertionFailure carrying
+///         a description of all mismatches.
+inline ::testing::AssertionResult verifySandbox(const ExpectedValues& expected)
+{
+    std::ostringstream failures;
+
+    if (expected.uid.has_value() || expected.gid.has_value())
+    {
+        TEST_STEP("Verify uid and gid")
```
MaciejKaszynski: I think it might be a bit confusing to add the `TEST_STEP` inside another function in another header. 
Also if something failed due to uid, it wouldn't show in the `Verify uid and gid` step as you always surface it in the `tests/integration/sandbox_options/sandbox_options_process.cpp` I would just remove the steps.

tests/integration/sandbox_options/verify_sandbox.hpp
```diff
@@ -0,0 +1,213 @@
...
+    }
+}
+
+/// @brief Verify that this process runs with the given sandbox options applied.
+/// @param[in] expected The expected sandbox option values; unset options are not verified.
+/// @return AssertionSuccess if every set expectation matches, otherwise AssertionFailure carrying
+///         a description of all mismatches.
+inline ::testing::AssertionResult verifySandbox(const ExpectedValues& expected)
+{
+    std::ostringstream failures;
```
MaciejKaszynski, line 107: Think you can basically use the `AssertionFailure` as the stream. e.g. https://godbolt.org/z/v6zc33h8o. But works either way so resolve if you don't mind

MaciejKaszynski (changes requested):

tests/integration/sandbox_options/verify_sandbox.hpp
TimoSteuerwaldETAS: That won't work in my use case. An empty AssertionFailure is still an AssertionFailure.

MaciejKaszynski (dismissed):

NicolasFussberger:
New test fails when run in devcontainer
Error: `sched_setscheduler() failed (errno 1)` (Operation not permitted)

@TimoSteuerwaldETAS Can you modify the devcontainer json to make sure this works?

tests/utils/testing_utils/run_until_file_deployed.py
NicolasFussberger: I think now after your latest rebase its no longer used for checking process state.
So now the documentation update is no longer correct

tests/integration/sandbox_options/BUILD
```diff
@@ -0,0 +1,85 @@
...
+        "sandbox_options_process.cpp",
+        "verify_sandbox.hpp",
+    ],
+    deps = [
+        "//score/launch_manager:control_cc",
+        "//score/launch_manager:lifecycle_cc",
+        "//tests/utils/test_helper",
+        "@googletest//:gtest_main",
+    ],
+)
```
NicolasFussberger: Suggestion:

Replace
```starlark
cc_binary(
    name = "sandbox_options_process_a",
    srcs = [
        "sandbox_options_process.cpp",
        "verify_sandbox.hpp",
    ],
    deps = [
        "//score/launch_manager:control_cc",
        "//score/launch_manager:lifecycle_cc",
        "//tests/utils/test_helper",
        "@googletest//:gtest_main",
    ],
)

cc_binary(
    name = "sandbox_options_process_b",
    srcs = [
        "sandbox_options_process.cpp",
        "verify_sandbox.hpp",
    ],
    deps = [
        "//score/launch_manager:control_cc",
        "//score/launch_manager:lifecycle_cc",
        "//tests/utils/test_helper",
        "@googletest//:gtest_main",
    ],
)

cc_binary(
    name = "sandbox_options_process_c",
    srcs = [
        "sandbox_options_process.cpp",
        "verify_sandbox.hpp",
    ],
    deps = [
        "//score/launch_manager:control_cc",
        "//score/launch_manager:lifecycle_cc",
        "//tests/utils/test_helper",
        "@googletest//:gtest_main",
    ],
)
```

with 

```starlark
[
    cc_binary(
        name = "sandbox_options_process_{}".format(suffix),
        srcs = [
            "sandbox_options_process.cpp",
            "verify_sandbox.hpp",
        ],
        deps = [
            "//score/launch_manager:control_cc",
            "//score/launch_manager:lifecycle_cc",
            "//tests/utils/test_helper",
            "@googletest//:gtest_main",
        ],
    )
    for suffix in ("a", "b", "c")
]
```

tests/integration/sandbox_options/control_daemon_mock.cpp
NicolasFussberger: Why is the separate Running and Off RunTarget needed for the test?
Can we not just put these processes into Startup and remove the control_daemon_mock altogether?

tests/integration/sandbox_options/BUILD
TimoSteuerwaldETAS: Good idea! 🚀 Thanks for pointing out!

TimoSteuerwaldETAS:
> New test fails when run in devcontainer Error: `sched_setscheduler() failed (errno 1)` (Operation not permitted)
> 
> @TimoSteuerwaldETAS Can you modify the devcontainer json to make sure this works?

@NicolasFussberger I tried, for proxy reasons I am unable to use devcontainer. So I can not verify it. Please give it a try.

tests/integration/sandbox_options/control_daemon_mock.cpp
TimoSteuerwaldETAS: Moving the processes to Startup state would remove the explicit `EXPECT` for the run target transition ( `ActivateRunTarget("Running")` ) from the test, which I don't like. If something is wrong with the run target under test, I do not want to dig in the production code, instead a simple assertion should fail.

Furthermore who writes then the `test_end file` on which `run_until_file_deployed` is waiting?
Of course `run_until_file_deployed(..)` could also wait on the list of files which are used in `assert_test_results(..)`, but then you are no longer verifying that all children processes have finished immediately after switching to "Off".

For both reasons I would like to keep it as it is.

What I could do later on (on the other branch): Get rid of the test specific `control_daemon_mock.cpp` (for most tests) and create a common version in `test_helper` subfolder to reduce the duplication.

patches/BUILD
```diff
@@ -1 +1,18 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
 exports_files(["hedron_compile_commands_env_vars.patch"])
+
+exports_files(["score_bazel_tools.patch"])
```
NicolasFussberger: These patches do no longer exist in the repo

tests/integration/sandbox_options/control_daemon_mock.cpp
NicolasFussberger: We can merge it for now, but I think we need to keep these tests minimal for what they testing. E.g. the control client API will be heavily changed and we need to update this test, even though the test scope has nothing to do with ControlClient or RunTarget transitions.

> explicit EXPECT for the run target transition

You will still only see an assertion failure for a transition to Running. You will anyway need to dig through the config and then the logs to see why transition has failed. I do not see a huge benefit here.

> who writes then the test_end file

I would just create a `touch` command as a component that depends on the sandbox processes. Then the file is only written if the sandbox processes were launched successfully.
The Off state is also not needed in the config or the explicit transition to Off state.
When launch_manager receives a SIGTERM it will terminate the processes.

NicolasFussberger:
> > New test fails when run in devcontainer Error: `sched_setscheduler() failed (errno 1)` (Operation not permitted)
> > @TimoSteuerwaldETAS Can you modify the devcontainer json to make sure this works?
> 
> @NicolasFussberger I tried, for proxy reasons I am unable to use devcontainer. So I can not verify it. Please give it a try.

Not working for me. Also passing '--privileged' in devcontainer.json runArgs did not make a difference.

tests/utils/testing_utils/run_until_file_deployed.py
TimoSteuerwaldETAS: Fixed.

tests/integration/sandbox_options/control_daemon_mock.cpp
TimoSteuerwaldETAS: Fixed.

patches/BUILD
TimoSteuerwaldETAS: Fixed.

NicolasFussberger (approved):

PR #362, opened 2026-07-23
Add cpp_supervised_app_simple example for the Alive API

### LINKED ISSUE #234: Create example app for Alive API
Depends-On: https://github.com/eclipse-score/lifecycle/pull/229

### What

Create a new example app that showcases the Alive API of the launch_manager.

### Acceptance Criteria (DoD)

New example app showcasing the alive API

### How

* Example app to be put under https://github.com/eclipse-score/lifecycle/tree/main/examples
* Create new folder cpp_supervised_app_simple for the app
* The app roughly looks like this:

```cpp
    score::mw::lifecycle::Alive alive("instance");

    // report Running state
    // ...

    // Report alive notifications
    while(!exitRequested) {
        alive.ReportAlive();
        std::this_thread::sleep_for(std::chrono::milliseconds(<Sleep-Time>));
    }

    return 0;
}
```
 - Maybe pass the <Sleep-Time> via CLI (e.g. "-s <sleep-ms>")
 - Setup signal handling for handling SIGTERM
 - Report Running state using lifecycle client
 - Report alive notifications in a look until shutdown is requested
 - Add the new example app to the existing demo configuration https://github.com/eclipse-score/lifecycle/blob/main/examples/demo_verification/lifecycle_demo_test.json and make sure everything still works

Narasimha2211:
Closes #234

## What

Adds `examples/cpp_supervised_app_simple`, a minimal example of `score::mw::lifecycle::Alive`: construct `Alive`, report `Running`, then call `ReportAlive()` in a loop until `SIGTERM`/`SIGINT`. No deadline/checkpoint supervision — that's what distinguishes it from the existing `cpp_supervised_app`, which uses the `HealthMonitor`/`DeadlineMonitor` API instead.

## How

- `main.cpp` / `BUILD` are modeled closely on `cpp_supervised_app`'s structure (signal handling, `set_process_name`, `getopt`-based CLI with `-s <sleep-ms>`), swapping the `HealthMonitor` block for `score::mw::lifecycle::Alive` + `report_running()`. Depends on the `//score/launch_manager:alive_cc` and `:lifecycle_cc` aliases — the same public-alias pattern the existing examples already use, since the underlying `alive`/`lifecycle_client` targets have restricted visibility. `Alive`'s constructor can throw per its documented contract, so it's wrapped in `try`/`catch` for a clean error message instead of an unhandled-exception `terminate()`.
- Registered as a `cpp_supervised_app_simple` component in `examples/demo_verification/lifecycle_demo_test.json`, added to the `Running` run target's `depends_on`. Left `application_type` unset (defaults to `"Reporting"`, same as `lifecycle_app`) since this app only demonstrates the alive heartbeat, not deadline-checkpoint supervision — happy to switch to `Reporting_And_Supervised` if that's the preferred classification.
- Added the binary to `demo_verification/BUILD`'s `integration_test` `binaries` list and to `test_examples.py`'s `_DEMO_APPS`.
- Documented the app in `examples/README.md`'s table.

## A bug I found and fixed along the way

Wiring this into the demo config surfaced a real, pre-existing bug: `cpp_supervised_app_simple` is a literal prefix-superstring of `cpp_supervised_app`, and `test_examples.py`'s process matching was a plain unanchored substring `grep` over `/proc/*/cmdline`. That means `grep -qa 'cpp_supervised_app'` would also match `cpp_supervised_app_simple`'s cmdline — breaking `_assert_not_running(target, "cpp_supervised_app")` right after the existing SIGKILL crash-recovery step (it would spuriously find the still-alive `_simple` process and report the killed app as still running, false-negating that assertion).

I verified this concretely (not just by inspection) by simulating `/proc/*/cmdline`-style files locally — NUL-separated argv, both bare-name and full-path forms — and reproducing the false match. Then I fixed `_assert_running`/`_assert_not_running`/`_send_signal` to turn cmdline's NUL-separated argv into one-token-per-line and anchor the match (`^binary$` or `^.*/binary$`) instead of a raw substring search, and re-ran the same simulation against the fixed logic to confirm it: distinguishes the two names correctly, still matches full-path `argv[0]`, and doesn't self-match the invoking shell's own command line (the anchoring subsumes the original bracket-class self-match-avoidance trick, so I dropped that in favor of the anchor).

## Verification note

This repo requires Bazel, which isn't available in my environment, so I could not build or run the integration test end-to-end. What I *did* verify:
- Read the actual headers (`score/launch_manager/src/alive/src/alive.h`, `.../lifecycle_client/src/report_running.h`) to confirm the API surface, include paths, and target/alias visibility.
- Modeled the new app closely on the existing, presumably-working `cpp_supervised_app` example.
- Empirically simulated the `test_examples.py` matching-logic bug and fix outside Bazel, as described above.
- `python3 -m json.tool` on the modified `lifecycle_demo_test.json` and `ast.parse` on `test_examples.py` — both syntactically valid.

CI should be treated as the real verification here — happy to iterate on anything that doesn't hold up once it actually builds and runs.

### DIFF
--- a/examples/cpp_supervised_app_simple/main.cpp
+++ b/examples/cpp_supervised_app_simple/main.cpp
@@ -1,0 +1,126 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include <unistd.h>
+#include <atomic>
+#include <chrono>
+#include <csignal>
+#include <cstdint>
+#include <iostream>
+#include <optional>
+#include <thread>
+
+#ifdef __linux__
+#include <linux/prctl.h>
+#include <sys/prctl.h>
+#endif
+
+#include <score/mw/lifecycle/alive.h>
+#include <score/mw/lifecycle/report_running.h>
+
+/// @brief CLI configuration options for the cpp_supervised_app_simple process
+struct Config
+{
+    std::uint32_t sleepInMs{50};
+};
+
+std::optional<Config> parseOptions(int argc, char* const* argv) noexcept
+{
+    Config config{};
+    int c;
+    while ((c = getopt(argc, argv, "s:h")) != -1)
+    {
+        switch (static_cast<char>(c))
+        {
+            case 's':
+                config.sleepInMs = std::stoi(optarg);
+                break;
+
+            case 'h':
+                std::cout << "Usage: \n\
+                            -s <Sleep time in ms between alive notifications>\n";
+                return std::nullopt;
+
+            default:
+                break;
+        }
+    }
+    return config;
+}
+
+std::atomic<bool> exitRequested{false};
+void signalHandler(int signal)
+{
+    if (signal == SIGINT || signal == SIGTERM)
+    {
+        exitRequested = true;
+    }
+}
+
+void set_process_name()
+{
+    const char* identifier = getenv("PROCESSIDENTIFIER");
+    if (identifier != nullptr)
+    {
+#if defined(__QNXNTO__)
+        if (pthread_setname_np(pthread_self(), identifier) != 0)
+        {
+            std::cerr << "Failed to set QNX thread name" << std::endl;
+        }
+#elif defined(__linux__)
+        if (prctl(PR_SET_NAME, identifier) < 0)
+        {
+            std::cerr << "Failed to set process name to " << identifier << std::endl;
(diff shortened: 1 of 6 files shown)


examples/cpp_supervised_app_simple/main.cpp
```diff
@@ -0,0 +1,126 @@
...
+std::atomic<bool> exitRequested{false};
+void signalHandler(int signal)
+{
+    if (signal == SIGINT || signal == SIGTERM)
+    {
+        exitRequested = true;
+    }
+}
+
+void set_process_name()
```
NicolasFussberger, line 70: I think we could remove this. This is not required for the functionality. In some other sample app we had this only for some demo purpose.

NicolasFussberger (changes requested):
Looks good to me @Narasimha2211. Thanks for your contribution!
I only added a minor comment.

Please ensure you have an eclipse account linked to your github account so that we can merge your contribution

MaciejKaszynski:
Closing as committer is not active, committer needs to follow the workflow described in https://eclipse-score.github.io/score/main/contribute/general/contribution_attribution.html#.

PR #367, opened 2026-07-24
Carve FixedSizeQueue out of MpscBoundedQueue

paulquiring:
Carve FixedSizeQueue out of MpscBoundedQueue for reuse.
MpscBoundedQueues capacity can be set now at runtime, to allow
setting the capacity based on configuration.

PR #370, opened 2026-07-24
Graph refactor initial components

WilliamRoebuck:
This PR introduces some initial components required for an upcoming refactor to the `Graph` and `ProcessInfoNode` classes. Firstly, we want to introduce some components that are more easily understood and unit tested. The next PR will integrate these into existing code. Excluding interfaces and data-only classes, the new components are:
- `ComponentEventQueue` - Allows us to handle component state changes sequentially in Graph
- `ProcessMonitor` - Decouples `ProcessInfoNode` from `Graph` and allows us to simplify its APIs

The refactor aims to address the following concerns:
- Race conditions throughout `Graph`
- Support for different component types than POSIX processes, e.g. containers
- Circular dependencies impeding unit test development
- Outdated "PG state" based transition logic
- Overlapping & confusing class responsibilities

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,2 +1,2 @@
 Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplus.*, clang-analyzer-security.*, cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-cstyle-cast, hicpp-* '
-WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*,hicpp-*'
+WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -2,4 +2,4 @@ Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplu
 WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*,hicpp-*'
 CheckOptions:
   - key: 'hicpp-special-member-functions.AllowSoleDefaultDtor'
-  - value: 1
+  - value: 1 # This allows us to make interfaces that have virtual destructor without the other functions
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,2 +1,2 @@
 Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplus.*, clang-analyzer-security.*, cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-cstyle-cast, hicpp-* '
-WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*
+WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*'
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,2 +1,5 @@
 Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplus.*, clang-analyzer-security.*, cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-cstyle-cast, hicpp-* '
-WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*'
+WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*,hicpp-*'
+CheckOptions:
+- key: 'hicpp-special-member-functions.AllowSoleDefaultDtor'
+- value: 1
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,5 +1,5 @@
 Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplus.*, clang-analyzer-security.*, cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-cstyle-cast, hicpp-* '
 WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*,hicpp-*'
 CheckOptions:
-- key: 'hicpp-special-member-functions.AllowSoleDefaultDtor'
-- value: 1
+  - key: 'hicpp-special-member-functions.AllowSoleDefaultDtor'
+  - value: 1
--- a/score/launch_manager/src/daemon/src/common/constants.hpp
+++ b/score/launch_manager/src/daemon/src/common/constants.hpp
@@ -11,7 +11,6 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
-
 #ifndef CONSTANTS_HPP_INCLUDED
 #define CONSTANTS_HPP_INCLUDED
 
@@ -19,11 +18,14 @@
 #include <cstddef>
 #include <cstdint>
 
-namespace score {
+namespace score
+{
 
-namespace lcm {
+namespace lcm
+{
 
-namespace internal {
+namespace internal
+{
 
 // coverity[autosar_cpp14_a0_1_1_violation:INTENTIONAL] These are constants that are used globally.
 constexpr std::size_t kMaxArg = 20U;  ///< Maximum number of arguments
@@ -36,29 +38,42 @@ constexpr std::size_t kArgvArraySize =
 constexpr std::size_t kEnvArraySize =
     kMaxEnv + 1U;  ///< As required by posix we need extra space in envp_ for NULL pointer
 
-constexpr std::chrono::milliseconds kMaxQueueDelay{500};  ///< The maximum time to wait trying to add items to, or get items from, a queue
-constexpr std::chrono::milliseconds kGraphTimeout{10000};  ///< Timeout duration for graph operations.
+constexpr std::chrono::milliseconds kMaxQueueDelay{
+    500};  ///< The maximum time to wait trying to add items to, or get items from, a queue
+constexpr std::chrono::milliseconds kGraphTimeout{10000};   ///< Timeout duration for graph operations.
 constexpr std::chrono::milliseconds kMaxSigKillDelay{500};  ///< The maximum time to wait for a process termination
 
-constexpr std::chrono::milliseconds kControlClientPollingDelay{1};  ///< Time Control Client will wait during polling for acknowledgement
+constexpr std::chrono::milliseconds kControlClientPollingDelay{
+    1};  ///< Time Control Client will wait during polling for acknowledgement
 
-constexpr std::chrono::milliseconds kMaxRunningDelay{1000};  ///< report_running() API will wait for Launch Manager to respond
+constexpr std::chrono::milliseconds kMaxRunningDelay{
+    1000};  ///< report_running() API will wait for Launch Manager to respond
 
-constexpr std::chrono::milliseconds kControlClientMaxIpcDelay{500};  ///< The maximum time to wait, when trying to communicate with LCM. When this time is exceeded kCommunicationError will be returned
+constexpr std::chrono::milliseconds kControlClientMaxIpcDelay{
+    500};  ///< The maximum time to wait, when trying to communicate with LCM. When this time is exceeded
+           ///< kCommunicationError will be returned
 constexpr std::chrono::milliseconds kControlClientBgThreadSleepTime{100};
 
-enum class ControlClientLimits : uint16_t {
-    kControlClientMaxInstances =
-        256U,  ///< Maximum number of ControlClient instances that should be created by state manager. If state manager create more instances than kMaxInstances, those instances will always return kCommunicationError when used
+enum class ControlClientLimits : uint16_t
+{
+    kControlClientMaxInstances = 256U,  ///< Maximum number of ControlClient instances that should be created by state
+                                        ///< manager. If state manager create more instances than kMaxInstances, those
+                                        ///< instances will always return kCommunicationError when used
     kControlClientMaxRequests =
-        512U  ///< Maximum number of active requests, for example SetState call, that ControlClient instance can send to LCM. If that number is exceeded ControlClient API will return kFailed, until one of the current requests is completed by LCM
+        512U  ///< Maximum number of active requests, for example SetState call, that ControlClient instance can send to
+              ///< LCM. If that number is exceeded ControlClient API will return kFailed, until one of the current
+              ///< requests is completed by LCM
 };
 
-enum class ProcessLimits : std::uint32_t {
+enum class ProcessLimits : std::uint32_t
+{
     kMaxProcesses = 1024U,    ///< Maximum number of processes allowed
     kNumWorkerThreads = 32U,  ///< Maximum number of worker threads allowed
     maxLocalBuffSize = 32U    ///< Maximum size for local buffer
 };
+
+/// @brief Capacity of the ComponentEventQueue.
+constexpr std::size_t kComponentEventQueueSize = 1024U;
 
 }  // namespace internal
 
--- a/score/launch_manager/src/daemon/src/common/constants.hpp
+++ b/score/launch_manager/src/daemon/src/common/constants.hpp
@@ -11,6 +11,7 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
+
 #ifndef CONSTANTS_HPP_INCLUDED
 #define CONSTANTS_HPP_INCLUDED
 
@@ -18,14 +19,11 @@
 #include <cstddef>
 #include <cstdint>
 
-namespace score
-{
+namespace score {
 
-namespace lcm
-{
+namespace lcm {
 
-namespace internal
-{
+namespace internal {
 
 // coverity[autosar_cpp14_a0_1_1_violation:INTENTIONAL] These are constants that are used globally.
 constexpr std::size_t kMaxArg = 20U;  ///< Maximum number of arguments
@@ -38,35 +36,25 @@ constexpr std::size_t kArgvArraySize =
 constexpr std::size_t kEnvArraySize =
     kMaxEnv + 1U;  ///< As required by posix we need extra space in envp_ for NULL pointer
 
-constexpr std::chrono::milliseconds kMaxQueueDelay{
-    500};  ///< The maximum time to wait trying to add items to, or get items from, a queue
-constexpr std::chrono::milliseconds kGraphTimeout{10000};   ///< Timeout duration for graph operations.
+constexpr std::chrono::milliseconds kMaxQueueDelay{500};  ///< The maximum time to wait trying to add items to, or get items from, a queue
+constexpr std::chrono::milliseconds kGraphTimeout{10000};  ///< Timeout duration for graph operations.
 constexpr std::chrono::milliseconds kMaxSigKillDelay{500};  ///< The maximum time to wait for a process termination
 
-constexpr std::chrono::milliseconds kControlClientPollingDelay{
-    1};  ///< Time Control Client will wait during polling for acknowledgement
+constexpr std::chrono::milliseconds kControlClientPollingDelay{1};  ///< Time Control Client will wait during polling for acknowledgement
 
-constexpr std::chrono::milliseconds kMaxRunningDelay{
-    1000};  ///< report_running() API will wait for Launch Manager to respond
+constexpr std::chrono::milliseconds kMaxRunningDelay{1000};  ///< report_running() API will wait for Launch Manager to respond
 
-constexpr std::chrono::milliseconds kControlClientMaxIpcDelay{
-    500};  ///< The maximum time to wait, when trying to communicate with LCM. When this time is exceeded
-           ///< kCommunicationError will be returned
+constexpr std::chrono::milliseconds kControlClientMaxIpcDelay{500};  ///< The maximum time to wait, when trying to communicate with LCM. When this time is exceeded kCommunicationError will be returned
 constexpr std::chrono::milliseconds kControlClientBgThreadSleepTime{100};
 
-enum class ControlClientLimits : uint16_t
-{
-    kControlClientMaxInstances = 256U,  ///< Maximum number of ControlClient instances that should be created by state
-                                        ///< manager. If state manager create more instances than kMaxInstances, those
-                                        ///< instances will always return kCommunicationError when used
+enum class ControlClientLimits : uint16_t {
+    kControlClientMaxInstances =
+        256U,  ///< Maximum number of ControlClient instances that should be created by state manager. If state manager create more instances than kMaxInstances, those instances will always return kCommunicationError when used
     kControlClientMaxRequests =
-        512U  ///< Maximum number of active requests, for example SetState call, that ControlClient instance can send to
-              ///< LCM. If that number is exceeded ControlClient API will return kFailed, until one of the current
-              ///< requests is completed by LCM
+        512U  ///< Maximum number of active requests, for example SetState call, that ControlClient instance can send to LCM. If that number is exceeded ControlClient API will return kFailed, until one of the current requests is completed by LCM
 };
 
-enum class ProcessLimits : std::uint32_t
-{
+enum class ProcessLimits : std::uint32_t {
     kMaxProcesses = 1024U,    ///< Maximum number of processes allowed
     kNumWorkerThreads = 32U,  ///< Maximum number of worker threads allowed
     maxLocalBuffSize = 32U    ///< Maximum size for local buffer
--- a/score/launch_manager/src/daemon/src/common/constants.hpp
+++ b/score/launch_manager/src/daemon/src/common/constants.hpp
@@ -72,9 +72,6 @@ enum class ProcessLimits : std::uint32_t
     maxLocalBuffSize = 32U    ///< Maximum size for local buffer
 };
 
-/// @brief Capacity of the ComponentEventQueue.
-constexpr std::size_t kComponentEventQueueSize = 1024U;
-
 }  // namespace internal
 
 }  // namespace lcm
--- a/score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
+++ b/score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
@@ -14,15 +14,9 @@
 #ifndef SCORE_LCM_COMPONENT_EVENT_HPP_INCLUDED
 #define SCORE_LCM_COMPONENT_EVENT_HPP_INCLUDED
 
-#include <atomic>
-#include <chrono>
 #include <cstdint>
-#include <optional>
 #include <variant>
 
-#include "score/mw/launch_manager/common/concurrency/concurrency_error_domain.hpp"
-#include "score/mw/launch_manager/common/concurrency/mpsc_bounded_queue.hpp"
-#include "score/mw/launch_manager/common/constants.hpp"
 #include "score/mw/launch_manager/common/identifier_hash.hpp"
 #include "score/mw/launch_manager/process_group_manager/details/icomponent.hpp"
 
@@ -64,67 +58,6 @@ struct SupervisionFailure
 /// identifier is needed to route most events. SupervisionFailure is routed by process identifier.
 using ComponentEvent = std::
     variant<ActivationSuccessful, ActivationFailed, DeactivationComplete, UnexpectedTermination, SupervisionFailure>;
-
-/// @brief Queue of ComponentEvents produced by worker/OS-handler threads and consumed
-/// exclusively by the main thread, backed by a fixed-capacity MpscBoundedQueue.
-class ComponentEventQueue final
-{
-  public:
-    ComponentEventQueue() = default;
-    ~ComponentEventQueue()
-    {
-        stop();
-    }
-
-    ComponentEventQueue(const ComponentEventQueue&) = delete;
-    ComponentEventQueue& operator=(const ComponentEventQueue&) = delete;
-    ComponentEventQueue(ComponentEventQueue&&) = delete;
-    ComponentEventQueue& operator=(ComponentEventQueue&&) = delete;
-
-    /// @brief Enqueues an event. If the queue is full, the event is dropped immediately and
-    /// getOverflow() will subsequently return true.
-    void push(ComponentEvent&& event)
-    {
-        auto result = queue_.push(std::move(event));
-        if (!result.has_value() && result.error() == ConcurrencyErrc::kOverflow)
-        {
-            overflow_.store(true, std::memory_order_release);
-        }
-    }
-
-    /// @brief Waits up to timeout for at least one event to become available.
-    /// @param timeout Maximum time to wait. A `timeout` of zero means "check once, don't
-    ///        block", NOT "wait forever" -- see MpscBoundedQueue::wait().
-    /// @return true if an event is available (drain via getNextEvent()), false on timeout.
-    bool waitForEvents(std::chrono::milliseconds timeout)
-    {
-        return queue_.wait(timeout).has_value();
-    }
-
-    /// @brief Returns the next available event without blocking, or std::nullopt if none right
-    /// now. Call repeatedly until nullopt to drain everything currently queued.
-    std::optional<ComponentEvent> getNextEvent()
-    {
-        return queue_.tryPop();
-    }
-
-    /// @return True if an event was ever dropped due to the queue being full.
-    bool getOverflow() const
-    {
-        return overflow_.load(std::memory_order_acquire);
-    }
-
-    /// @brief Permanently marks the queue stopped and wakes any thread currently blocked in
-    /// waitForEvents().
-    void stop()
-    {
-        queue_.stop();
-    }
-
-  private:
-    MpscBoundedQueue<ComponentEvent, kComponentEventQueueSize> queue_;
-    std::atomic<bool> overflow_{false};
-};
 
 }  // namespace score::lcm::internal
 
--- a/score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
+++ b/score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
@@ -1,0 +1,131 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#ifndef SCORE_LCM_COMPONENT_EVENT_HPP_INCLUDED
+#define SCORE_LCM_COMPONENT_EVENT_HPP_INCLUDED
+
+#include <atomic>
+#include <chrono>
+#include <cstdint>
+#include <optional>
+#include <variant>
+
+#include "score/mw/launch_manager/common/concurrency/concurrency_error_domain.hpp"
+#include "score/mw/launch_manager/common/concurrency/mpsc_bounded_queue.hpp"
+#include "score/mw/launch_manager/common/constants.hpp"
+#include "score/mw/launch_manager/common/identifier_hash.hpp"
+#include "score/mw/launch_manager/process_group_manager/details/icomponent.hpp"
+
+namespace score::lcm::internal
+{
+
+/// @brief A node finished activating successfully.
+struct ActivationSuccessful
+{
+    uint32_t node_index;
+};
+
+/// @brief A node failed to activate.
+struct ActivationFailed
+{
+    uint32_t node_index;
+    IComponent::ComponentError reason;
+};
+
+/// @brief A node finished deactivating.
+struct DeactivationComplete
+{
+    uint32_t node_index;
+};
+
+/// @brief A node terminated without having been requested to.
+struct UnexpectedTermination
+{
+    uint32_t node_index;
+};
+
+/// @brief Alive supervision has failed for the given process identifier.
+struct SupervisionFailure
+{
+    IdentifierHash process_identifier;
+};
+
+/// @brief A graph-relevant state change. There is only ever a single graph, so no process-group
+/// identifier is needed to route most events. SupervisionFailure is routed by process identifier.
+using ComponentEvent = std::
+    variant<ActivationSuccessful, ActivationFailed, DeactivationComplete, UnexpectedTermination, SupervisionFailure>;
+
+/// @brief Queue of ComponentEvents produced by worker/OS-handler threads and consumed
+/// exclusively by the main thread, backed by a fixed-capacity MpscBoundedQueue.
+class ComponentEventQueue final
+{
+  public:
+    ComponentEventQueue() = default;
+    ~ComponentEventQueue()
+    {
+        stop();
+    }
+
+    ComponentEventQueue(const ComponentEventQueue&) = delete;
+    ComponentEventQueue& operator=(const ComponentEventQueue&) = delete;
+    ComponentEventQueue(ComponentEventQueue&&) = delete;
+    ComponentEventQueue& operator=(ComponentEventQueue&&) = delete;
+
+    /// @brief Enqueues an event. If the queue is full, the event is dropped immediately and
+    /// getOverflow() will subsequently return true.
+    void push(ComponentEvent&& event)
+    {
+        auto result = queue_.push(std::move(event));
+        if (!result.has_value() && result.error() == ConcurrencyErrc::kOverflow)
+        {
+            overflow_.store(true, std::memory_order_release);
+        }
+    }
+
+    /// @brief Waits up to timeout for at least one event to become available.
+    /// @param timeout Maximum time to wait. A `timeout` of zero means "check once, don't
+    ///        block", NOT "wait forever" -- see MpscBoundedQueue::wait().
+    /// @return true if an event is available (drain via getNextEvent()), false on timeout.
+    bool waitForEvents(std::chrono::milliseconds timeout)
+    {
+        return queue_.wait(timeout).has_value();
+    }
+
+    /// @brief Returns the next available event without blocking, or std::nullopt if none right
+    /// now. Call repeatedly until nullopt to drain everything currently queued.
+    std::optional<ComponentEvent> getNextEvent()
+    {
+        return queue_.tryPop();
+    }
+
+    /// @return True if an event was ever dropped due to the queue being full.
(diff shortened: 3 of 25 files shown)


score/launch_manager/src/daemon/src/common/constants.hpp
```diff
@@ -11,19 +11,21 @@
...
-
 #ifndef CONSTANTS_HPP_INCLUDED
 #define CONSTANTS_HPP_INCLUDED
 
 #include <chrono>
 #include <cstddef>
 #include <cstdint>
 
-namespace score {
+namespace score
```
MaciejKaszynski: I think the namespace should be `score::mw::lifecycle` so here `score::mw::lifecycle::internal`

score/launch_manager/src/daemon/src/common/constants.hpp
```diff
@@ -11,19 +11,21 @@
...
+{
 
-namespace lcm {
+namespace lcm
```
MaciejKaszynski: We should also enable the `concat-nested-namespaces` clang-tidy rule to concatenate the nested namepsaces

score/launch_manager/src/daemon/src/common/constants.hpp
```diff
@@ -36,30 +38,43 @@ constexpr std::size_t kArgvArraySize =
...
+    500};  ///< The maximum time to wait, when trying to communicate with LCM. When this time is exceeded
+           ///< kCommunicationError will be returned
 constexpr std::chrono::milliseconds kControlClientBgThreadSleepTime{100};
 
-enum class ControlClientLimits : uint16_t {
-    kControlClientMaxInstances =
-        256U,  ///< Maximum number of ControlClient instances that should be created by state manager. If state manager create more instances than kMaxInstances, those instances will always return kCommunicationError when used
+enum class ControlClientLimits : uint16_t
+{
+    kControlClientMaxInstances = 256U,  ///< Maximum number of ControlClient instances that should be created by state
```
MaciejKaszynski: I don't think you actually touched this line but could you convert this to just a variable and have the enum become a namespace. I think then this all would be less weird.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
```diff
@@ -0,0 +1,64 @@
...
+#include <variant>
+
+#include "score/mw/launch_manager/common/identifier_hash.hpp"
+#include "score/mw/launch_manager/process_group_manager/details/icomponent.hpp"
+
+namespace score::lcm::internal
+{
+
+/// @brief A node finished activating successfully.
+struct ActivationSuccessful
```
MaciejKaszynski: I don't think this fits into process_group_manager.
I think in general we could split some stuff in this PR to something like
`component_state_manager` then maybe
`component_state_manager/process` or something like this.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+class ComponentEventQueueTest : public ::testing::Test
+{
+  protected:
+    ComponentEventQueue queue_;
+};
+
+TEST_F(ComponentEventQueueTest, WaitForEventsReturnsFalseOnEmptyQueue)
+{
+    RecordProperty("Description", "Verify waitForEvents returns false promptly when no event has been pushed.");
+    EXPECT_FALSE(queue_.waitForEvents(std::chrono::milliseconds{1}));
```
MaciejKaszynski:
```suggestion
    EXPECT_FALSE(queue_.waitForEvents(std::chrono::milliseconds{0}));
```
The queue should handle 0 here.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+}
+
+TEST_F(ComponentEventQueueTest, WaitForEventsReturnsTrueAfterPush)
+{
+    RecordProperty("Description", "Verify waitForEvents returns true once an event has been pushed.");
+    queue_.push(ActivationSuccessful{7U});
```
MaciejKaszynski: Do we need to do the comments for "Given", "Expect"
https://eclipse-score.github.io/process_description//main/process_areas/verification/guidance/verification_specification.html#test-description

score/launch_manager/src/daemon/src/process_group_manager/details/task.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+
+enum class TaskType : std::uint_least8_t
+{
+    kActivate = 0U,    /// This task is to start activation of the component
+    kDeactivate = 1U,  /// This task is to deactivate the component
+};
+
+/// @brief Work to perform on a component that can be sent to the job queue to be executed in parallel.
+struct Task
```
MaciejKaszynski: I think the name Task is too vague here `ComponentTask` would be quicker to understand without docs I think.

score/launch_manager/src/daemon/src/process_group_manager/details/run_target.hpp
```diff
@@ -0,0 +1,76 @@
...
+#define SCORE_LCM_RUN_TARGET_HPP_INCLUDED
+
+#include "score/mw/launch_manager/process_group_manager/details/icomponent.hpp"
+
+#include <atomic>
+
+namespace score::lcm::internal
+{
+
+/// @brief A virtual node in the dependency graph corresponding to a ProcessGroupState.
```
MaciejKaszynski: `RunTarget` doesn't talk in "processes" but I think it's actually the fault of `ProcessGroupState` and should be `ComponentGroupState`

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+    EXPECT_TRUE(queue_.waitForEvents(std::chrono::milliseconds{1}));
+}
+
+TEST_F(ComponentEventQueueTest, GetNextEventReturnsNulloptWhenEmpty)
```
MaciejKaszynski, line 49: In general I am a fan of making new test classes to separate the UTs, but this UT might be short enough. Up to you, just a suggestion.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+{
+    RecordProperty("Description", "Verify getNextEvent never blocks and returns nullopt when nothing is queued.");
```
MaciejKaszynski: Not sure if this tests if it "never blocks". I would change the description and not add timing tests...

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    auto event = queue_.getNextEvent();
+    ASSERT_TRUE(event.has_value());
+    ASSERT_TRUE(std::holds_alternative<ActivationFailed>(*event));
+    const auto& failed = std::get<ActivationFailed>(*event);
+    EXPECT_EQ(failed.node_index, 3U);
+    EXPECT_EQ(failed.reason, IComponent::ComponentError::kErrorBeforeReady);
+}
+
+TEST_F(ComponentEventQueueTest, GetNextEventReturnsSupervisionFailureWithPayloadIntact)
```
MaciejKaszynski, line 68: Is this testing anything new from the previous test? Seems like just testing with a different type but if it gets coverage then ignore

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+TEST_F(ComponentEventQueueTest, GetNextEventDrainsMultipleEventsInFifoOrder)
+{
+    RecordProperty("Description",
+                   "Verify the intended drain pattern -- waitForEvents() once, then getNextEvent() in a loop "
+                   "until nullopt -- returns every queued event exactly once in FIFO order.");
+    queue_.push(ActivationSuccessful{1U});
+    queue_.push(DeactivationComplete{2U});
+    queue_.push(UnexpectedTermination{3U});
+
+    ASSERT_TRUE(queue_.waitForEvents(std::chrono::milliseconds{20}));
```
MaciejKaszynski:
```suggestion
    ASSERT_TRUE(queue_.waitForEvents(std::chrono::milliseconds{0}));
```
Or remove the line?

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+    ASSERT_TRUE(event.has_value());
+    ASSERT_TRUE(std::holds_alternative<SupervisionFailure>(*event));
+    const auto& failure = std::get<SupervisionFailure>(*event);
+    EXPECT_EQ(failure.process_identifier, process_identifier);
+}
+
+TEST_F(ComponentEventQueueTest, GetNextEventDrainsMultipleEventsInFifoOrder)
+{
+    RecordProperty("Description",
+                   "Verify the intended drain pattern -- waitForEvents() once, then getNextEvent() in a loop "
```
MaciejKaszynski: Not sure if this is needed, might be better to test in the underlying queue ut.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+    /// @brief Returns the next available event without blocking, or std::nullopt if none right
+    /// now. Call repeatedly until nullopt to drain everything currently queued.
+    std::optional<ComponentEvent> getNextEvent()
+    {
+        return queue_.tryPop();
+    }
+
+    /// @return True if an event was ever dropped due to the queue being full.
+    bool getOverflow() const
```
MaciejKaszynski: What do you think about returning a bool from push? Wouldn't need the overflow_member then. Also this whole class might just become a typedef?

score/launch_manager/src/daemon/src/common/constants.hpp
MaciejKaszynski: This whole PR could update this.

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent.hpp
```diff
@@ -0,0 +1,69 @@
...
+#ifndef SCORE_LCM_ICOMPONENT_HPP_INCLUDED
+#define SCORE_LCM_ICOMPONENT_HPP_INCLUDED
+
+#include <score/stop_token.hpp>
+#include <cstdint>
+
+namespace score::lcm::internal
+{
+
+/// @brief A class that implements IComponent is a node in the configuration that can be activated or deactivated.
```
MaciejKaszynski, line 23: I like the explanation from our User Docs  'A Component is an independent, deployable software unit managed by the Launch Manager' but not sure if we should have such things here?

Also I would remove the example and say something like
'Each component shall define the implementation and semantics of what it means to be ready or not.'

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent.hpp
```diff
@@ -0,0 +1,69 @@
...
+    /// error.
+    virtual RequestResult deactivate(score::cpp::stop_token stop_token) = 0;
+    /// @brief Notify the component that it has terminated with status @p status
+    /// @returns kSuccess if the component is now ready, kWaiting if the component is waiting for a notification, or an
+    /// error if the termination was not expected.
+    virtual RequestResult tryHandleTermination(int32_t status) = 0;
+    /// @returns the index of the component in the graph.
+    virtual uint32_t getIndex() const = 0;
+    /// @returns True if the component is active in the active run target.
+    virtual bool active() const = 0;
```
MaciejKaszynski: should it be ready()?

Also I think it would be nicer to have a `ComponentState` enum and return it from a `state()` method even if this means just `kReady` or `kNotReady` for future extensions.

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_controller.hpp
```diff
@@ -0,0 +1,34 @@
...
+#ifndef SCORE_LCM_ICOMPONENT_CONTROLLER_HPP_INCLUDED
+#define SCORE_LCM_ICOMPONENT_CONTROLLER_HPP_INCLUDED
+
+#include "score/mw/launch_manager/process_group_manager/details/icomponent.hpp"
+#include "score/mw/launch_manager/process_group_manager/details/task.hpp"
+
+namespace score::lcm::internal
+{
+
+class IComponentController
```
MaciejKaszynski, line 25: briefs

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_event_receiver.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#ifndef SCORE_LCM_ICOMPONENT_EVENT_RECEIVER_HPP_INCLUDED
+#define SCORE_LCM_ICOMPONENT_EVENT_RECEIVER_HPP_INCLUDED
+
+#include "score/mw/launch_manager/process_group_manager/details/component_event.hpp"
+
+namespace score::lcm::internal
+{
+
+class IComponentEventReceiver
```
MaciejKaszynski: briefs

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.hpp
```diff
@@ -0,0 +1,50 @@
...
+    ~ProcessMonitor() override;
+
+    ProcessMonitor(const ProcessMonitor&) = delete;
+    ProcessMonitor& operator=(const ProcessMonitor&) = delete;
+    ProcessMonitor(ProcessMonitor&&) = delete;
+    ProcessMonitor& operator=(ProcessMonitor&&) = delete;
+
+    /// @brief Start work on @p task and push the result to the event queue if the task completes
+    void doWork(Task task) override;
+    /// @brief Notify @p component that it has terminated with status @p status. If this is an error or finishes a
```
MaciejKaszynski, line 39: Can you space stuff out so it's more readable.

```suggestion

    /// @brief Notify @p component that it has terminated with status @p status. If this is an error or finishes a
```

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.hpp
```diff
@@ -0,0 +1,50 @@
...
+    /// component activation, report to the event queue
+    void terminated(IComponent& component, int32_t status) override;
+
+  private:
+    void taskFinished(const Task& task, const score::cpp::expected_blank<IComponent::ComponentError>& error);
```
MaciejKaszynski: still need briefs on private :) also feel like the name should be `markFinished`.

MaciejKaszynski: Actually you might want to just make this a lambda in `doWork`

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor_UT.cpp
```diff
@@ -0,0 +1,128 @@
...
+ ********************************************************************************/
+
+#include "score/mw/launch_manager/process_group_manager/details/process_monitor.hpp"
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+using namespace testing;
+using namespace score::lcm::internal;
+
+class MockComponent : public IComponent
```
MaciejKaszynski: Can you make this into a separate header for easier re-use.

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_controller.hpp
```diff
@@ -0,0 +1,34 @@
...
+{
+  public:
+    virtual void doWork(Task task) = 0;
```
MaciejKaszynski: Are you sure about copying, seems like we could just always take a `&&` here?

MaciejKaszynski (changes requested):
Looks good, would really like to split this into separate folders because it's a bit unclear if we are at the component level or process level.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
NicolasFussberger: I think we'll need a separate flag because the location where the overflow is checked/handled is very different from the location where threads push events.
The overflow would be checked in the main loop in processgroupmanager, such that the failure can be escalated to the watchdog.

Due to the interface, I guess it cannot be simply a typedef even if we moved the overflow to the generic queue

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
```diff
@@ -0,0 +1,93 @@
...
+#include "score/mw/launch_manager/common/constants.hpp"
+#include "score/mw/launch_manager/process_group_manager/details/component_event.hpp"
+#include "score/mw/launch_manager/process_group_manager/details/icomponent_event_receiver.hpp"
+
+namespace score::lcm::internal
+{
+
+/// @brief Queue of ComponentEvents produced by worker/OS-handler threads and consumed
+/// exclusively by the main thread, backed by a fixed-capacity MpscBoundedQueue.
+class ComponentEventQueue final : public IComponentEventReceiver
```
NicolasFussberger: Should we rename this interface to e.g. IComponentEventPublisher?
It looks like its for publishing events into the queue, not actually for reading events from the queue

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
NicolasFussberger: I assume this is only a recommendation in the process.
I would say for such small tests, its probably not necessary.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
NicolasFussberger: I wonder if we should do this reorganization in a separate PR.
Then we could rename the whole process group manager folder as the process groups will vanish entirely once the new config is fully rolled out.

score/launch_manager/src/daemon/src/common/constants.hpp
NicolasFussberger: Should we use this namespace for the new code and then update the other namespaces with a separate PR?

MaciejKaszynski: I would say let's just fix what is touched with this PR

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
danth: Yes, this is only testing that the method eventually returns; it can block just not infinitely. I would just remove "never blocks" from the description.

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
danth: I would say `push` should return `bool` to indicate the overflow, even if we provide the `getOverflow` method as well. Just to draw attention to the fact that it can drop

score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
MaciejKaszynski: My suggestion:
```
score/launch_manager/src/daemon/src/component_manager
├── details
│   └── process    // Each implementation of a component has a separate sub folder
│       ├── process.hpp
│       └── process_monitor.hpp
├── icomponent.hpp  // All files need to be on the component level
└── run_target.hpp
```

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
danth: Yes, the name seems to be incorrect. Should we have two interfaces `IComponentEventPublisher` and `IComponentEventSubscriber`? And the queue would implement both

score/launch_manager/src/daemon/src/common/constants.hpp
```diff
@@ -36,30 +38,43 @@ constexpr std::size_t kArgvArraySize =
...
-enum class ProcessLimits : std::uint32_t {
+enum class ProcessLimits : std::uint32_t
+{
     kMaxProcesses = 1024U,    ///< Maximum number of processes allowed
     kNumWorkerThreads = 32U,  ///< Maximum number of worker threads allowed
     maxLocalBuffSize = 32U    ///< Maximum size for local buffer
 };
 
+/// @brief Capacity of the ComponentEventQueue.
+constexpr std::size_t kComponentEventQueueSize = 1024U;
```
danth: Maybe copy the way it is done in `MpscBoundedQueue`, and make the `ComponentEventQueue` take the capacity as a type parameter:

https://github.com/eclipse-score/lifecycle/blob/2e68cc852c3bba409d9d2be210806f6329e1cd39/score/launch_manager/src/daemon/src/common/concurrency/mpsc_bounded_queue.hpp#L40-L41

This would allow mocking the capacity in tests, having multiple queues with different capacities, etc

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
```diff
@@ -0,0 +1,94 @@
...
+    switch (task.type)
+    {
+        case TaskType::kActivate:
+            res = component.activate(task.stop_token);
+            break;
+        case TaskType::kDeactivate:
+            res = component.deactivate(task.stop_token);
+            break;
+        default:
+            break;
```
danth: Should we add `SCORE_LANGUAGE_FUTURECPP_UNREACHABLE` here? If someone defines a new task type and does not add code to handle it the behaviour is not clear

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (res.has_value())
+    {
+        taskFinished(task, {});
+    }
+    else
+    {
+        taskFinished(task, score::cpp::make_unexpected(res.error()));
+    }
```
danth: Personally I would find this easier to read if it were nested

```suggestion
    if (res.has_value())
    {
        if (res.value() == IComponent::RequestState::kWaiting)
        {
          return;
        }
        taskFinished(task, {});
    }
    else
    {
        taskFinished(task, score::cpp::make_unexpected(res.error()));
    }
```

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
```diff
@@ -0,0 +1,94 @@
...
+        event_queue_.push(DeactivationComplete{node_index});
+    }
+    else if (error.has_value())
+    {
+        event_queue_.push(ActivationSuccessful{node_index});
+    }
+    else
+    {
+        event_queue_.push(ActivationFailed{node_index, error.error()});
+    }
```
danth: This is correct but because of the naming it reads the wrong way round- "if there is an error, push successful"...

Maybe rename the parameter to something like `status` or `result`?

.clang-tidy
```diff
@@ -1,2 +1,2 @@
 Checks: ' -*, bugprone-*, cert-*, clang-analyzer-core.*, clang-analyzer-cplusplus.*, clang-analyzer-security.*, cppcoreguidelines-pro-bounds-pointer-arithmetic, cppcoreguidelines-pro-type-cstyle-cast, hicpp-* '
-WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*,hicpp-*'
+WarningsAsErrors: 'bugprone-*,cert-*,clang-analyzer-*'
```
danth: Why was this changed?

danth (changes requested):

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+    EXPECT_TRUE(queue_.getOverflow());
+}
+
+TEST_F(ComponentEventQueueTest, StopUnblocksWaitForEventsOnEmptyQueue)
+{
+    RecordProperty("Description",
+                   "Verify stop() causes a subsequently-called waitForEvents() to return false immediately "
+                   "rather than blocking, matching the shutdown usage in ProcessGroupManager::deinitialize().");
+    queue_.stop();
+    EXPECT_FALSE(queue_.waitForEvents(std::chrono::milliseconds{2000}));
```
danth: This test would still pass if it blocks. Or we are relying on the timeout of the test framework to make it fail

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.hpp
danth: Resolved in ccdddfed44e52744cc47bbb4581178088449c864

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
danth: Resolved in ccdddfed44e52744cc47bbb4581178088449c864

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
danth: Resolved in ccdddfed44e52744cc47bbb4581178088449c864

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
danth: Resolved in ccdddfed44e52744cc47bbb4581178088449c864

score/launch_manager/src/daemon/src/process_group_manager/details/task.hpp
danth: Resolved in 9c769a6e9ae9d9b84133551384d7b8209c213552

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
danth: Resolved in c87ac1cfa5ee4e120c9a932baf019a846c9723bb

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
danth: Resolved in c87ac1cfa5ee4e120c9a932baf019a846c9723bb

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_controller.hpp
danth: Resolved in cd69ca927329fef54863894a4f105af029fd7ecf

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
NicolasFussberger: sounds good to me

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
```diff
@@ -0,0 +1,152 @@
...
+#include "score/mw/launch_manager/process_group_manager/details/component_event_queue.hpp"
+
+#include <gtest/gtest.h>
+#include <chrono>
+
+namespace score::lcm::internal
+{
+
+class ComponentEventQueueTest : public ::testing::Test
+{
```
NicolasFussberger, line 25: I think we need to add here 
        RecordProperty("TestType", "interface-test");
        RecordProperty("DerivationTechnique", "explorative-testing");

score/launch_manager/src/daemon/src/common/constants.hpp
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/370/commits/bcbda43c52c2b6c0ce1eae706152f72d1ce062ec

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent.hpp
WilliamRoebuck: I'll rename it to `ready()`, but I think I disagree with the enum in this one instance. Component state should remain a simple "ready or not ready", and adding an enum makes it tempting to complicate this. Plus, I think the syntax of `if component.ready() ... else` is easier to understand than a switch case or comparison

score/launch_manager/src/daemon/src/common/constants.hpp
WilliamRoebuck: I've reverted the changes in this file now, it wouldn't make sense to change this in this PR

score/launch_manager/src/daemon/src/process_group_manager/details/run_target.hpp
WilliamRoebuck: Updated the docs in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
WilliamRoebuck: Fixed in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
WilliamRoebuck: Removed in acaf3c9b9644dfb1d516bf21fa62921ff2239dea, the test already exists in the queue UT

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
WilliamRoebuck: Done in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent.hpp
WilliamRoebuck: Updated in acaf3c9b9644dfb1d516bf21fa62921ff2239dea, not exactly what you said, let me know what you think

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_controller.hpp
WilliamRoebuck: Done in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_event_receiver.hpp
WilliamRoebuck: Done in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.hpp
WilliamRoebuck: Done in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor_UT.cpp
WilliamRoebuck: Done in dc6aa46c93f66913683318b67c249cbb89fe4ccb

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
WilliamRoebuck: Added in acaf3c9b9644dfb1d516bf21fa62921ff2239dea and 2052521081d89fc9e608110b29babe327fbafe10

score/launch_manager/src/daemon/src/common/constants.hpp
WilliamRoebuck: No longer relevant due to #367

.clang-tidy
WilliamRoebuck: I made all of my interfaces show up as errors 😄 . I've refined it now in acaf3c9b9644dfb1d516bf21fa62921ff2239dea, I added an option that allows classes with default destructors only

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
WilliamRoebuck: I don't think there's any way to test this without fragility, I've changed the test slightly: ebc75c33d55ecbcfb238a4f0acfa49c20e6888f2

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
WilliamRoebuck: Fixed in acaf3c9b9644dfb1d516bf21fa62921ff2239dea

score/launch_manager/src/daemon/src/process_group_manager/details/component_event.hpp
MaciejKaszynski: Shall be done in https://github.com/eclipse-score/lifecycle/issues/375

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue.hpp
```diff
@@ -0,0 +1,104 @@
...
+
+    ComponentEventQueue(const ComponentEventQueue&) = delete;
+    ComponentEventQueue& operator=(const ComponentEventQueue&) = delete;
+    ComponentEventQueue(ComponentEventQueue&&) = delete;
+    ComponentEventQueue& operator=(ComponentEventQueue&&) = delete;
+
+    /// @brief Enqueues an event. If the queue is full, the event is dropped immediately and
+    /// getOverflow() will subsequently return true.
+    /// @returns False if the event is dropped.
+    bool push(ComponentEvent&& event) override
```
MaciejKaszynski: Most of these methods should be `[[nodiscard]]`

score/launch_manager/src/daemon/src/process_group_manager/details/run_target.hpp
```diff
@@ -0,0 +1,76 @@
...
+        active_ = false;
+        return RequestState::kSuccess;
+    }
+
+    RequestResult tryHandleTermination(int32_t) override
+    {
+        return RequestState::kSuccess;
+    }
+
+    uint32_t getIndex() const override
```
MaciejKaszynski, line 59: Need more briefs

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor_UT.cpp
```diff
@@ -0,0 +1,117 @@
...
+using namespace testing;
+using namespace score::mw::lifecycle::internal;
+
+class MockComponentEventQueue : public IComponentEventPublisher
+{
+  public:
+    MOCK_METHOD(bool, push, (ComponentEvent && event), (override));
+    MOCK_METHOD(bool, getOverflow, (), (override, const));
+    MOCK_METHOD(std::size_t, capacity, (), ());
+};
```
MaciejKaszynski: Can we actually separate all mocks into separate mock files.

score/launch_manager/src/daemon/src/process_group_manager/details/mock_component.hpp
```diff
@@ -0,0 +1,34 @@
...
+    MOCK_METHOD(RequestResult, activate, (score::cpp::stop_token stop_token), (override));
+    MOCK_METHOD(RequestResult, deactivate, (score::cpp::stop_token stop_token), (override));
+    MOCK_METHOD(RequestResult, tryHandleTermination, (int32_t status), (override));
+    MOCK_METHOD(uint32_t, getIndex, (), (override, const));
+    MOCK_METHOD(bool, active, (), (override, const));
+};
+
+}  // namespace score::mw::lifecycle::internal
+
+#endif
```
MaciejKaszynski: Missing comment
```suggestion
#endif // MOCK_COMPONENT_HPP_INCLUDED
```

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_event_queue.hpp
```diff
@@ -0,0 +1,53 @@
...
+  public:
+    virtual bool waitForEvents(std::chrono::milliseconds timeout) = 0;
+    virtual std::optional<ComponentEvent> getNextEvent() = 0;
+    virtual void stop() = 0;
+
+    virtual ~IComponentEventConsumer() = default;
+};
+
+/// @brief Interface including publisher and consumer methods
+class IComponentEventQueue : public IComponentEventPublisher, public IComponentEventConsumer
```
MaciejKaszynski: I don't like specifying this to be a queue, would rather just keep it as `PublisherConsumer`

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_controller.hpp
```diff
@@ -0,0 +1,39 @@
...
+{
+
+/// @brief Interface to send requests & notifications about components. The class should handle these notifications by
+/// informing the component, the graph, or event queue.
+class IComponentController
+{
+  public:
+    /// @brief Start work on @p task and push the result to the event queue if the task completes
+    virtual void doWork(ComponentTask&& task) = 0;
+    /// @brief Notify @p component that it has terminated with status @p status. Forward the appropriate event to the
```
MaciejKaszynski, line 31:
```suggestion
    
    /// @brief Notify @p component that it has terminated with status @p status. Forward the appropriate event to the
```

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent.hpp
MaciejKaszynski: ok just rename it

score/launch_manager/src/daemon/src/process_group_manager/details/component_task.hpp
```diff
@@ -0,0 +1,44 @@
...
+{
+
+enum class ComponentTaskType : std::uint_least8_t
+{
+    kActivate = 0U,    /// This task is to start activation of the component
+    kDeactivate = 1U,  /// This task is to deactivate the component
+};
+
+/// @brief Work to perform on a component that can be sent to the job queue to be executed in parallel.
+struct ComponentTask
```
MaciejKaszynski: Mark the struct as [[nodiscard]]

MaciejKaszynski (changes requested):

score/launch_manager/src/daemon/src/common/constants.hpp
danth: This could be done in a separate PR I think

score/launch_manager/src/daemon/src/process_group_manager/details/run_target.hpp
WilliamRoebuck: These are in `IComponent`

MaciejKaszynski (dismissed):

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
FScholPer, line 1: I would prefer to have every test in the tests folder

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_event_publisher_consumer.hpp
```diff
@@ -0,0 +1,53 @@
...
+/// @brief Interface an event consumer can get component events from
+class IComponentEventConsumer
+{
+  public:
+    [[nodiscard]] virtual bool waitForEvents(std::chrono::milliseconds timeout) = 0;
+    [[nodiscard]] virtual std::optional<ComponentEvent> getNextEvent() = 0;
+    virtual void stop() = 0;
+
+    virtual ~IComponentEventConsumer() = default;
+};
```
danth, line 51: These methods are missing descriptions

FScholPer (commented):
Small remark for the tests

danth (commented):
The rest looks good to me :))

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
WilliamRoebuck: Ah I'm following the module template here, if we want to do that we'll have to make a separate PR too, there are some other UTs in the `score/` dir as well

score/launch_manager/src/daemon/src/process_group_manager/details/icomponent_event_publisher_consumer.hpp
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/370/commits/c0740fd168ea8c776cf6dff735e5b4cfab28d930

danth (approved):

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
danth: I agree this would need to be a separate PR as all existing unit tests are currently under `score`

MaciejKaszynski (approved):

score/launch_manager/src/daemon/src/process_group_manager/details/component_event_queue_UT.cpp
NicolasFussberger: Here is the documentation: https://eclipse-score.github.io/score/main/contribute/general/folder.html#module-folder-structure

This defines score/tests to contain only feature integration tests and score/<comp>/tests to contain only component tests. Unit tests are there next to the implementation as we have it right now. This is also followed in other repos like [baselibs](https://github.com/eclipse-score/baselibs/tree/main/score/concurrency) (e.g. clock.cpp and clock_test.cpp).

I think we should then discuss this outside the PR if we want to deviate from this in the future

NicolasFussberger (approved):

PR #374, opened 2026-07-27
Move lifecycle documentation to module

aschemmel-tech:
connected to https://github.com/eclipse-score/score/pull/3123 which deletes all the lifecycle artefacts in score repo moved here


score/launch_manager/docs/requirements/requirements.rst
```diff
@@ -17,42 +17,793 @@ Component Launch Manager Requirements
...
 
-<Headlines (for the list of requirements if structuring is needed)>
-===================================================================
+Launching Processes
+===================
 
-Functional Requirements
------------------------
+.. comp_req:: Forward process information
+    :id: comp_req__launch_man__process_input_output
```
MaciejKaszynski, line 31: @NicolasFussberger is the shortening to `launch_man` fine? Think just `lm` could be used?

MaciejKaszynski (approved):

score/launch_manager/docs/requirements/requirements.rst
NicolasFussberger: I am not sure about the scope of these elements, maybe just "lm" could more easily lead to conflicts with elements in other repos?

I have no opinion about leaving it with launch_man or using the full name launch_manager

NicolasFussberger (approved):

score/launch_manager/docs/requirements/requirements.rst
MaciejKaszynski: I think all requirements have a char limit of 50, so the shorter it is the more characters we can have for the actual requirement :D. I'm fine with any.

PR #377, opened 2026-07-28
First version of new API for Launch Manager

SimonKozik:
First draft of new API for Launch Manager

### DIFF
--- a/score/launch_manager/src/lm_control/src/fixed_string.hpp
+++ b/score/launch_manager/src/lm_control/src/fixed_string.hpp
@@ -55,7 +55,7 @@ struct FixedString
     /// Only `storage_[0]` is set to `'\0'`; the rest of the array is indeterminate.
     /// This is intentional — zeroing `MaxLength + 1` bytes on every default
     /// construction would be wasteful when the value is immediately overwritten.
-    FixedString() noexcept
+    constexpr FixedString() noexcept
     {
         storage_[0] = '\0';
     }
@@ -68,10 +68,12 @@ struct FixedString
     /// available.
     ///
     /// @param s Source string. Does not require to be null-terminated.
-    explicit FixedString(std::string_view s) noexcept
+    constexpr explicit FixedString(std::string_view s) noexcept
     {
         const auto len = std::min(s.size(), MaxLength);
-        std::memcpy(storage_.data(), s.data(), len);
+        for (std::size_t i = 0; i < len; ++i) {
+            storage_[i] = s[i];
+        }
         storage_[len] = '\0';
     }
 
@@ -92,10 +94,12 @@ struct FixedString
     /// @param s Source string. Does not require to be null-terminated.
     ///
     /// @returns Reference to this instance.
-    FixedString& operator=(std::string_view s) noexcept
+    constexpr FixedString& operator=(std::string_view s) noexcept
     {
         const auto len = std::min(s.size(), MaxLength);
-        std::memcpy(storage_.data(), s.data(), len);
+        for (std::size_t i = 0; i < len; ++i) {
+            storage_[i] = s[i];
+        }
         storage_[len] = '\0';
         return *this;
     }
@@ -190,7 +194,9 @@ struct FixedString
     ///         otherwise, false.
     bool operator==(const FixedString& other) const noexcept
     {
-        return as_string_view() == other.as_string_view();
+        // strcmp does a single optimised pass — avoids the two strlen calls
+        // that constructing string_views would require.
+        return std::strcmp(storage_.data(), other.storage_.data()) == 0;
     }
 
     /// @brief Equality comparison with a string_view.
@@ -204,7 +210,13 @@ struct FixedString
     ///         @p other; otherwise, false.
     bool operator==(std::string_view other) const noexcept
     {
-        return as_string_view() == other;
+        // other.size() is already known — use it as a length guard and for
+        // memcmp, avoiding strlen on our storage.
+        if (other.size() > MaxLength) {
+            return false;
+        }
+        return std::strncmp(storage_.data(), other.data(), other.size()) == 0
+               && storage_[other.size()] == '\0';
     }
 
     /// @brief Inequality comparison with another FixedString of the same capacity.
@@ -256,7 +268,7 @@ struct FixedString
 template <std::size_t N, std::size_t M>
 bool operator==(const FixedString<N>& lhs, const FixedString<M>& rhs) noexcept
 {
-    return lhs.as_string_view() == rhs.as_string_view();
+    return std::strcmp(lhs.data(), rhs.data()) == 0;
 }
 
 /// @brief Inequality comparison between FixedString instances of different capacities.
--- a/score/launch_manager/src/lm_control/src/fixed_string.hpp
+++ b/score/launch_manager/src/lm_control/src/fixed_string.hpp
@@ -1,0 +1,304 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#ifndef SCORE_MW_LIFECYCLE_FIXED_STRING_HPP
+#define SCORE_MW_LIFECYCLE_FIXED_STRING_HPP
+
+#include <array>
+#include <cstddef>
+#include <cstring>
+#include <string_view>
+
+namespace score::mw::lifecycle
+{
+
+/// @brief Allocation-free, fixed-capacity string with a null-termination invariant.
+///
+/// Stores a null-terminated string in a plain `std::array` of size `MaxLength + 1`.
+/// Designed for use across API boundaries where heap allocation is undesirable and
+/// a bounded string length is acceptable.
+///
+/// @tparam MaxLength Maximum number of usable characters, excluding the null terminator.
+///                   The internal storage is `MaxLength + 1` bytes so the terminator
+///                   always fits.
+///
+/// @par Invariant
+/// The stored string is always null-terminated. Every constructor and assignment path
+/// enforces this. Characters beyond the null terminator are indeterminate and must
+/// not be read directly — use `as_string_view()` or `data()` instead.
+///
+/// @par Copying and moving
+/// Copy and move are compiler-generated (rule of zero). The `std::array` storage
+/// carries no heap resource, so the generated forms are correct and efficient.
+///
+/// @par Ordering
+/// No `operator<` is provided. `FixedString` values have no meaningful ordering.
+///
+/// @par Cross-capacity comparison
+/// `FixedString<N>` and `FixedString<M>` are distinct types, but free-function
+/// `operator==` / `operator!=` overloads allow comparing them by string content.
+template <std::size_t MaxLength>
+struct FixedString
+{
+    /// @brief Construct an empty string (equivalent to "").
+    ///
+    /// Only `storage_[0]` is set to `'\0'`; the rest of the array is indeterminate.
+    /// This is intentional — zeroing `MaxLength + 1` bytes on every default
+    /// construction would be wasteful when the value is immediately overwritten.
+    FixedString() noexcept
+    {
+        storage_[0] = '\0';
+    }
+
+    /// @brief Construct from a string_view, truncating if necessary.
+    ///
+    /// If `s.size() > MaxLength`, only the first `MaxLength` characters are stored.
+    /// No error is raised at this level — detection and logging of over-length
+    /// strings must happen at the user side where the source and context are
+    /// available.
+    ///
+    /// @param s Source string. Does not require to be null-terminated.
+    explicit FixedString(std::string_view s) noexcept
+    {
+        const auto len = std::min(s.size(), MaxLength);
+        std::memcpy(storage_.data(), s.data(), len);
+        storage_[len] = '\0';
+    }
+
+    /// @brief Assign from a string_view, truncating if necessary.
+    ///
+    /// Behaves identically to the string_view constructor: copies up to
+    /// `MaxLength` characters and always null-terminates. If `s.size() >
+    /// MaxLength`, only the first `MaxLength` characters are stored - no
+    /// error is raised at this level.
+    ///
+    /// Allows direct assignment from string literals and std::string_view
+    /// without constructing a temporary FixedString:
+    /// @code
+    ///   FixedString name;
+    ///   name = "Running";
+    /// @endcode
+    ///
+    /// @param s Source string. Does not require to be null-terminated.
+    ///
+    /// @returns Reference to this instance.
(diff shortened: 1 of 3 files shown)


paulquiring:
It looks like it could work. To see if this if this is the right fit in combination with `score::mw::com` a test would be good before we switch to this.

paulquiring (approved):

score/launch_manager/src/lm_control/src/run_target_activation_source.hpp
```diff
@@ -0,0 +1,40 @@
...
+///       Run Target. This enum describes *why* it resolved,
+///       not whether it succeeded.
+enum class RunTargetActivationSource
+{
+    /// @brief Activation was explicitly requested by a State Manager.
+    kStateManagerRequest = 0,
+
+    /// @brief Activation happened automatically as part of a recovery action,
+    ///        without an explicit State Manager request.
+    kRecoveryAction = 1
```
FScholPer, line 35: I think we should add reason

SimonKozik: I assume this will mean we need to adapt `ActivationCallback` then.
https://github.com/eclipse-score/lifecycle/pull/377/changes#diff-0112cf4be18e494bec2720f425278091273c52be4ffbe12f527cd85178077895R66-R67

Do you have something in mind?

NicolasFussberger:
> It looks like it could work. To see if this if this is the right fit in combination with `score::mw::com` a test would be good before we switch to this.

I like this suggestion. Would it be possible to add a dummy details/lm_controlimpl.hpp with a test that just shows how the API interaction would look like from the users perspective?

score/launch_manager/src/lm_control/src/fixed_string.hpp
```diff
@@ -0,0 +1,304 @@
...
+/// carries no heap resource, so the generated forms are correct and efficient.
+///
+/// @par Ordering
+/// No `operator<` is provided. `FixedString` values have no meaningful ordering.
+///
+/// @par Cross-capacity comparison
+/// `FixedString<N>` and `FixedString<M>` are distinct types, but free-function
+/// `operator==` / `operator!=` overloads allow comparing them by string content.
+template <std::size_t MaxLength>
+struct FixedString
```
MaciejKaszynski, line 51: Would this do most of the work
```cpp
#include <algorithm>
#include <array>
#include <string_view>


template <std::size_t N>
class FixedString : public std::array<char, N>
{
public:
    constexpr FixedString(std::string_view name)
    {
        this->fill('\0');
        const std::size_t copy_size = std::min(name.size(), N - 1);
        std::copy_n(name.data(), copy_size, this->data());
    }
};


template <typename Stream, std::size_t MaxLength>
Stream& operator<<(Stream& os, const FixedString<MaxLength>& fs)
{
    return os << fs.data();
}

```

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+    ///                           queue is cleared, and this activation starts immediately.
+    ///
+    /// @returns void when the Launch Manager accepted the request.
+    ///
+    /// @error kRequestQueueIsFull   Activation was rejected because Launch Manager cannot accept another activation
+    /// request.
+    /// @error kRunTargetDoesntExist Name of the requested Run Target does not exist in current configuration.
+    /// @error kCommunicationError         Connection with Launch Manager cannot be established and request cannot be
+    /// sent.
+    virtual score::Result<void> activate_run_target(std::string_view runTargetName, bool force = false) = 0;
```
MaciejKaszynski: should this not be `RunTargetName`

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+    /// and the call returns kActivationInProgress instead. In that case, the caller
+    /// should wait for the activation completion callback and retry if needed.
+    ///
+    /// @returns the name of the currently active Run Target.
+    ///
+    /// @error kActivationInProgress Launch Manager is currently executing a Run Target
+    ///                              activation and thus there is no single Run Target active.
+    /// @error kCommunicationError         Connection with Launch Manager cannot be established
+    ///                              and information about active Run Target cannot be retrieved.
+    virtual score::Result<RunTargetName> get_active_run_target() = 0;
```
MaciejKaszynski, line 163: Not sure how the the interaction with COM would look like but would it make sense to return a const view of the name, so that it is clear the user shall not edit this.

SimonKozik: I would personally prefer a type that owns the data, on the return side of the function, but I'm open for discussion here.

@NicolasFussberger  do you have an opinion here?

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: @NicolasFussberger do you remember why we went with std::string_view here?

I have feeling that `RunTargetName` maybe better here... As this is what probably will travel through mw::com

score/launch_manager/src/lm_control/src/fixed_string.hpp
SimonKozik: Thanks for the input. Changes discussed in the call are committed now.

score/launch_manager/src/lm_control/src/ilm_control.hpp
NicolasFussberger: It looks to me it may be more difficult to use if we return string_view.
User would need to be really careful about the lifetime of the string_view.
I think we could also not return a view to the data from mw::com as this may be overwritten by new incoming events, but we would need to maintain a copy.
To keep it simple I would be in favour of returning some owning string type

score/launch_manager/src/lm_control/src/ilm_control.hpp
NicolasFussberger: For consistency, RunTargetName is probably better.
Depending on how the data is represented in the mw::com interface it may then require an additional copy though.

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+    /// un-register path; empty callbacks are rejected.
+    ///
+    /// @param[in] callback The callback to invoke when Run Target activation finishes.
+    ///
+    /// @returns void on success.
+    ///
+    /// @error kCallbackInProgress A new callback cannot be registered because the
+    ///                            current callback is still executing. Retry after it returns.
+    /// @error kInvalidArguments   The callback is empty.
+    virtual score::Result<void> register_run_target_activation_callback(ActivationCallback callback) = 0;
```
NicolasFussberger, line 147: AFAIK within the context of the callback, you may not call other mw::methods / send events.
If this is true, it is not easily supported to trigger the next run target activation after the previous one finished (in the scope of the callback). User would need to store the information provided by the callback and then pick this up from a different thread.

@eduard-moskalchuk do you have some insights if this is true?

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+    ///
+    /// @param[in] activationSource   What caused the activation to occur.
+    /// @param[in] activatedRunTarget The Run Target that was activated — may
+    ///                               differ from the one originally requested.
+    using ActivationCallback =
+        std::function<void(RunTargetActivationSource activationSource, RunTargetName activatedRunTarget)>;
+
+    /// @brief Factory method — create a connected ILmControl instance.
+    ///
+    /// Establishes the mw::com connection to the Launch Manager.
```
pawelrutkaq, line 79: I'd say interface shall be silent about internals (in this case mw::com)

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+    /// @error kInvalidArguments instance_specifier is empty or malformed.
+    /// @error kCommunicationError     The Launch Manager service could not be reached.
+    static score::Result<std::unique_ptr<ILmControl>> Create(std::string_view instance_specifier);
+
+    /// @brief Virtual destructor for safe deletion through this interface.
+    virtual ~ILmControl() noexcept = default;
+
+    // Non-copyable and non-movable. Polymorphic types must not be copied or
+    // moved through the interface — it would slice the concrete implementation.
+    // Transfer ownership via std::unique_ptr<ILmControl> instead.
```
pawelrutkaq, line 99: Useless comment, AI ?

score/launch_manager/src/lm_control/src/ilm_control.hpp
```diff
@@ -0,0 +1,165 @@
...
+
+    /// @brief Request Run Target activation.
+    ///
+    /// Posts the request into the Launch Manager's fixed-capacity FIFO queue
+    /// and returns as soon as the request is accepted. The Launch Manager
+    /// executes activations one at a time in FIFO order. Completion is
+    /// notified asynchronously via the callback registered with
+    /// register_run_target_activation_callback().
+    /// If the queue is full, kRequestQueueIsFull is returned immediately
+    /// and the request is discarded.
```
pawelrutkaq, line 113: AI double desc with @errors

score/launch_manager/src/lm_control/src/ilm_control.hpp
pawelrutkaq: I would not expos here any mw::com internals and go with RunTargetName

score/launch_manager/src/lm_control/src/ilm_control.hpp
pawelrutkaq: There was some limitations for sure with calling methods from methods callbacks. But it's our work in LM to address this issue and make the interface functional and not expose some communication channel limitations to the user.

pawelrutkaq (changes requested):
I would fix most of descriptions as they are bloated

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: I'm not sure how we can avoid that...

The instance_specifier param has to be the `mw::com` instance specifier and additionally, there is also requirement that `mw::com` has to be initialized by the user, before this method is called.

Do you have a proposal in mind?

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: This is public API documentation, so I'm not sure there's a strong reason to remove this information.

As far as I'm aware, public API documentation should avoid making assumptions about the reader's level of C++ expertise. Providing this context helps lower the adoption barrier and makes the API more accessible to a broader audience.

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: The mention of `kRequestQueueIsFull` is intentional.

The goal of this description is to highlight what I consider the most important characteristics of this API:
* Requests are queued rather than executed immediately.
* The queue follows FIFO semantics.
* The queue has a fixed capacity and can overflow.

I agree that the selection of what is "most important" is ultimately subjective. However, I tried to focus on aspects that may not be immediately obvious or transparent from the API itself, and kRequestQueueIsFull helps communicate one of those key behaviors.

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: RunTargetName it is then

score/launch_manager/src/lm_control/src/ilm_control.hpp
NicolasFussberger: I would say we just go ahead with this API and if we run into issues we can still change it later if it cannot be solved internally.

score/launch_manager/src/lm_control/src/ilm_control.hpp
pawelrutkaq: @NicolasFussberger why do we expose mw::com instance specifier to the user ? This means we expose the service definition to the user. It feels like you can put any name here - did i miss something that mw_com in no longer internal communication channel ?

score/launch_manager/src/lm_control/src/ilm_control.hpp
pawelrutkaq: Ye but its like documenting how C++ works ;]  Basically what = delete means and also why. Why would a user be interested in information like "Polymorphic types must not be copied or moved through the interface"

score/launch_manager/src/lm_control/src/ilm_control.hpp
NicolasFussberger: As far as I know the instance specifier name passed here has to be defined in the mw::com config file.
So it would be either the user can pass it to the API,
or we would need to use a hardcoded name and documentation tells the user to configure exactly this name in the mw::com config file.

score/launch_manager/src/lm_control/src/ilm_control.hpp
NicolasFussberger: I agree that maybe just "// Non-copyable and non-movable." would be enough here

score/launch_manager/src/lm_control/src/run_target_activation_source.hpp
SimonKozik: As per agreement during the telco, this issue will be solved on a separate PR.

https://github.com/orgs/eclipse-score/discussions/2386#discussioncomment-17494186

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: There is nothing in the documentation that says `activate_run_target` cannot be called from this callback. So AFAIK no change is needed and the rest is implementation detail.

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: Comment changed.

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: I would say we need more time to experiment on how this should be configured.
We can find solution later when we have more experience with `mw::com`

FScholPer (dismissed):

score/launch_manager/src/lm_control/src/ilm_control.hpp
pawelrutkaq: It has to be passed, but it shall be internal inho. This string connects to EXACT declaration of service type and its instance. So we shall provide an "mw_com_config fragment" with opur definition which user shall put into its json file until mw_com will not solve the issue of multiple configs. But exposing this to user sounds like bad idea.

pawelrutkaq: Should this interface then be marked WIP somehow that its probably going to change? or we remove this immediately, really  (i prefer second option)

NicolasFussberger (dismissed):

pawelrutkaq (approved):

score/launch_manager/src/lm_control/src/ilm_control.hpp
SimonKozik: WIP comment added

Issue #368, opened 2026-07-24
Add assertion_handler for all unit tests


MaciejKaszynski:
### What

Currently our assertion handler is registered in the [main](https://github.com/eclipse-score/lifecycle/blob/57c3c4600f3e086734a95482aa200fbf87256451/score/launch_manager/src/daemon/src/main.cpp#L164-L175).

This means that no assertion handler is registered for unit tests if they do not use the `main.cpp`, this makes testing harder as no message can be seen from the assertions.

### Acceptance Criteria (DoD)

- Unit tests have the assertion handler registered.

### How

We should separate the handler into a separate header file, and create a bazel macro for unit tests so that it include the header.
The header should register the handler statically so that no further code is needed.

PR #378, opened 2026-07-28
Add assertion handler to all LCM unit tests via lm_cc_test macro

shegazyy:
Creates a shared testing library (score/launch_manager/src/testing) that registers a diagnostic assertion handler at static-init time via alwayslink = True. Adds a lm_cc_test Bazel macro in tests/utils/bazel/ unit_test.bzl that wraps cc_test and injects the handler dependency automatically. Updates all 11 unit-test BUILD files to use lm_cc_test.


shegazyy:
Hey @MaciejKaszynski 🙏 opened this to fix #368, would appreciate a review whenever you get a chance. Open to any feedback!

score/launch_manager/src/testing/assertion_handler.hpp
```diff
@@ -0,0 +1,40 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#pragma once
```
MaciejKaszynski: everywhere else we use `#ifdefs` so please change this.

score/launch_manager/src/daemon/src/common/assertion_handler.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <score/assert.hpp>
+
+#include <iostream>
+#include <sstream>
+
+namespace score::lcm::testing
+{
+
+inline void registerAssertionHandler() noexcept
```
MaciejKaszynski, line 25: I think we can use the same assertion handler for the daemon as tests.

Would move this maybe into `//score/launch_manager/src/daemon/src/common:assertion_handler` and then also link it for the daemons `cc_binary`.

MaciejKaszynski (changes requested):

MaciejKaszynski (approved):

PR #379, opened 2026-07-28
Adding OSAL inotify

MaciejKaszynski:
Adding a small wrapper around `score::os::InotifyInstance` to get iterator syntax.

Usage:
```cpp
    auto instance = std::make_unique<score::os::InotifyInstanceImpl>();
    instance.AddWatch("/tmp", score::os::Inotify::EventMask::kInCreate);

    for (const score::os::InotifyEvent& event : INotifyRange{std::move(instance)})
    {
        // will block until a new event is present
        // running touch /tmp/test will print test!
        std::cout << event.GetName() << std::endl;
    };
```


score/launch_manager/src/daemon/src/osal/details/posix/inotify_iterator.cpp
```diff
@@ -0,0 +1,191 @@
...
+        ev.data.fd = fd;
+        ::epoll_ctl(epoll_fd_, EPOLL_CTL_ADD, fd, &ev);
+    };
+    add(event_queue_fd_);
+    add(event_fd_);
+}
+
+score::Result<INotifyWatcher> INotifyWatcher::Create() noexcept
+{
+    int event_queue_fd = ::inotify_init1(IN_CLOEXEC);
```
MaciejKaszynski: Move to baselibs if https://github.com/eclipse-score/baselibs/pull/420 is done

score/launch_manager/src/daemon/src/osal/details/posix/inotify_iterator.cpp
```diff
@@ -0,0 +1,191 @@
...
+    int event_fd = ::eventfd(0, EFD_CLOEXEC | EFD_NONBLOCK);
+    int epoll_fd = ::epoll_create1(EPOLL_CLOEXEC);
+
+    if (event_queue_fd < 0 || event_fd < 0 || epoll_fd < 0)
+    {
+        score::Result<INotifyWatcher>{score::MakeUnexpected(score::mw::lifecycle::ExecErrc::kCommunicationError)};
```
SimonKozik: Probably return statement is missing here.

score/launch_manager/src/daemon/src/osal/details/posix/inotify_iterator.cpp
SimonKozik, line 1: This code is in posix folder, but I'm not sure if we don't need separate code for Linux and QNX

score/launch_manager/src/daemon/src/osal/details/posix/inotify_iterator.cpp
MaciejKaszynski: Remade the class so no results anymore.

score/launch_manager/src/daemon/src/osal/details/posix/inotify_iterator.cpp
MaciejKaszynski: It's all using baselibs abstraction so this is posix now.

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#ifndef INOTIFY_RANGE_HPP_
```
paulquiring: OSAL_INOTIFY_RANGE_HPP_INCLUDED

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+
+#include <cstddef>
+#include <iterator>
+#include <memory>
+#include <utility>
+
+#include "score/assert.hpp"
+#include "score/os/utils/inotify/inotify_event.h"
+#include "score/os/utils/inotify/inotify_instance.h"
+
```
paulquiring, line 26: #include "score/static_vector.hpp"

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+namespace score::mw::lifecycle
+{
+
+/// @brief A wrapper around score::os::InotiyInstance to get iterator syntax.
```
paulquiring:
```suggestion
/// @brief A wrapper around score::os::InotifyInstance to get iterator syntax.
```

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+            if (events_.empty())
+            {
+                return false;
+            }
+
+            event_index_ = 0U;
+            return true;
+        }
+
+      public:
```
paulquiring:
```suggestion
```

score/launch_manager/src/daemon/src/osal/BUILD
```diff
@@ -109,6 +109,41 @@ cc_library(
...
+    hdrs = [
+        "inotify_range.hpp",
+    ],
+    include_prefix = "score/mw/launch_manager/osal",
+    strip_include_prefix = "/score/launch_manager/src/daemon/src/osal",
+    visibility = [
+        "//score:__subpackages__",
+        "//tests:__subpackages__",
+    ],
+    deps = [
```
paulquiring, line 123: deps = [
        "@score_baselibs//score/language/futurecpp",
        "@score_baselibs//score/os/utils/inotify:inotify_event",
        "@score_baselibs//score/os/utils/inotify:inotify_instance",
    ]

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+        /// @brief The underlying `InotifyInstance` instance.
+        score::os::InotifyInstance* instance_{nullptr};
+
+        /// @brief Internal buffer of events.
+        score::cpp::static_vector<score::os::InotifyEvent, score::os::InotifyInstance::max_events> events_{};
```
paulquiring, line 146: @NicolasFussberger @SimonKozik @MaciejKaszynski how sure can we be that `score::os::InotifyInstance::max_events` will be sufficient?

paulquiring (approved):
Some findings nothing major.

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
MaciejKaszynski: Can't really control this as this is the buffer that is returned from baselibs, but `max_event`s is just the buffer size. Event's won't be dropped after this they will just be in the file.

By default the value can be found with `cat /proc/sys/fs/inotify/max_queued_events` which is 16384
The user can configure this so don't think we shall worry about this.

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
SimonKozik: Just a suggestion, shall we also include namespace in the #def?
`SCORE_MW_LIFECYCLE_OSAL_INOTIFY_RANGE_HPP`

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include <cstddef>
+#include <iterator>
+#include <memory>
+#include <utility>
+
+#include "score/assert.hpp"
+#include "score/os/utils/inotify/inotify_event.h"
+#include "score/os/utils/inotify/inotify_instance.h"
+
+namespace score::mw::lifecycle
```
SimonKozik, line 27: Shall we add `osal` to the namespace?

`score::mw::lifecycle::osal` or `score::mw::lifecycle::internal::osal`

But this can get long, so I'm not sure here

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
SimonKozik: Looks like this is max events per read only...
```cpp
namespace score
{
namespace os
{

class InotifyInstance
{
  public:
    /**
     * @brief The maximum amount of events a call to Read() can return
     */
    static constexpr const unsigned int max_events{10U};
```
So you will need to read multiple times, but events should not really be dropped .

NicolasFussberger: As we are only interested in deletion or existence of a file I think this should be more than sufficient for our use case.

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
MaciejKaszynski: I think I will resolve this comment and make an issue and bring up in the meeting https://github.com/eclipse-score/lifecycle/issues/410

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+#include "score/static_vector.hpp"
+
+namespace score::mw::lifecycle
+{
+
+/// @brief A wrapper around score::os::InotifyInstance  to get iterator syntax.
+class INotifyRange
+{
+  public:
+    explicit INotifyRange(std::unique_ptr<score::os::InotifyInstance> instance) noexcept
```
NicolasFussberger, line 34: I am not sure I completely understand how this would be used.
It looks like you would still need to interact with the InotifyInstance to setup the file notification first.
And then later to call Close() but the pointer has already been moved to the INotifyRange

E.g. within some kind of FileWatchReadyCondition class:
```c++
std::unique_ptr<InotifyInstance> inotify = std::make_unique<InotifyInstanceImpl>();

// init
bool init() {
  if(!inotify->IsValid) {
    // ready condition failed
     return false;
  }
  
  if(!inotify->AddWatch(path, event_mask)) {
    // ready condition failed
    return false;
  }
}

// at runtime
score::cpp::expected_blank<Error> WaitForReadyCondition(stop_token) {
   INotifyRange range(std::move(inotify);
   while(!ready_timeout_over) {
       for(events : range) {  // does this block? How to interrupt this on timeout?
          // if the expected event occured -> return Success
          // else continue until timeout
       }
   }
}

// at destruction
// want to call inotify->Close() but its already moved out
```

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
```diff
@@ -0,0 +1,168 @@
...
+                instance_ != nullptr, "The parent INotifyRange must have been deleted, documented as UB!");
+
+            const bool events_in_internal_buffer = event_index_ + 1U < events_.size();
+            if (events_in_internal_buffer)
+            {
+                ++event_index_;
+                return true;
+            }
+
+            auto result = instance_->Read();
```
NicolasFussberger, line 125: As far as I understood, this Read() call blocks until there is an event.
How to interrupt this when the ready timeout is over?
Or the startup needs to be aborted?

MaciejKaszynski: `InotifyInstance.Close()` aborts this

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
MaciejKaszynski: I will add this to a FileHandler class that will work like OsHandler (seperate thread).

I will add a `stop()` method I think then this would be the code

```cpp
FileHandler::run(){
    auto instance = std::make_unique<score::os::InotifyInstanceImpl>();
    instance.AddWatch("/tmp", score::os::Inotify::EventMask::kInCreate);
    range_ = INotifyRange{std::move(instance)}
    for (const score::os::InotifyEvent& event : range_)
    {
        std::cout << event.GetName() << std::endl;
    };
}

FileHandler::stop()
{
    range_.stop();
}
```

score/launch_manager/src/daemon/src/osal/inotify_range.hpp
NicolasFussberger: Who would call this close method when ready timeout expires when the worker thread is blocked on the Read call?

SimonKozik: I think a high-level architecture diagram could be beneficial here.

My understanding is that we're moving away from the `wait_for(what, timeout)` model towards a dedicated timeout service that runs continuously in its own thread.

MaciejKaszynski:
So after thinking about it more I think using `inotify` for the wait-for might not too good.

We run into a race condition as `inotify` doesn't allow for setting a watch on a non-existing file.
We could circumvent that by making some mechanism to scan the last valid file, but then the race condition comes if between scanning for the last, multiple directories are made.
A smaller optimization could be done by using `inotify` on the parent dir, if that fails falling back to polling, however not sure if this is worth the extra complexity and need to explain to the user this behavior.

PR #381, opened 2026-07-28
cleanup and update

AlexanderLanin:
Cleanup and update several places to get the repo "up to code". Introduces several best practices like tagged dockerfiles, simplified on-pr workflow, dependabot and score docs config unification.


NicolasFussberger (dismissed):

MaciejKaszynski (dismissed):
Looks good, we have some failures on main which we are sorting out but these changes are not the cause of the pipeline failure.

MaciejKaszynski (approved):

PR #386, opened 2026-07-29
Fixing log message spam

MaciejKaszynski:
Fixes unlimited polling on the control channel for a response to be available causing the spam of info messages:
```
[2026-07-29 11:57:46.662] [INFO] [launch_manager] 2026/07/29 11:57:46.6266501 184502847 000 ECU1 LM LM log debug verbose 1 Failed to send response: response is not empty.
[2026-07-29 11:57:46.662] [INFO] [launch_manager] 2026/07/29 11:57:46.6266501 184502847 000 ECU1 LM LM log debug verbose 1 Control Client handler nudged
[2026-07-29 11:57:46.662] [INFO] [launch_manager] 2026/07/29 11:57:46.6266501 184502847 000 ECU1 LM LM log debug verbose 8 ProcessGroupManager::ControlClientHandler: Sending kSetStateSuccess ( 20 ) re state MainPG/fallback_run_target of PG MainPG
```

Also some small cleanup of `mmap`s

### DIFF
--- a/score/launch_manager/src/daemon/src/control/control_client_channel.cpp
+++ b/score/launch_manager/src/daemon/src/control/control_client_channel.cpp
@@ -20,6 +20,7 @@
 #include "control_client_channel.hpp"
 #include "score/mw/launch_manager/common/constants.hpp"
 #include "score/mw/launch_manager/common/log.hpp"
+#include "score/mw/launch_manager/common/signal_safe_log.hpp"
 
 namespace score
 {
@@ -29,6 +30,28 @@ namespace lcm
 
 namespace internal
 {
+
+bool ControlClientChannel::loadControlNudge()
+{
+    if (nudgeControlClientHandler_ != nullptr)
+    {
+        LM_LOG_ERROR() << "Semaphore was already mapped!";
+        return false;
+    }
+
+    void* nudgeBuf = mmap(
+        NULL, sizeof(osal::Semaphore), PROT_WRITE, MAP_SHARED, osal::IpcCommsSync::control_client_handler_nudge_fd, 0);
+
+    if (nudgeBuf == MAP_FAILED)
+    {
+        LM_LOG_ERROR() << "mmap of nudge semaphore failed in initializeControlClientChannel:"
+                       << std::string_view{std::strerror(errno)};
+        return false;
+    }
+
+    nudgeControlClientHandler_ = static_cast<osal::Semaphore*>(nudgeBuf);
+    return true;
+}
 
 void ControlClientChannel::initialize()
 {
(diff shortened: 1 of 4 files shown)


score/launch_manager/src/daemon/src/process_group_manager/details/process_group_manager.cpp
```diff
@@ -464,10 +462,6 @@ bool ProcessGroupManager::sendResponse(ControlClientMessage msg)
                            << static_cast<int>(msg.request_or_response_) << ") re state"
                            << msg.process_group_state_.pg_state_name_ << "of PG" << msg.process_group_state_.pg_name_;
             ret = scc->sendResponse(msg);
```
danth: Would be nice to clean up this variable which is not necessary since the `if` was removed

```suggestion
            return scc->sendResponse(msg);
```

score/launch_manager/src/daemon/src/control/control_client_channel.cpp
```diff
@@ -196,6 +180,9 @@ ControlClientChannelP ControlClientChannel::initializeControlClientChannel(int f
         lock.unlock();
         init_cv_.notify_all();
     }
+
+    loadControlNudge();
```
danth, line 202: Should we check for errors here?

WilliamRoebuck (approved):
Looks good, I ran the complex rep failure test to make sure we don't lose 100ms due to the lack of polling, but look at these results!
```
Before:
Stats over 10 runs: max = 15.0s, min = 10.1s, avg = 12.9s, dev = 1.7s
After:
Stats over 10 runs: max = 9.2s, min = 7.4s, avg = 8.3s, dev = 0.6s
```

score/launch_manager/src/daemon/src/process_group_manager/details/process_group_manager.cpp
```diff
@@ -30,6 +30,7 @@ static std::atomic_bool em_cancelled{false};
 static void my_signal_handler(int)
 {
     em_cancelled.store(true);
+    ControlClientChannel::nudgeControlClientHandler();
```
NicolasFussberger, line 34: nitpick: nudgeControlClientHandler uses SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE and we do not have a signal-safe assertion handler right now.
It also uses LM_LOG_DEBUG to log a message, which probably also not valid to call from a signal handler.

NicolasFussberger: Could this be called before the control client is setup? E.g. launch manager is started but directly receives a SIGTERM/SIGINT

danth: Once #303 is merged we can use the new signal-safe logging functions from there.

WilliamRoebuck (approved):

MaciejKaszynski:
Closing as the refactor of https://github.com/eclipse-score/lifecycle/pull/392 seems to have fixed it

PR #388, opened 2026-07-29
Replace floating-point seconds with integer milliseconds in timing config

### LINKED ISSUE #345: [launch_manager] Replace floating point seconds with integer milliseconds for timing configuration
### What

Problem: Timing configuration currently uses floating-point seconds. Floating-point values are prone to computational rounding errors and increase the verification burden for safety-critical code paths.

Proposal: Replace all timing-related configuration values expressed as floating-point seconds with integer milliseconds. Use integers (ms) everywhere in configuration files, parsing, internal representation, and public APIs to remove floating-point arithmetic from timing config.

Benefits:

- Eliminates floating-point rounding issues in configuration values.
- Reduces safety verification scope and simplifies reasoning about timing.
- Provides consistent, unambiguous units across the codebase.

### Acceptance Criteria (DoD)

- No floating-point numbers used for timing configuration anywhere in Launch Manager codebase, configs, or public APIs.
- All timing configuration fields documented as integer milliseconds.
- Tests updated and passing.

### How

_No response_

shegazyy:
…nfig

All 8 timing fields in `new_lm_flatcfg.fbs` are changed from `double` (seconds) to `uint32` (milliseconds), eliminating the need for the `secondsToMs()` conversion helper and removing the floating-point representation from the binary format entirely.

Changes:
- `new_lm_flatcfg.fbs`: 8 timing fields changed from `double` to `uint32`
- `flatbuffer_config_loader.hpp`: `kExpectedSchemaVersion` bumped 1 → 2
- `flatbuffer_type_converters.hpp/.cpp`: removed `secondsToMs()` and `kSecondsToMilliseconds`; all 7 call sites simplified to direct assignment
- `flatbuffer_type_converters_UT.cpp`: removed 5 `SecondsToMsTest` cases; updated all FlatBuffer builder calls to pass millisecond integers
- `flatbuffer_config_loader_UT.cpp`: updated all builder calls likewise
- `lifecycle_config.py` (`gen_config`): added `sec_to_ms()` conversion for all 8 timing fields so the generated JSON matches the new uint32 schema
- `launch_manager.schema.json`: updated `schema_version` enum from `[1]` to `[2]`
- All 20 JSON config files: `"schema_version": 1` → `"schema_version": 2`

Closes #345

### DIFF
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -373,7 +373,7 @@
       "type": "integer",
       "description": "Specifies the schema version number that the Launch Manager uses to determine how to parse and validate this configuration file.",
       "enum": [
-        1
+        2
       ]
     },
     "defaults": {
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -373,7 +373,7 @@
       "type": "integer",
       "description": "Specifies the schema version number that the Launch Manager uses to determine how to parse and validate this configuration file.",
       "enum": [
-        2
+        1
       ]
     },
     "defaults": {
--- a/score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
+++ b/score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
@@ -174,7 +174,8 @@ score::cpp::expected<std::optional<RestartAction>, IConfigLoader::Error> convert
     {
         return score::cpp::make_unexpected(number_of_attempts.error());
     }
-    auto delay_before_restart = requireScalarValue(ra->delay_before_restart(), "RestartAction::delay_before_restart");
+    auto delay_before_restart =
+        requireScalarValue(ra->delay_before_restart_ms(), "RestartAction::delay_before_restart_ms");
     if (!delay_before_restart.has_value())
     {
         return score::cpp::make_unexpected(delay_before_restart.error());
@@ -207,7 +208,7 @@ score::cpp::expected<ComponentAliveSupervision, IConfigLoader::Error> convertCom
     if (fb_cas != nullptr)
     {
         auto reporting_cycle =
-            requireScalarValue(fb_cas->reporting_cycle(), "ComponentAliveSupervision::reporting_cycle");
+            requireScalarValue(fb_cas->reporting_cycle_ms(), "ComponentAliveSupervision::reporting_cycle_ms");
         if (!reporting_cycle.has_value())
         {
             return score::cpp::make_unexpected(reporting_cycle.error());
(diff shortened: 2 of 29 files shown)


score/launch_manager/src/daemon/src/configuration/details/new_lm_flatcfg.fbs
```diff
@@ -35,8 +35,8 @@ enum SchedulingPolicy : byte {
 
 // Defines the configuration parameters used for alive monitoring of a component.
 table ComponentAliveSupervision {
-  // Duration in seconds of the time interval used to verify alive notifications.
-  reporting_cycle:double = null; // required
+  // Duration in milliseconds of the time interval used to verify alive notifications.
+  reporting_cycle:uint32 = null; // required
```
NicolasFussberger: could we change the names here also with _ms suffix? Then it also be obvious in the generated code that this is in ms

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -373,7 +373,7 @@
       "type": "integer",
       "description": "Specifies the schema version number that the Launch Manager uses to determine how to parse and validate this configuration file.",
       "enum": [
-        1
+        2
```
NicolasFussberger: We have the difficulty a bit that we are currently in transition.
The new_lm_flatcfg.fbs is the new flatbuffer schema that shall replace the legacy flatbuffer configs.
Though this is currently still only under a feature flag active. The regular build is still using the legacy configs.

My proposal for this PR would be:
* We keep the adaptation for the new config path
* We do not adapt the old config anymore as this is removed soon
* We leave the user-facing json schema unchanged for now, and adapt this once the feature flag is gone and the new config is in use. Then we'll change the user-facing json config from seconds to ms as well.
* This means the translation btw. seconds->ms is done in the lifecycle_config.py for now, as the user-facing json is still configured with seconds (unchanged) and the new flatbuffer schema expects milliseconds (changed).

I think this matches what you already did, only the user-facing json schema change would need to be reverted. As the user-facing json schema is anyway not changed here, I think we don't need to increase the schema version.

What do you think?

NicolasFussberger:
The build with new config feature flag `bazel build //score/... --config=x86_64-linux --//config:use_new_configuration=True` is currently broken (not because of your changes, but because the compiler warnings are now enforced).

Could you apply this simple fix, to get your branch build successfully with new config flag active?

score/launch_manager/src/daemon/src/main.cpp:
```
         auto config_result = config_loader.load(config_path);
-        if (!config_result.has_value()) {
-            LM_LOG_FATAL() << "Failed to load config from: " << config_path;
+        if (!config_result.has_value())
+        {
+            LM_LOG_FATAL() << "Failed to load config from: " << std::string_view(config_path);
             return EXIT_FAILURE;
         }
 #endif
```

NicolasFussberger (changes requested):
Thanks for your contribution! Changes look good to me.

As explained in another comments, I'd only revert the schema version increase, as the json schema is not changed here.

Please take a look at the PR checks, there are some small things to fix to make them pass.

shegazyy:
hey @NicolasFussberger PR checks fixed, ready for review whenever you have the time.

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
NicolasFussberger: Could you also adapt this to `polling_interval_ms`?

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
NicolasFussberger: I think once the polling interval is also adapted this method can be removed entirely.

NicolasFussberger: Could you take these checks also over into the lifecycle_config.py `sec_to_ms` function you introduced?
I guess for any of these cases then the config should already be rejecting in lifecycle_config.py

NicolasFussberger (changes requested):
Thanks for updating the PR.
One minor point about the new polling_interval that was added recently, otherwise looks great to me

shegazyy:
Hello @NicolasFussberger changes are applied please let me know if any other changes are requested

MaciejKaszynski (approved):
looks good

NicolasFussberger (approved):

NicolasFussberger:
thanks a lot @shegazyy. Would you also do the json schema changes as a follow up so that the user actually configured ms values instead of seconds?
This was previously only postponed due to the legacy config support which in the meantime has been removed.

PR #391, opened 2026-07-30
feat(lifecycle_client): add MwLifeCycleManagerMock for OSS consumers

### LINKED ISSUE #265: score_lifecycle_health: Provide MwLifeCycleManagerMock for OSS consumers
### What

[main_function_UT](https://github.com/eclipse-score/config_management/blob/421914926fa92a72681a88862ed341eb1aed92d3/score/config_management/config_daemon/code/main_test.cpp)  fails to compile because it depends on MwLifeCycleManagerMock, which does not exist in the [score_lifecycle_health](https://github.com/eclipse-score/lifecycle/tree/main/score/launch_manager/src/lifecycle_client/src) 

score_lifecycle_health needs to provide *MwLifeCycleManagerMock* — that intercepts score::mw::lifecycle::LifeCycleManager::report_running(), report_shutdown(), constructor and destructor via global GMock callbacks, plus a corresponding Bazel target.

### Acceptance Criteria (DoD)

score::mw::lifecycle::MwLifeCycleManagerMock class is available in score_lifecycle_health

A corresponding Bazel target (e.g. mw_lifecycle_mock) is declared

### How

_No response_

Chahult:
Closes #265

- Add MwLifeCycleManagerMock that intercepts LifeCycleManager::report_running() and LifeCycleManager::report_shutdown() via global GMock callbacks, allowing unit tests to set expectations on both methods without real IPC calls to the Launch Manager.
- Remove the empty report_running() stub from lifecyclemanagermock.cpp; it is now owned by mwlifecyclemanagermock.cpp with a null-safe callback dispatch.
- Move report_shutdown() out of the inline header body and into lifecyclemanager.cpp (empty no-op) to allow the mock translation unit to provide a separate definition.
- Add mw_lifecycle_manager_mock Bazel target and mw_lifecycle_manager_mock_cc public alias in //score/launch_manager.
- lifecycle_mock now depends on mw_lifecycle_manager_mock so existing consumers of lifecycle_mock continue to link without changes.
- Add MwLifecycleMockTest fixture and two tests to lifecycle_mocks_UT.cpp.


score/launch_manager/BUILD
```diff
@@ -100,6 +100,11 @@ alias(
     actual = "//score/launch_manager/src/lifecycle_client:report_running_mock",
 )
 
+alias(
+    name = "mw_lifecycle_manager_mock_cc",
```
NicolasFussberger: Is a separate public target actually needed, or is it enough to depend on existing lifecycle_mock_cc which already has mw_lifecycle_manager_mock as a dependency?

Chahult: lifecycle_mock_cc is sufficient for consumers that use both mocks together. 
The separate alias is only useful for tests that need to intercept report_running/report_shutdown without a full [LifeCycleManagerMock]
e.g. applications that call report_running() directly rather than via Run<>. 
Happy to remove it if you consider that too niche.

NicolasFussberger (approved):

PR #392, opened 2026-07-30
Graph refactor

WilliamRoebuck:
The description below is **not** AI generated. It includes some key information needed to begin a review.

----

## Overview

This is the completion of the refactor started in #370. With these changes, the following is achieved:

- Transitions now activate run targets instead of states and operate on the component level, rather than processes. See `IComponent`. This means that different types of component can now be implemented, such as containers.
- Ready conditions implemented in code.
- `Graph` now processes events sequentially rather than in parallel. See `ComponentEventQueue`.
- `ProcessInfoNode` is no longer responsible for dependency resolution. This responsibility is moved to `Transition`.
- `ProcessInfoNode` & `Graph` coupling resolved - `ProcessInfoNode` no longer has access to `Graph`.

### New components

- `DependencyGraph` - Data structure for storing and traversing component dependencies
- `Transition` - Resolves run target activation logic. Activates and deactivates components as required.

### Behavioral change

Processes are now managed based on component state, rather than process state. This means that once a component is activated, the underlying process shall not be started again until the component becomes inactive (either due to an error or a run target activation). Previously, a self-terminating process would be restarted on every run target activation that included it.

### Minor changes

- `kLaunchManager` comms type removed. This was legacy and no longer needed.
- Configuration now generates an "Off" run target if none is configured
- Several new interfaces have been added to allow for unit testing of existing components

## Review strategy

As this PR is quite large, it may be necessary for reviewers each to tackle subsections of the changes. In terms of line numbers, these sections are roughly equal:

- `graph.cpp` + `graph.hpp` (+ unit tests) 
- `processinfonode.cpp` + `processinfonode.hpp` (+ unit tests)
- `transition.hpp` + `transition_UT.hpp`
- `dependency_graph.hpp` + `dependency_graph_UT.cpp` + `process_group_manager.cpp` + `process_group_manager.hpp`
- Any remaining unreviewed changes


score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+    void traverse(const GraphIndex start, PerNodeFn per_node, FilterFn filter)
+    {
+        visited.assign(visited.size(), false);
+        auto push_res = traversal_queue.push(start);
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(push_res, "Traversal queue was already full");
+        visited[start] = true;
+        while (!traversal_queue.empty())
+        {
+            const auto pop_res = traversal_queue.tryPop();
+            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
```
danth: Why do some checks use `PRD` and some not?

score/launch_manager/src/daemon/src/alive_monitor/details/ifexm/ProcessState.hpp
```diff
@@ -77,7 +77,8 @@ class ProcessState : public saf::common::Observable<ProcessState>
                                     starting = static_cast<uint8_t>(score::lcm::ProcessState::kStarting),
                                     running = static_cast<uint8_t>(score::lcm::ProcessState::kRunning),
                                     sigterm = static_cast<uint8_t>(score::lcm::ProcessState::kTerminating),
-                                    off = static_cast<uint8_t>(score::lcm::ProcessState::kTerminated)};
+                                    off = static_cast<uint8_t>(score::lcm::ProcessState::kTerminated),
+                                    failed = static_cast<uint8_t>(score::lcm::ProcessState::kFailed)};
```
NicolasFussberger, line 81: I see that failed state is currently not reachable from Running state.
However, to be safe maybe we should nonetheless make sure the AliveSupervision treats a failed process state the same as off or sigterm

score/launch_manager/src/daemon/src/alive_monitor/details/supervision/Alive.cpp:
```c++
    if (std::holds_alternative<ProcessStateSnapshot>(f_updateEvent))
    {
        const auto& snapshot = std::get<ProcessStateSnapshot>(f_updateEvent);
        if (snapshot.eProcState == ifexm::ProcessState::EProcState::running)
        {
            return EUpdateEventType::kActivation;
        }
        if (snapshot.eProcState == ifexm::ProcessState::EProcState::sigterm ||
            snapshot.eProcState == ifexm::ProcessState::EProcState::off || 
+           snapshot.eProcState == ifexm::ProcessState::EProcState::failed)
        {
            return EUpdateEventType::kDeactivation;
        }
        return EUpdateEventType::kNoChange;
    }
```

score/launch_manager/src/daemon/src/process_group_manager/details/safe_process_map.hpp
```diff
@@ -27,32 +28,13 @@ namespace lcm
...
-    virtual void terminated(int32_t process_status) = 0;
-
-  protected:
-    ITerminationCallback() = default;
-    ITerminationCallback(const ITerminationCallback&) = default;
-    ITerminationCallback& operator=(const ITerminationCallback&) = default;
-    ITerminationCallback(ITerminationCallback&&) = default;
-    ITerminationCallback& operator=(ITerminationCallback&&) = default;
-};
+using namespace score::mw::lifecycle::internal;
```
NicolasFussberger, line 31: I wonder if we should remove any "using namespace" from header files.
Anything that includes the header transitively pulls every name from score::mw::lifecycle::internal into score::lcm::internal scope, which could result in unexpected naming collisions.

Though this exists in several places: process_info_node.hpp:37, transition.hpp:34, component_of.hpp:25, dependency_graph.hpp:26, graph.hpp:47

It could be done in a separate PR to clean this up.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// @return The number of nodes this graph can hold without reallocating (the @c count
+    /// reserved at construction).
+    std::size_t capacity() const
+    {
+        return nodes.capacity();
+    }
+
+    T& operator[](GraphIndex index)
```
NicolasFussberger, line 98: should we add a const overload?

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// @return The nodes that depend on @p index.
+    const std::vector<GraphIndex>& dependents(GraphIndex index) const
+    {
+        return nodes[index].dependents;
+    }
+
+    /// @brief Traverse the graph, starting at @p start, performing @p per_node on each node and moving to the nodes
+    /// provided by the return value from @p per_node. Only nodes such that @c filter(node) is true are traversed.
```
NicolasFussberger: nitpick: The provided start node is traversed whether the filter for this node is true or not

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+    /// Nodes are visited at most once.
+    template <typename PerNodeFn, typename FilterFn>
+    void traverse(const GraphIndex start, PerNodeFn per_node, FilterFn filter)
```
NicolasFussberger: Maybe the filter could be optional or removed entirely, as it is currently not being used by the code.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+        /// @brief Constructor to allow in-place construction of T.
+        template <typename... Args>
+        GraphNode(Args&&... args) : value(std::forward<Args>(args)...)
+        {
+        }
+    };
+
+  public:
+    /// @param count The exact number of nodes that will be added.
+    DependencyGraph(std::size_t count) : traversal_queue(std::max(count, std::size_t(2)) - 1)
```
NicolasFussberger: Maybe a comment why the queue size is determined like this would be good

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /// @brief Add an edge: @p node depends on @p depends_on.
+    /// During activation, depends_on will be started before node.
+    /// During deactivation, node will be stopped before depends_on.
+    void addDependency(const GraphIndex node, const GraphIndex depends_on)
+    {
+        nodes[node].depends_on.push_back(depends_on);
+        nodes[depends_on].dependents.push_back(node);
+    }
```
NicolasFussberger, line 83: I wonder if an assert makes sense that we are not pushing dependencies beyond the capacity.
At lest for now, I think this is the expectation.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
```diff
@@ -0,0 +1,183 @@
...
+            return *this;
+        }
+
+        bool operator!=(const ValueIterator& other) const
+        {
+            return it != other.it;
+        }
+    };
+
+    /// @returns Iterator at the beginning of the nodes store.
```
danth, line 173: I think it's `@return` not `@returns`, same below

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.hpp
danth, line 1: A few of these methods have docs for parameters which do not exist

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
danth: [It is used](https://github.com/etas-contrib/score_lifecycle/blob/ce227942a057601a9573bc7b6adc678c966f734b/score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp#L128) but making it optional would be nice

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -76,43 +78,120 @@ void Graph::initProcessGroupNodes(IdentifierHash pg_name, uint32_t num_processes
...
+
+        const auto index = nodes_.emplace(
+            std::in_place_type<ProcessInfoNode>,
+            config,
+            process_id,
+            ready_condition,
+            report_state_lambda,
+            pgm_->getProcessInterface(),
+            pgm_->getProcessMap());
+        assert(index == process_id && "Graph indicies must line up with os process indices");
```
danth: Use the `SCORE_LANGUAGE_FUTURECPP_` assertions

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -140,82 +219,108 @@ void Graph::setState(GraphState new_state)
...
     {
-        auto push_res = pgm_->getWorkerJobs()->push(node, kMaxQueueDelay);
+        auto push_res = pgm_->getWorkerJobs()->push(task, kMaxQueueDelay);
         if (push_res)
         {
-            markNodeInFlight();
+            jobs_in_progress_++;
+            // LM_LOG_DEBUG() << "Queued node " << task.component.get().getIndex() << " for "
+            //                << (task.type == ComponentTaskType::kDeactivate ? "deactivation" : "activation")
+            //                << " execution, jobs in progress:" << jobs_in_progress_;
```
danth, line 332: Uncomment or delete this

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -316,145 +398,196 @@ bool Graph::startInitialTransition(ProcessGroupStateID pg_state)
...
+                // never reach here. So this is always a post-ready crash.
+                abort(1U, IComponent::ComponentError::kErrorAfterReady);
+
+                // If there are no jobs in progress, we need to trigger cleanup immediately
+                // Otherwise the graph will stay stuck in kAborting state forever
+                // ???
+                // if (jobs_in_progress_ == 0)
+                // {
+                //     handleNonTransitionExecution(GraphState::kAborting);
+                // }
```
danth: Uncomment or remove

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -316,145 +398,196 @@ bool Graph::startInitialTransition(ProcessGroupStateID pg_state)
...
-            {
-                queueStartJobs();
-            }
-        }
+        // RULECHECKER_comment(1, 3, check_c_style_cast, "This is the definition provided by the OS and does a
+        // C-style cast.", true)
+        LM_LOG_FATAL() << "clock() at failed initial state transition:"
+                       // coverity[cert_err33_c_violation:INTENTIONAL] Does not matter if clock() gives a weird
+                       // value in debug messages.
+                       << (static_cast<double>(clock()) / (static_cast<double>(CLOCKS_PER_SEC) / 1000.0)) << "ms";
```
danth: Why is this marked as fatal?

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -316,145 +398,196 @@ bool Graph::startInitialTransition(ProcessGroupStateID pg_state)
...
+        ControlClientChannel::nudgeControlClientHandler();
     }
 }
 
-inline void Graph::handleNonTransitionExecution(GraphState current_state)
+void Graph::abort(uint32_t code, IComponent::ComponentError reason)
 {
-    if (0 >= --nodes_in_flight_)
+    auto from_state = getState();
+    if (from_state < GraphState::kAborting)
```
danth: Not sure if I like relying on the numeric value of the enum for comparisons. Feel free to argue

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
```diff
@@ -28,131 +27,92 @@ namespace lcm
...
+            desired_state = ProcessState::kRunning;
+            break;
+        case ReadyCondition::kTerminated:
+            desired_state = ProcessState::kTerminated;
+            break;
+        default:
+            break;
     }
-    else
+    assert(desired_state.has_value() && "Ready condition is not implemented");
```
danth: If this must be implemented for every possible ready condition, removing the default case would let the compiler emit a warning if one is missing. Since we treat warnings as errors it would therefore fail to build.

Otherwise: Use `SCORE_LANGUAGE_FUTURECPP_` assertions rather than normal `assert`

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
```diff
@@ -163,381 +123,255 @@ bool ProcessInfoNode::setState(score::lcm::ProcessState new_state)
...
-    if (has_semaphore_.exchange(false))
-    {
-        static_cast<void>(terminator_.post());
-    }
+
+    return res;
 }
 
-void ProcessInfoNode::startProcess()
+IComponent::RequestResult ProcessInfoNode::startProcess(score::cpp::stop_token stop_token)
```
danth, line 191: Same thing with the `SCORE_LANGUAGE_FUTURECPP_` assertions in this method 😄

score/launch_manager/src/daemon/src/process_group_manager/details/transition.hpp
```diff
@@ -0,0 +1,476 @@
...
+/// deactivated (those nodes not reachable from the target state) and those that need to be activated (those reachable
+/// from the target that are not yet active). This list is then used to drive the transition by iterating over the ready
+/// nodes (@ref nextReady()) and marking them as finished once activated/deactivated (@ref onNodeFinished()). The
+/// transition is split into two phases: Stopping phase and Starting phase. Stopping Phase: Every node that is currently
+/// running (@ref stopped() is false) and is not part of the target subgraph is deactivated. The stop set is derived
+/// from live component state across the whole graph rather than from a source subgraph, so it also captures nodes left
+/// running by a previously aborted transition. Once all nodes in the Stopping phase are finished, the transition moves
+/// to the Starting phase. Starting Phase: The nodes that need to be activated are processed next. Once all nodes in the
+/// Starting phase are finished, the transition is complete.
+///
```
danth: Could we break up this text into paragraphs?

score/launch_manager/src/daemon/src/process_group_manager/details/transition.hpp
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    /// @brief Reset the internal state before setting up the next phase
+    void beginSetup(GraphIndex target)
+    {
+        state_.target_root = target;
+        clearNextNodes();
+        state_.pending = 0;
+        state_.phase = Phase::Stopping;
+    }
```
danth: This `beginSetup` method is currently only used for `setupDeactivation`, I think the `_target_root` and `phase` lines could be removed, and done directly by the caller.

Then it could be called within `setupActivation` / `setupDeactivation` so the caller does not need to do additional setup first.

The `std::fill(false)` which is duplicated across both methods could also be moved into `beginSetup`.

danth (changes requested):
Various nitpicks and ramblings. I have seen all the code but did not check the unit tests in much detail yet.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
NicolasFussberger: True its used inside the method, but the code using DependencyGraph is always passing an empty lambda just to satisfy the function signature.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
NicolasFussberger: There are differences when which asserts are active:
> ///  * Default level checks will be compiled unless `SCORE_LANGUAGE_FUTURECPP_ASSERT_LEVEL_PRODUCTION` is defined.
> ///  * Debug level checks will only be compiled if `SCORE_LANGUAGE_FUTURECPP_ASSERT_LEVEL_DEBUG` is defined.
> ///  * Production level checks will always be compiled.

The PRP checks are the production level checks that are always active.

I think we don't have super clear rules when to use what check though.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
NicolasFussberger: Both is valid in doxygen
https://www.doxygen.nl/manual/commands.html#cmdreturn
https://www.doxygen.nl/manual/commands.html#cmdreturns

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
NicolasFussberger: I also dislike it. But it seems a pattern that exists in various places in the codebase.
I would recommend we create a separate ticket for this to fix all places in one go.

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
danth: I was more wondering why a production level assert was chosen for `tryPop`, but not for the queue being full.

If anything I would have done it the other way round. The queue getting full is a real possibility if the input is too large. Whereas the assert after `tryPop` is just testing the behaviour of the queue and should never fail if it is implemented correctly.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
danth: #401

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Ah I misunderstood what these macros do, I'll have another look at the places I've used them

score/launch_manager/src/daemon/src/process_group_manager/details/graph.hpp
```diff
@@ -118,35 +120,48 @@ struct StateTransition {
 /// kUndefinedState -> kSuccess         kUndefinedState
 /// kUndefinedState -> kAborting        kUndefinedState
 // coverity[autosar_cpp14_m3_4_1_violation:INTENTIONAL] The value is used in a global context.
+// clang-format off
 static constexpr GraphState state_results[][static_cast<uint>(GraphState::kUndefinedState) + 1U] = {
-    //from kSuccess                     kInTransition               kAborting                 kCancelled                      kUndefinedState              to new_state
-    {GraphState::kSuccess, GraphState::kSuccess, GraphState::kUndefinedState, GraphState::kUndefinedState,
+    // from kSuccess                     kInTransition               kAborting                 kCancelled
```
MaciejKaszynski: This table shouldn't be formatted.

paulquiring:
@WilliamRoebuck @NicolasFussberger @danth @MaciejKaszynski  
Most of the remaining comments are just improvements which we can collect in issues to continue to work on.  This PR in good shape, lets merge today.

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
```diff
@@ -91,6 +91,10 @@ void ProcessMonitor::doWork(ComponentTask&& task)
 
     if (task.stop_token.stop_requested())
     {
+        if (!event_queue_.push(JobSkipped{task.component.get().getIndex()}))
+        {
+            LM_LOG_ERROR() << "Failed to send activation failed event to event queue!";
```
danth:
```suggestion
            LM_LOG_ERROR() << "Failed to send job skipped event to event queue!";
```

score/launch_manager/src/daemon/src/process_group_manager/details/safe_process_map.hpp
MaciejKaszynski: https://github.com/eclipse-score/lifecycle/issues/411

MaciejKaszynski (dismissed):

score/launch_manager/src/daemon/src/alive_monitor/details/ifexm/ProcessState.hpp
WilliamRoebuck: Done in https://github.com/eclipse-score/lifecycle/pull/392/commits/6cbd7fe77bdc6ecd079eed11968389f560b7dcdf

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Done https://github.com/eclipse-score/lifecycle/pull/392/commits/e855e15b0decef2188b18a80c59990a1fe75303c

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Removed the filter in https://github.com/eclipse-score/lifecycle/pull/392/commits/e855e15b0decef2188b18a80c59990a1fe75303c

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Removed in https://github.com/eclipse-score/lifecycle/pull/392/commits/e855e15b0decef2188b18a80c59990a1fe75303c

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Added in 6cbd7fe77bdc6ecd079eed11968389f560b7dcdf

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Added 5405417744f954884976703bc57ca45739dde66f

score/launch_manager/src/daemon/src/process_group_manager/details/dependency_graph.hpp
WilliamRoebuck: Done in 5405417744f954884976703bc57ca45739dde66f

score/launch_manager/src/daemon/src/process_group_manager/details/transition.hpp
WilliamRoebuck: Done in e855e15b0decef2188b18a80c59990a1fe75303c

score/launch_manager/src/daemon/src/process_group_manager/details/process_monitor.cpp
WilliamRoebuck: Done in e855e15b0decef2188b18a80c59990a1fe75303c

MaciejKaszynski (approved):

NicolasFussberger (approved):

danth (approved):

PR #400, opened 2026-08-03
Doing doc update

MaciejKaszynski:
Doing doc changes to update docs-as-code & score_platform  version

- Added version field to elements that require it.
- Removed feature/architecture section as it has moved to the `score` repo.
- Added version tags for realizes elements.

This PR replaces:
- https://github.com/eclipse-score/lifecycle/pull/399
- https://github.com/eclipse-score/lifecycle/pull/395


docs/features/lifecycle/architecture/index.rst
```diff
@@ -1,143 +0,0 @@
-..
```
NicolasFussberger, line 1: Why is index.rst removed?
I suppose it comes from the module template and that we need to fill this in the future

MaciejKaszynski: This is duplicated with more detail in https://github.com/eclipse-score/score/blob/main/docs/features/lifecycle/architecture/index.rst

Are these two supposed to be different documents?

NicolasFussberger: I think the documentation from score repo will be removed and moved into our repo

MaciejKaszynski: [a4fdd33](https://github.com/eclipse-score/lifecycle/pull/400/commits/a4fdd3377084c7a8ef87d6f472162dc5832cf734)

NicolasFussberger (approved):

PR #402, opened 2026-08-03
Adding a new ready condition `FileState`

MaciejKaszynski:
Adding a new ready condition for file state.

E.g. a component like "setup network" can wait for some `/dev/<network file>` to make the component active. Then Later components can depend on this.

### DIFF
--- a/score/launch_manager/docs/user_guide/configuration.rst
+++ b/score/launch_manager/docs/user_guide/configuration.rst
@@ -208,6 +208,17 @@ component_properties (object)
             * **Allowed Values:**
                 * ``"Running"``: The process has started and reached its running state.
                 * ``"Terminated"``: The process has started, reached its running state, and then terminated successfully.
+        * **file_state** (object, optional)
+            * **Description:** Specifies a ready condition based on the existence state of a file at a given path.
+            * **Properties:**
+                * **file_path** (string, required)
+                    * **Description:** Specifies the absolute path to the file being watched.
+                * **state** (string, optional)
+                    * **Description:** Specifies the required existence state of the file.
+                    * **Allowed Values:**
+                        * ``"Exists"``: The ready condition is ready when the file at ``file_path`` exists.
+                        * ``"Deleted"``: The ready condition is ready when teh file at ``file_path`` is deleted.
+                    * **Default:** ``"Exists"``
 
 .. _lm_conf_deployment_config_object_:
 
--- a/score/launch_manager/docs/user_guide/configuration.rst
+++ b/score/launch_manager/docs/user_guide/configuration.rst
@@ -216,8 +216,8 @@ component_properties (object)
                 * **state** (string, optional)
                     * **Description:** Specifies the required existence state of the file.
                     * **Allowed Values:**
-                        * ``"Exists"``: The ready condition is ready when the file at ``file_path`` exists.
-                        * ``"Deleted"``: The ready condition is ready when teh file at ``file_path`` is deleted.
+                        * ``"Exists"``: The component is ready when the file at ``file_path`` exists.
+                        * ``"Deleted"``: The component is ready when the file at ``file_path`` is deleted.
                     * **Default:** ``"Exists"``
 
 .. _lm_conf_deployment_config_object_:
--- a/score/launch_manager/docs/user_guide/configuration.rst
+++ b/score/launch_manager/docs/user_guide/configuration.rst
@@ -219,6 +219,10 @@ component_properties (object)
                         * ``"Exists"``: The component is ready when the file at ``file_path`` exists.
                         * ``"Deleted"``: The component is ready when the file at ``file_path`` is deleted.
                     * **Default:** ``"Exists"``
+                * **polling_interval** (number, optional)
+                    * **Description:** Specifies the time interval, in seconds (e.g., ``0.5`` for 500 milliseconds), at which the **Launch Manager** checks the file existence state.
+                    * **Constraint:** Must be greater than 0.
+                    * **Default:** ``10ms``
 
 .. _lm_conf_deployment_config_object_:
 
--- a/score/launch_manager/src/daemon/src/configuration/config.hpp
+++ b/score/launch_manager/src/daemon/src/configuration/config.hpp
@@ -37,6 +37,12 @@ enum class ProcessState : uint8_t
     Terminated = 1
 };
 
+enum class FileExistenceState : uint8_t
+{
+    Exists = 0,
+    Deleted,
+};
+
 struct ComponentAliveSupervision
 {
     uint32_t reporting_cycle_ms{};
@@ -52,9 +58,16 @@ struct ApplicationProfile
     std::optional<ComponentAliveSupervision> alive_supervision;
 };
 
+struct FileState
+{
+    std::string file_path;
+    FileExistenceState state{FileExistenceState::Exists};
+};
+
 struct ReadyCondition
 {
-    ProcessState process_state{ProcessState::Running};
+    std::optional<ProcessState> process_state{ProcessState::Running};
+    std::optional<FileState> file_state;
 };
 
 struct ComponentProperties
--- a/score/launch_manager/src/daemon/src/configuration/config.hpp
+++ b/score/launch_manager/src/daemon/src/configuration/config.hpp
@@ -18,6 +18,8 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <sys/types.h>
+#include <variant>
 #include <vector>
 
 namespace score::mw::launch_manager::configuration
@@ -64,11 +66,8 @@ struct FileState
     FileExistenceState state{FileExistenceState::Exists};
 };
 
-struct ReadyCondition
-{
-    std::optional<ProcessState> process_state{ProcessState::Running};
-    std::optional<FileState> file_state;
-};
+
+using ReadyCondition = std::variant<ProcessState, FileState>;
 
 struct ComponentProperties
 {
--- a/score/launch_manager/src/daemon/src/configuration/config.hpp
+++ b/score/launch_manager/src/daemon/src/configuration/config.hpp
@@ -14,11 +14,11 @@
 #define CONFIG_HPP
 
 #include <sys/types.h>
+#include <chrono>
 #include <cstdint>
 #include <optional>
 #include <string>
 #include <string_view>
-#include <sys/types.h>
 #include <variant>
 #include <vector>
 
@@ -64,8 +64,8 @@ struct FileState
 {
     std::string file_path;
     FileExistenceState state{FileExistenceState::Exists};
+    std::chrono::milliseconds polling_interval{10};
 };
-
 
 using ReadyCondition = std::variant<ProcessState, FileState>;
 
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -106,6 +106,11 @@
                   ],
                   "pattern": "^/.*",
                   "description": "Specifies the required existence state of the file. 'Exists': the file must be present at 'file_path'. 'Deleted': the file must be absent from 'file_path'. Defaults to 'Exists' if not specified."
+                },
+                "polling_interval": {
+                  "type": "number",
+                  "exclusiveMinimum": 0,
+                  "description": "Specifies the time interval, in seconds (e.g., '0.5' for 500 milliseconds), at which the Launch Manager checks the file existence state."
                 }
               },
               "required": [
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -104,6 +104,7 @@
                     "Exists",
                     "Deleted"
                   ],
+                  "pattern": "^/.*",
                   "description": "Specifies the required existence state of the file. 'Exists': the file must be present at 'file_path'. 'Deleted': the file must be absent from 'file_path'. Defaults to 'Exists' if not specified."
                 }
               },
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -96,6 +96,7 @@
               "properties": {
                 "file_path": {
                   "type": "string",
+                  "pattern": "^/.*",
                   "description": "Specifies the absolute path to the file being watched."
                 },
                 "state": {
@@ -104,7 +105,6 @@
                     "Exists",
                     "Deleted"
                   ],
-                  "pattern": "^/.*",
                   "description": "Specifies the required existence state of the file. 'Exists': the file must be present at 'file_path'. 'Deleted': the file must be absent from 'file_path'. Defaults to 'Exists' if not specified."
                 },
                 "polling_interval": {
--- a/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
+++ b/score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
@@ -89,6 +89,28 @@
                 "Terminated"
               ],
               "description": "Specifies the required state of the component's POSIX process. 'Running': the process has started and reached its running state. 'Terminated': the process has started, reached its running state, and then terminated successfully."
+            },
+            "file_state": {
+              "type": "object",
+              "description": "Specifies a ready condition based on the existence state of a file at a given path.",
+              "properties": {
+                "file_path": {
+                  "type": "string",
+                  "description": "Specifies the absolute path to the file being watched."
+                },
+                "state": {
+                  "type": "string",
+                  "enum": [
+                    "Exists",
+                    "Deleted"
+                  ],
+                  "description": "Specifies the required existence state of the file. 'Exists': the file must be present at 'file_path'. 'Deleted': the file must be absent from 'file_path'. Defaults to 'Exists' if not specified."
+                }
+              },
+              "required": [
+                "file_path"
+              ],
+              "additionalProperties": false
             }
           },
           "required": [],
--- a/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
+++ b/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cstring>
+#include <iostream>
 #include <map>
 #include <set>
 
@@ -199,19 +200,36 @@ DependencyList ConfigurationAdapter::buildDependencyList(const ComponentProperti
 
     for (const auto& dep_name : props.depends_on)
     {
+        auto dep_it = component_by_name_.find(dep_name);
+        if (dep_it == component_by_name_.end())
+        {
+            // Couldn't find component, continue
+            continue;
+        }
+
+        const auto& dep_props = dep_it->second->component_properties;
+
         Dependency dep{};
         dep.process_state_ = score::lcm::ProcessState::kRunning;
-
-        auto dep_it = component_by_name_.find(dep_name);
-        if (dep_it != component_by_name_.end())
+        if (dep_props.ready_condition.has_value())
         {
-            const auto& dep_props = dep_it->second->component_properties;
-            if (dep_props.ready_condition.has_value())
-            {
-                dep.process_state_ = dep_props.ready_condition->process_state == ProcessState::Running
-                                         ? score::lcm::ProcessState::kRunning
-                                         : score::lcm::ProcessState::kTerminated;
-            }
+            std::visit(
+                [&dep](auto&& arg) {
+                    using argT = std::decay_t<decltype(arg)>;
+
+                    if constexpr (std::is_same_v<argT, score::mw::launch_manager::configuration::ProcessState>)
+                    {
+                        dep.process_state_ = arg == ProcessState::Running ? score::lcm::ProcessState::kRunning
+                                                                          : score::lcm::ProcessState::kTerminated;
+                        return;
+                    }
+                    else if constexpr (std::is_same_v<argT, score::mw::launch_manager::configuration::FileState>)
+                    {
+                        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(false, "FileState is not yet supported");
+                        return;
+                    }
+                },
+                dep_props.ready_condition.value());
         }
 
         dep.target_process_id_ = IdentifierHash{dep_name};
--- a/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
+++ b/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
@@ -222,7 +222,7 @@ DependencyList ConfigurationAdapter::buildDependencyList(const ComponentProperti
                     }
                     else if constexpr (std::is_same_v<argT, score::mw::launch_manager::configuration::FileState>)
                     {
-                        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(false, "FileState is not yet supported");
+                        SCORE_LANGUAGE_FUTURECPP_UNREACHABLE_MESSAGE("FileState is not yet supported");
                         return;
                     }
                 },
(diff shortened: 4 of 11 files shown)


score/launch_manager/docs/user_guide/configuration.rst
```diff
@@ -208,6 +208,17 @@ component_properties (object)
...
+        * **file_state** (object, optional)
+            * **Description:** Specifies a ready condition based on the existence state of a file at a given path.
+            * **Properties:**
+                * **file_path** (string, required)
+                    * **Description:** Specifies the absolute path to the file being watched.
+                * **state** (string, optional)
+                    * **Description:** Specifies the required existence state of the file.
+                    * **Allowed Values:**
+                        * ``"Exists"``: The component is ready when the file at ``file_path`` exists.
+                        * ``"Deleted"``: The component is ready when the file at ``file_path`` is deleted.
```
NicolasFussberger, line 220: Are there any files which cannot be watched via inotify and thus we need to poll for their existence with some to-be-configured interval?

score/launch_manager/docs/user_guide/configuration.rst
```diff
@@ -208,6 +208,17 @@ component_properties (object)
             * **Allowed Values:**
                 * ``"Running"``: The process has started and reached its running state.
                 * ``"Terminated"``: The process has started, reached its running state, and then terminated successfully.
+        * **file_state** (object, optional)
```
NicolasFussberger, line 211: Is it valid to configure both process_state and file_state at the same time?

score/launch_manager/src/daemon/src/configuration/config.hpp
```diff
@@ -52,9 +58,16 @@ struct ApplicationProfile
...
+{
+    std::string file_path;
+    FileExistenceState state{FileExistenceState::Exists};
+};
+
 struct ReadyCondition
 {
-    ProcessState process_state{ProcessState::Running};
+    std::optional<ProcessState> process_state{ProcessState::Running};
+    std::optional<FileState> file_state;
```
NicolasFussberger: I wonder if here or even in the user config we shall introduce some 'type' attribute that defines which type of ready condition is configured/used, or is it intended to support any combinations of these?
```json
"readyCondition" : {
   "type": "PROCESS_STATE",
   ....
}
```

For now with just 2 types its probably still manageable as is.
If we introduce more ready conditions, it might get out of hand.

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
```diff
@@ -53,9 +59,18 @@ table ApplicationProfile {
   alive_supervision:ComponentAliveSupervision; // optional
 }
 
+// Defines a ready condition based on the existence state of a file at a given path.
+table FileState {
+  // Absolute path to the file being watched.
+  file_path:string (required); // required
+  // Existence state of the file. Defaults to Exists if not specified.
+  state:FileExistenceState = Exists; // optional, defaults to Exists
```
NicolasFussberger, line 67: Currently, all the default values are in scripts/config_mapping/lifecycle_config.py

I think this mapping is not existing in the PR

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,28 @@
                 "Terminated"
               ],
               "description": "Specifies the required state of the component's POSIX process. 'Running': the process has started and reached its running state. 'Terminated': the process has started, reached its running state, and then terminated successfully."
+            },
+            "file_state": {
+              "type": "object",
+              "description": "Specifies a ready condition based on the existence state of a file at a given path.",
+              "properties": {
+                "file_path": {
+                  "type": "string",
```
NicolasFussberger, line 98: Is it possible to specify a regex here to validate the basic format of the path?
For example that it is an absolute path that starts with "/".

score/launch_manager/docs/user_guide/configuration.rst
```diff
@@ -208,6 +208,17 @@ component_properties (object)
...
+            * **Description:** Specifies a ready condition based on the existence state of a file at a given path.
+            * **Properties:**
```
NicolasFussberger, line 213: I assume we do not need a dedicated timeout here and we will use the existing ready_timeout parameter.
We need to refactor the usage of ready_timeout so that it actually tracks the time until ready condition is fulfilled and not until Running state is reached

score/launch_manager/docs/user_guide/configuration.rst
MaciejKaszynski: I made it now that only one ready condition can be set.

score/launch_manager/docs/user_guide/configuration.rst
MaciejKaszynski: I think the same parameter shall be used, I think later this we might have the decouple this from the process and make the component care about this.

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
MaciejKaszynski: Added a validation

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
MaciejKaszynski: The Config also defaults this in https://github.com/eclipse-score/lifecycle/blob/main/score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp#L175-L198

score/launch_manager/docs/user_guide/configuration.rst
MaciejKaszynski: As discussed only containers would be problematic here, and I think for now since we do not support containers anyways I don't think we should support this yet. I think in the future we could have an optional polling_rate var that when set would switch to using polling but don't think it makes sense to pay so much of a slow down to support a very specific scenario.

danth: I agree, support for polling only needs to be added if/when we support containers.

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,29 @@
...
+                  "description": "Specifies the absolute path to the file being watched."
+                },
+                "state": {
+                  "type": "string",
+                  "enum": [
+                    "Exists",
+                    "Deleted"
+                  ],
+                  "pattern": "^/.*",
```
danth: Looks like this validation was added to the wrong attribute?

If possible we should also test that the validation fails.

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
```diff
@@ -108,6 +108,18 @@ ProcessState convertProcessState(fb::ProcessState fb_state)
...
 }
 
+FileExistenceState convertFileExistenceState(fb::FileExistenceState fb_state)
+{
+    switch (fb_state)
+    {
+        case fb::FileExistenceState::Deleted:
+            return FileExistenceState::Deleted;
+        case fb::FileExistenceState::Exists:
+        default:
```
danth:
```suggestion
```

score/launch_manager/src/daemon/src/configuration/configuration_adapter.cpp
```diff
@@ -199,19 +200,36 @@ DependencyList ConfigurationAdapter::buildDependencyList(const ComponentProperti
...
+
+                    if constexpr (std::is_same_v<argT, score::mw::launch_manager::configuration::ProcessState>)
+                    {
+                        dep.process_state_ = arg == ProcessState::Running ? score::lcm::ProcessState::kRunning
+                                                                          : score::lcm::ProcessState::kTerminated;
+                        return;
+                    }
+                    else if constexpr (std::is_same_v<argT, score::mw::launch_manager::configuration::FileState>)
+                    {
+                        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(false, "FileState is not yet supported");
```
danth:
```suggestion
                        SCORE_LANGUAGE_FUTURECPP_UNREACHABLE_MESSAGE("FileState is not yet supported");
```

danth (changes requested):

score/launch_manager/docs/user_guide/configuration.rst
```diff
@@ -208,6 +208,21 @@ component_properties (object)
             * **Allowed Values:**
                 * ``"Running"``: The process has started and reached its running state.
                 * ``"Terminated"``: The process has started, reached its running state, and then terminated successfully.
+        * **file_state** (object, optional)
+            * **Description:** Specifies a ready condition based on the existence state of a file at a given path.
```
SimonKozik:
```suggestion
            * **Description:** Specifies a ready condition based on the existence of a file at a given path.
```

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,34 @@
                 "Terminated"
               ],
               "description": "Specifies the required state of the component's POSIX process. 'Running': the process has started and reached its running state. 'Terminated': the process has started, reached its running state, and then terminated successfully."
+            },
+            "file_state": {
+              "type": "object",
+              "description": "Specifies a ready condition based on the existence state of a file at a given path.",
```
SimonKozik:
```suggestion
              "description": "Specifies a ready condition based on the existence of a file at a given path.",
```

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,34 @@
...
+              "properties": {
+                "file_path": {
+                  "type": "string",
+                  "pattern": "^/.*",
```
SimonKozik:
```suggestion
                  "pattern": "^/(?:[^/]+(?:/[^/]+)*)$",
```

This should:
* Enforce that path starts with `/` (absolute Unix path)
  * The same as in original check
* Root path is excluded (we are looking for a file after all).
* There should be no double slashes `//`
* No trailing slash `/tmp/file.txt/`

For validation you can use this schema:
```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "title": "File Paths",
  "type": "object",
  "properties": {
    "paths": {
      "type": "array",
      "items": {
        "type": "string",
        "pattern": "^/(?:[^/]+(?:/[^/]+)*)$"
      }
    }
  },
  "required": ["paths"],
  "additionalProperties": false
}
```

And this data:
```
{
  "paths": [
    "/etc/passwd",
    "/home/test_user/config.yaml",
    "/var/log/syslog",
    "/tmp/file.txt",
    "/",
    "tmp/file",
    "/tmp/",
    "/tmp//file",
    "//tmp/file"
  ]
}
```

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,34 @@
...
+                  "description": "Specifies the absolute path to the file being watched."
+                },
+                "state": {
+                  "type": "string",
+                  "enum": [
+                    "Exists",
+                    "Deleted"
```
SimonKozik:
```suggestion
                    "Not_Exists"
```

I think proving deletion can be difficult and implies two checks... First that file existed and that it was deleted after that. I would propose to go with a simpler check, do not exist at a given point in time

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,34 @@
...
+                  ],
+                  "description": "Specifies the required existence state of the file. 'Exists': the file must be present at 'file_path'. 'Deleted': the file must be absent from 'file_path'. Defaults to 'Exists' if not specified."
+                },
+                "polling_interval": {
+                  "type": "integer",
+                  "exclusiveMinimum": 0,
+                  "description": "Specifies the time interval, in milliseconds, at which the Launch Manager checks the file existence state."
```
SimonKozik: At the moment all time in schema is in seconds...
I know there is plan to change this, but this is the current state. The question is, what we should do with new config?

score/launch_manager/src/daemon/src/configuration/config_schema/launch_manager.schema.json
```diff
@@ -89,6 +89,34 @@
...
+                }
+              },
+              "required": [
+                "file_path"
+              ],
+              "additionalProperties": false
```
SimonKozik, line 119: From earlier comments on this PR, I'm suspecting that we want to prevent situation where `process_state` and `file_state` is configured at the same time...

AFAIK this schema doesn't prevent this, so we may want to look into this

score/launch_manager/src/daemon/src/configuration/config.hpp
MaciejKaszynski: I think this should be a separate ticket, we'd have to change the config in a backwards breaking way so I would move this out of here.

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
NicolasFussberger: true, though this involves some logic at least and is an implementation detail rather than a configuration default.

All the defaults for the raw config values e.g. default uid/gid, transition timeout, default scheduling priority, default scheduling policy, etc. are currently all defined in the mapping script.

I am not really convinced whether this is the best long term approach. But it looks like to me it would be more consistent to the current code if the default polling interval and the default FileExistenceState are set via the defaults in the mapping script.

I think it might also otherwise interfere with the ProcessState ready condition which would be set by default - causing the conflict that both ProcessState and FileState are set.

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
```diff
@@ -53,9 +59,23 @@ table ApplicationProfile {
...
+  // Time in seconds to wait between each poll if the file is present.
+  polling_interval: double = 0.01; //optional, defaults to 0.01s (10ms)
+}
+
 // Defines the conditions that determine when the component enters the ready state.
+// Either process_state or file_state should be set, but not both.
 table ReadyCondition {
-  process_state:ProcessState = null; // required
+  // Required state of the component's POSIX process.
+  process_state:ProcessState = null; // optional
```
NicolasFussberger, line 76:
```suggestion
  process_state:ProcessState; // optional
```

optional non-scalar values are currently all defined without "= null" in this schema because flatbuffer will just return nullptr if not configured which can be used for checking.

Its a bit counter intuitive, but the "= null" is used here for required scalar values so that it can be verified in the code that they have been configured (if they are just marked as required, then flatbuffer will return a default value and its not detectable if they have actually been configured or not).

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.hpp
```diff
@@ -105,8 +109,7 @@ score::cpp::expected<TargetT, IConfigLoader::Error> validateRange(int64_t value,
 [[nodiscard]] score::cpp::expected<ApplicationProfile, IConfigLoader::Error> convertApplicationProfile(
     const fb::ApplicationProfile* fb_ap);
 /// @brief Converts a FlatBuffer ReadyCondition to the config equivalent.
-[[nodiscard]] score::cpp::expected<ReadyCondition, IConfigLoader::Error> convertReadyCondition(
-    const fb::ReadyCondition* fb_rc);
+[[nodiscard]] std::optional<ReadyCondition> convertReadyCondition(const fb::ReadyCondition* fb_rc);
```
NicolasFussberger: I assume convertReadyCondition should always return a ReturnCondition unless there is an error.
All the other functions use score::cpp::expected return value for this pattern. So I wonder if we shall stick to this pattern here

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
```diff
@@ -301,19 +313,58 @@ score::cpp::expected<ApplicationProfile, IConfigLoader::Error> convertApplicatio
...
-        auto process_state = requireScalarValue(fb_rc->process_state(), "ReadyCondition::process_state");
-        if (!process_state.has_value())
+        return std::nullopt;
+    }
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+        fb_fs->file_path(), "FileState::file_path must never be nullptr as it is required in the schema");
+
+    const auto polling_interval_seconds = fb_fs->polling_interval();
+    const auto polling_interval_ms =
+        std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::duration<double>(polling_interval_seconds));
```
NicolasFussberger: There is already `secondsToMs` helper method which has some additional safeguards e.g. to make sure that it does not round down to 0ms

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.hpp
```diff
@@ -69,6 +69,10 @@ score::cpp::expected<TargetT, IConfigLoader::Error> validateRange(int64_t value,
 [[nodiscard]] ApplicationType convertApplicationType(fb::ApplicationType fb_type);
 /// @brief Converts a FlatBuffer ProcessState enum to the config ProcessState.
 [[nodiscard]] ProcessState convertProcessState(fb::ProcessState fb_state);
+/// @brief Converts a FlatBuffer FileState struct to the config equivalent.
+std::optional<FileState> convertFileState(const fb::FileState* fb_fs);
```
NicolasFussberger: All the other methods use score::cpp::expected to return the converted value.

score/launch_manager/src/daemon/src/configuration/configuration_adapter_UT.cpp
```diff
@@ -571,5 +571,117 @@ TEST(ConfigurationAdapterFallbackTest, FallbackRunTargetResolvesDependenciesRecu
     adapter.deinitialize();
 }
 
+TEST(ConfigurationAdapterDependencyTest, DependencyOnNonExistentComponentIsIgnored)
```
NicolasFussberger: Name and description of the test suggest that the dependency is silently ignored, though this case actually crashes the process.

Using the assert is probably a good idea, as silently dropping the dependency would lead to the component being started under unexpected circumstances.

NicolasFussberger (changes requested):

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.hpp
MaciejKaszynski: Well we allow for not configuring a ReadyCondition and default to `ProcessState::Running`, currently this is defaulted in the ConfigAdapter. Should config handle default values? I think the `<config type>loaders` isn't a good place as this is specific to the fbs config.

1. Fbs default - only applies to Fbs configs, if we change to others we would have to remember what is defaulted where.
2. Python script - Not sure if adding more code here also makes sense, think we should reduce this file as much as possible.
3. Code
    1. Config/FbsLoader - only applies to Fbs configs
    2. Config additional wrapper - would need to add wrapper.
    3. Component? Would allow for default values for a specific component type, might be useful.


I think we should define this somehow.

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
MaciejKaszynski: So I think the behavior we want is to be able to either configure process_state or file_state (might want more than one in the future but for now just one).
I think since `ProcessState` is an enum so a scalar value and I do not get a ptr so can't do it this way from what I see.

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.hpp
MaciejKaszynski: I saw the post on https://github.com/eclipse-score/lifecycle/pull/402/changes#r3796455761 so I will also just add to the script

anmittag:
@MaciejKaszynski a) branch is outdated b) some review comments are open

can you please update accordingly?

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
NicolasFussberger: Your right, I got confused since FileState is an object

score/launch_manager/src/daemon/src/configuration/details/lm_flatcfg.fbs
```diff
@@ -53,9 +59,23 @@ table ApplicationProfile {
...
 }
 
+// Defines a ready condition based on the existence state of a file at a given path.
+table FileState {
+  // Absolute path to the file being watched.
+  file_path:string (required); // required
+  // Existence state of the file. Defaults to Exists if not specified.
+  state:FileExistenceState = Exists; // optional, defaults to Exists
+  // Time in seconds to wait between each poll if the file is present.
+  polling_interval: double = 0.01; //optional, defaults to 0.01s (10ms)
```
NicolasFussberger: This default value 0.01 exists in many places:
* Flatbuffer schema
* config.hpp
* mapping script

It would be difficult if we want to change it at some point.
The other config values use this approach:

```
table FileState {
  // Absolute path to the file being watched.
  file_path:string (required); // required
  state:FileExistenceState = null; // required
  polling_interval: double = null; // required
```
The mapping script will always create a json file which has all the values defined (taking the defaults if not configured by the user).
The code can assert that this is in fact the case, since the values would be null if not configured (which would be an implementation error, since the Launch Manager shall always read a fully configured config).
Then the default value is solely defined in the mapping script and could be easily changed in a single location

score/launch_manager/src/daemon/src/configuration/config.hpp
```diff
@@ -52,18 +60,22 @@ struct ApplicationProfile
     std::optional<ComponentAliveSupervision> alive_supervision;
 };
 
-struct ReadyCondition
+struct FileState
 {
-    ProcessState process_state{ProcessState::Running};
+    std::string file_path;
+    FileExistenceState state{FileExistenceState::Exists};
+    std::chrono::milliseconds polling_interval{10};
```
NicolasFussberger: As described here https://github.com/eclipse-score/lifecycle/pull/402/changes#r3813014421 I would rather leave this at 0 and assert during loading that it is always configured in the loaded config (default provided by the mapping script).
Then we do not have the default 10ms in multiple locations

scripts/config_mapping/lifecycle_config.py
```diff
@@ -90,7 +90,16 @@ def report_error(message):
...
 # There are various dictionaries in the config where only a single entry is allowed.
 # We do not want to merge the defaults with the user specified values for these dictionaries.
-not_merging_dicts = ["ready_recovery_action", "recovery_action"]
+not_merging_dicts = ["ready_recovery_action", "recovery_action", "ready_condition"]
+
+
+# Defaults for the optional fields of a "file_state" ready condition.
+# note: these are only default when a file_state is configured so it's no in
+# the main default config.
+file_state_defaults: Dict[str, Any] = {
```
NicolasFussberger, line 99: Could you also add the filestate to one of the integration test configs under scripts/config_mapping/tests and update the expected result json

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_type_converters.cpp
```diff
@@ -301,19 +313,56 @@ score::cpp::expected<ApplicationProfile, IConfigLoader::Error> convertApplicatio
...
+        convertFileExistenceState(fb_fs.state()),
+        std::chrono::milliseconds{*polling_interval_ms}};
+}
+
 score::cpp::expected<ReadyCondition, IConfigLoader::Error> convertReadyCondition(const fb::ReadyCondition* fb_rc)
 {
-    ReadyCondition result{};
-    if (fb_rc != nullptr)
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+        fb_rc != nullptr, "No ReadyCondition is confitured, this should have been defaulted with the script.");
```
NicolasFussberger:
```suggestion
        fb_rc != nullptr, "No ReadyCondition is configured, this should have been defaulted with the script.");
```

NicolasFussberger (approved):

Issue #317, opened 2026-07-15
Create feature integration test specification


TimoSteuerwaldETAS:
### What

Create test specification and test cases for the launching and terminating process with specific properties requirements from the list below: 

**Requirements list:**

Process Launching:

    feat_req__lifecycle__process_ordering
    feat_req__lifecycle__launch_support
    feat_req__lifecycle__parallel_launch_support
    feat_req__lifecycle__custom_cond_support
    feat_req__lifecycle__process_input_output 
    feat_req__lifecycle__process_launch_args
    feat_req__lifecycle__debug_support
    feat_req__lifecycle__support_held_state
    feat_req__lifecycle__uid_gid_support
    feat_req__lifecycle__launch_priority_support
    feat_req__lifecycle__cwd_support
    feat_req__lifecycle__terminal_support
    feat_req__lifecycle__std_handle_redir 
    feat_req__lifecycle__secpol_non_root 
    feat_req__lifecycle__retries_configurable
    feat_req__lifecycle__capability_support
    feat_req__lifecycle__fd_inheritance
    feat_req__lifecycle__support_secpol_type
    feat_req__lifecycle__supplementary_groups
    feat_req__lifecycle__scheduling_policy
    feat_req__lifecycle__runmask_support
    feat_req__lifecycle__aslr_support
    feat_req__lifecycle__process_rlimit_support
    feat_req__lifecycle__detach_parent_process

Terminating Processes

    feat_req__lifecycle__configurable_timeout
    feat_req__lifecycle__process_termination
    feat_req__lifecycle__terminationn_dependency
    feat_req__lifecycle__time_to_wait_config
    feat_req__lifecycle__launch_manager_shutdown
    feat_req__lifecycle__slow_shutdown_support
    feat_req__lifecycle__fast_shutdown_support
    feat_req__lifecycle__launcher_exit_shutdown
    feat_req__lifecycle__shutdown_signal

### Acceptance Criteria (DoD)

- Test specification covers the mentioned requirements are available
- Test case is running
- Failed test cases are tracked in issues

MaciejKaszynski:
@TimoSteuerwaldETAS There has been a lot of tests added by you, is this task complete?

PR #414, opened 2026-08-05
Add test for feat_req__lifecycle__shutdown_signal

TimoSteuerwaldETAS:
Test which verifies that the Launch Manager shuts a process down by sending a SIGTERM and, if the process does not terminate itself in time, escalates to a SIGKILL.


tests/integration/shutdown_signal/BUILD
```diff
@@ -0,0 +1,55 @@
...
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+load("@rules_cc//cc:cc_binary.bzl", "cc_binary")
```
MaciejKaszynski: nit pick, if you do `load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_test")`

tests/integration/shutdown_signal/control_daemon_mock.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include <gtest/gtest.h>
+#include <filesystem>
```
MaciejKaszynski, line 16: You should check building for qnx, I think you need an extra linker flag to get std::filesystem

tests/integration/shutdown_signal/shutdown_signal_process.cpp
```diff
@@ -0,0 +1,79 @@
...
+/// clearly larger than the configured shutdown_timeout so that the process does
+/// not terminate itself in time, forcing the Launch Manager to send SIGKILL.
+constexpr unsigned int kSleepAfterSigtermSeconds = 5U;
+
+/// @brief Creates an empty file using only async-signal-safe calls so it is safe
+/// to invoke from within a signal handler.
+void createFileAsyncSignalSafe(const std::string_view path)
+{
+    const int fd = open(path.data(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+    if (fd >= 0)
```
MaciejKaszynski, line 39: would it make sense to exit_ if file is created to invalidate the test?

MaciejKaszynski (changes requested):

tests/integration/shutdown_signal/shutdown_signal.json
```diff
@@ -0,0 +1,116 @@
+{
+    "schema_version": 1,
+    "defaults": {
+        "deployment_config": {
```
NicolasFussberger, line 4: are all these defaults actually needed?

tests/integration/shutdown_signal/shutdown_signal_process.cpp
```diff
@@ -0,0 +1,79 @@
+/********************************************************************************
```
NicolasFussberger, line 1: Can we give the gtest_process a more meaningful name?

tests/integration/shutdown_signal/shutdown_signal.py
```diff
@@ -0,0 +1,68 @@
...
+    `sigterm_received` file, which is written before the sleep and therefore
+    survives SIGKILL) and is then force-terminated by SIGKILL (proven by the
+    absence of the `graceful_exit` file, which would only exist had the process
+    been allowed to finish sleeping). Both assertions are checked in
+    control_daemon_mock after the "Startup" transition succeeds.
+    """
+
+    # launch manager will simply ignore the arguments if run with --//config:use_new_configuration=False.
+    # the old configuration will be used, which is the default behavior.
+    # The new configuration will be used if run with --//config:use_new_configuration=True
```
NicolasFussberger: I think these comments are a bit too verbose

tests/integration/shutdown_signal/control_daemon_mock.cpp
TimoSteuerwaldETAS: No, you are wrong. This is only needed for QNX 7.

tests/integration/shutdown_signal/shutdown_signal_process.cpp
TimoSteuerwaldETAS: No, it is executed in a signal handler, so error reactions are limited. If the file can not be written, the test will fail in python code.

tests/integration/shutdown_signal/shutdown_signal_process.cpp
TimoSteuerwaldETAS: Fixed.

tests/integration/shutdown_signal/shutdown_signal.py
TimoSteuerwaldETAS: Fixed.

tests/integration/shutdown_signal/shutdown_signal.json
TimoSteuerwaldETAS: Removed a large part of the defaults section.

tests/integration/shutdown_signal/BUILD
TimoSteuerwaldETAS: Fixed. However changed it to `load("@rules_cc//cc:defs.bzl", "cc_binary", "cc_library")`

tests/integration/shutdown_signal/shutdown_signal_process.cpp
danth: The test asserts that `graceful_exit_file` was not created. If opening the `graceful_exit_file` fails, this would be incorrectly interpreted as successful.

> it is executed in a signal handler, so error reactions are limited

`_exit` and `abort` are both async signal safe.

tests/integration/shutdown_signal/shutdown_signal.py
```diff
@@ -0,0 +1,65 @@
...
+        args=["-c", new_config_path],
+        timeout_s=10.0,
+    )
+
+    # Additionally verify on the target filesystem that the SIGTERM was received
+    # and that the process was force-killed (no graceful exit was recorded).
+    res, _ = target.execute(f"test -f {remote_test_dir / 'sigterm_received'}")
+    assert res == 0, "shutdown_signal_process did not receive a SIGTERM during shutdown"
+    res, _ = target.execute(f"test -f {remote_test_dir / 'graceful_exit'}")
+    assert res != 0, "shutdown_signal_process was not force-terminated with SIGKILL"
```
danth: These assertions are also present in `control_daemon_mock.cpp`, we should not duplicate parts of the test.

tests/integration/shutdown_signal/common.hpp
```diff
@@ -0,0 +1,31 @@
...
+/// It is written before the process starts sleeping, so it survives the
+/// subsequent SIGKILL (no code can run after SIGKILL).
+constexpr std::string_view sigterm_received_file = "sigterm_received";
+
+/// @brief Written by gtest_process only if its SIGTERM handler ever returns from
+/// the (long) sleep, i.e. if it was allowed to shut down gracefully. Because the
+/// process intentionally sleeps past its shutdown_timeout, the Launch Manager
+/// must escalate to SIGKILL, which interrupts the sleep. Hence this file must
+/// NOT exist: its absence proves SIGKILL forcibly terminated the process.
+constexpr std::string_view graceful_exit_file = "graceful_exit";
```
danth: Renaming this would make the test much clearer I think.

```suggestion
constexpr std::string_view sigkill_received_file = "sigkill_received";
```

danth (changes requested):

tests/integration/shutdown_signal/common.hpp
TimoSteuerwaldETAS: This would be wrong. The file means, that the process was able to exit gracefully after receiving sigterm, and did _not_ receive sigkill. Process could not even react on sigkill anyways.

tests/integration/shutdown_signal/shutdown_signal.py
TimoSteuerwaldETAS: Good point!

tests/integration/shutdown_signal/common.hpp
danth: Good point. Maybe something like `sigkill_not_received_file` then

tests/integration/shutdown_signal/shutdown_signal_process.cpp
TimoSteuerwaldETAS: Ah, you are right. Thanks!
To make sure to detect such file system errors, I am logging it in the cpp file and asserting the non-existence in python code. Without it, even with abort or exit the test passes, as the process has been executed by launch manager and launch manager is not caring about the actual result here.

tests/integration/shutdown_signal/shutdown_signal_process.cpp
```diff
@@ -0,0 +1,86 @@
...
+{
+    const int fd = open(path.data(), O_WRONLY | O_CREAT | O_TRUNC, 0644);
+    if (fd < 0)
+    {
+        // write()/_exit() are async-signal-safe; std::cerr/std::exit are not.
+        static constexpr char prefix[] = "[FAILED] Failed to create file ";
+        static constexpr char suffix[] = " in signal handler\n";
+        static_cast<void>(write(STDERR_FILENO, prefix, sizeof(prefix) - 1));
+        static_cast<void>(write(STDERR_FILENO, path.data(), path.size()));
+        static_cast<void>(write(STDERR_FILENO, suffix, sizeof(suffix) - 1));
```
MaciejKaszynski: I think if we just exit it would fail the test and in the python you don't need to assert for this message. If we are in a case that open failed it's possible the write could also be flaky?

MaciejKaszynski (changes requested):

tests/integration/shutdown_signal/shutdown_signal_process.cpp
danth: See https://github.com/eclipse-score/lifecycle/pull/414#discussion_r3727367753, it seems like exit is not enough.

However it is indeed possible that the write could also fail.

Perhaps, the python code should check whether the process is still running, rather than trying to have the process itself report whether it was killed.

tests/integration/shutdown_signal/shutdown_signal.py
```diff
@@ -0,0 +1,65 @@
...
+# *******************************************************************************
+from tests.utils.testing_utils.run_until_file_deployed import run_until_file_deployed
+from tests.utils.testing_utils.setup_test import setup_test
+from tests.utils.testing_utils.test_results import assert_test_results
+from attribute_plugin import add_test_properties
+
+
+@add_test_properties(
+    fully_verifies=[
+        "feat_req__lifecycle__shutdown_signal",
```
NicolasFussberger:
```suggestion
        "comp_req__launch_man__shutdown_signal",
```

tests/integration/shutdown_signal/shutdown_signal_process.cpp
TimoSteuerwaldETAS: In Python code it is checked if file system error happened, which is a rare bad case. Everything else is verified in `control_daemon.cpp`. 
If I understand you correctly you suggest to get rid of all the files and let Python check if the process is still there. Still we need to verify that sigterm has been received, so partly we would still stick to writing files. So for sigterm we would use file approach, but for sigkill Python.

This is why I would like to stick to the current solution. I fear searching for child process PID in Python code is a more complex solution and it would move a relevant part of the test to the python code, which would make it special compared to most of the other tests.

tests/integration/shutdown_signal/shutdown_signal.py
TimoSteuerwaldETAS: Fixed.

tests/integration/shutdown_signal/common.hpp
TimoSteuerwaldETAS: Fixed.

tests/integration/shutdown_signal/shutdown_signal_process.cpp
danth: What about, on `SIGTERM`, the process calls `getpid` and writes it to some file. After `SIGKILL` should have been sent, the test reads the file. If the file does not exist, the test fails, indicating that `SIGTERM` was not sent. Otherwise, it has an easy way to find the child process and verify that it is gone.

`getpid`, `open` and `write` are all signal safe. The PID can just be written as raw bytes, no need to encode it to a string.

> it would move a relevant part of the test to the python code, which would make a difference compared to most of the other tests

It could also be done in `control_daemon_mock.cpp` instead of Python

---

Maybe it is too unlikely to actually matter. But I think it would be nice if the test cannot pass as the result of an error.

TimoSteuerwaldETAS: Good idea! Rework done. 👍

danth (approved):
Looks good!

MaciejKaszynski (approved):

PR #428, opened 2026-08-06
Adapt requirements links after feat. req and comp. req split

NicolasFussberger:
We have various tests that link against (formally feature) requirements.

After https://github.com/eclipse-score/score/pull/3123/changes#diff-211e0c5afc85ee5392bfc1c7fb61318a034099410cbc12d767e56f832002758d, many requirements have been converted to component requirements.

This PR only updates the requirements links in the tests to the new elements.


tests/integration/switch_run_target/switch_run_target.py
```diff
@@ -20,7 +20,7 @@
     fully_verifies=[
         "feat_req__lifecycle__request_run_target_start",
         "feat_req__lifecycle__switch_run_targets",
-        "feat_req__lifecycle__process_state_comm",
+        "comp_req__launch_man__process_state_comm",
```
MaciejKaszynski, line 23: I think we need to look back at the linkage for these.

We also "fully verify" this requirement in `tests/integration/process_launch_args/process_launch_args.py`
which doesn't make much sense.

I think the tooling shall verify this but I think it's not currently working made the issue https://github.com/eclipse-score/lifecycle/issues/430

MaciejKaszynski (approved):
The PRs goal is just to replace the stale links and this is working but some more work is needed on requirements on https://github.com/eclipse-score/lifecycle/issues/430

PR #429, opened 2026-08-06
Remove unnecessary inline usage in score/launch_manager

### LINKED ISSUE #421: Remove "inline" from the score/launch_manager code
Many functions are methods are marked as "inline" even though it does not make any sense.
"inline" keyword should be removed in almost all places.

NicolasFussberger:
Closes: https://github.com/eclipse-score/lifecycle/issues/421


paulquiring (approved):
Thx for the cleanup

MaciejKaszynski (approved):

PR #432, opened 2026-08-06
Add AliveMonitorConfig header

paulquiring:
Adds temporary functionality required to copy configuration data until we can, as part of our refactoring efforts, move the configuration data to the entities intended for that purpose.

### DIFF
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
@@ -17,11 +17,7 @@
 
 #include <score/assert.hpp>
 
-namespace score
-{
-namespace lcm
-{
-namespace saf
+namespace score::mw::lifecycle::internal::alive
 {
 
 namespace
@@ -62,6 +58,4 @@ AliveMonitorConfig aliveMonitorConfig(const score::mw::launch_manager::configura
     return result;
 }
 
-}  // namespace saf
-}  // namespace lcm
-}  // namespace score
+}  // namespace score::mw::lifecycle::internal::alive
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
@@ -1,0 +1,67 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "score/mw/launch_manager/alive_monitor/details/common/AliveMonitorConfig.hpp"
+
+#include <utility>
+
+#include <score/assert.hpp>
+
+namespace score
+{
+namespace lcm
+{
+namespace saf
+{
+
+namespace
+{
+
+using ApplicationType = score::mw::launch_manager::configuration::ApplicationType;
+
+bool isSupervisedType(ApplicationType app_type)
+{
+    return app_type == ApplicationType::ReportingAndSupervised || app_type == ApplicationType::StateManager;
+}
+
+}  // namespace
+
+AliveMonitorConfig aliveMonitorConfig(const score::mw::launch_manager::configuration::Config& config)
+{
+    AliveMonitorConfig result{};
+    result.evaluation_cycle_ms = config.aliveSupervision().evaluation_cycle_ms;
+
+    for (const auto& comp : config.components())
+    {
+        if (isSupervisedType(comp.component_properties.application_profile.application_type))
+        {
+            SupervisedComponentConfig info{};
+            info.name = comp.name;
+            info.alive_supervision = comp.component_properties.application_profile.alive_supervision;
+            info.uid = comp.deployment_config.sandbox.uid;
+            result.supervised_components.push_back(std::move(info));
+        }
+    }
+
+    // The evaluation cycle is only meaningful when there is something to supervise; a zero cycle from a config
+    // without an alive-supervision section is benign as long as no supervised components exist.
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+        result.supervised_components.empty() || result.evaluation_cycle_ms > 0,
+        "evaluation_cycle_ms cannot be zero when supervised components are configured");
+
+    return result;
+}
+
+}  // namespace saf
+}  // namespace lcm
+}  // namespace score
--- a/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
+++ b/score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
@@ -22,11 +22,7 @@
 
 #include "score/mw/launch_manager/configuration/config.hpp"
 
-namespace score
-{
-namespace lcm
-{
-namespace saf
+namespace score::mw::lifecycle::internal::alive
 {
 
 /// @file AliveMonitorConfig.hpp
@@ -36,15 +32,20 @@ namespace saf
 /// @brief Supervised component configuration.
 struct SupervisedComponentConfig
 {
+    /// @brief Component short name.
     std::string name;
+    /// @brief Alive-supervision parameters.
     std::optional<score::mw::launch_manager::configuration::ComponentAliveSupervision> alive_supervision;
+    /// @brief Uid the component runs as.
     uid_t uid{};
 };
 
 /// @brief AliveMonitor configuration.
 struct AliveMonitorConfig
 {
+    /// @brief Configuration for every component that is subject to alive supervision.
     std::vector<SupervisedComponentConfig> supervised_components;
+    /// @brief Global supervision evaluation cycle, in milliseconds.
     uint32_t evaluation_cycle_ms{};
 };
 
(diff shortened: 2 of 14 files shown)


score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig_UT.cpp
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+#include "score/mw/launch_manager/alive_monitor/details/common/AliveMonitorConfig.hpp"
+
+#include <gtest/gtest.h>
+
+#include <utility>
+#include <vector>
```
paulquiring, line 19: add `RecordProperty` s

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <sys/types.h>
+
+#include <cstdint>
+#include <optional>
+#include <string>
+#include <vector>
+
+#include "score/mw/launch_manager/configuration/config.hpp"
+
+namespace score
```
MaciejKaszynski: I think we shall use `score::mw::lifecycle::internal::alive` here?

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
```diff
@@ -0,0 +1,59 @@
...
+{
+namespace lcm
+{
+namespace saf
```
MaciejKaszynski: Can we update the clang-tidy with `modernize-concat-nested-namespaces` and concatenate the namespace

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
```diff
@@ -0,0 +1,59 @@
...
+/// @file AliveMonitorConfig.hpp
+/// @brief Adds temporary functionality required to copy configuration data until we can, as part of our refactoring
+/// efforts, move the configuration data to the entities intended for that purpose.
+
+/// @brief Supervised component configuration.
+struct SupervisedComponentConfig
+{
+    std::string name;
+    std::optional<score::mw::launch_manager::configuration::ComponentAliveSupervision> alive_supervision;
+    uid_t uid{};
```
MaciejKaszynski, line 40: I think it would be nice to add briefs even it they are obvious. but would need a brief to why uid is needed

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig_UT.cpp
```diff
@@ -0,0 +1,138 @@
...
+        .build();
+}
+
+class AliveMonitorConfigTest : public ::testing::Test
+{
+  protected:
+    void SetUp() override
+    {
+        RecordProperty("TestType", "interface-test");
+        RecordProperty("DerivationTechnique", "explorative-testing");
```
MaciejKaszynski:
```suggestion
        RecordProperty("DerivationTechnique", "boundary-values");
```

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
MaciejKaszynski, line 1: I think moving this to config component would make more sense. Could be confusing trying to find a config type in alive_moniotr

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/AliveMonitorImpl.cpp
```diff
@@ -32,7 +31,9 @@ AliveMonitorImpl::AliveMonitorImpl(
     SptrIRecoveryClient recovery_client,
     UptrIProcessStateReceiver process_state_receiver,
```
MaciejKaszynski, line 32: Think it should be && so that it is clear we are invalidating the users var

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
```diff
@@ -0,0 +1,67 @@
...
+{
+    AliveMonitorConfig result{};
+    result.evaluation_cycle_ms = config.aliveSupervision().evaluation_cycle_ms;
+
+    for (const auto& comp : config.components())
+    {
+        if (isSupervisedType(comp.component_properties.application_profile.application_type))
+        {
+            SupervisedComponentConfig info{};
+            info.name = comp.name;
```
MaciejKaszynski, line 45: I think the name of the component will always be alive for the lifetime of the alive supervision so we can use std::string_view here and remove a string copy.

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/PhmDaemon.cpp
```diff
@@ -64,7 +64,8 @@ void PhmDaemon::performCyclicTriggers(void)
     }
 }
 
-bool PhmDaemon::construct(const Config& config, const SupervisionBufferConfig& f_bufferConfig_r) noexcept(false)
+bool PhmDaemon::construct(const AliveMonitorConfig& config, const SupervisionBufferConfig& f_bufferConfig_r) noexcept(
```
MaciejKaszynski, line 67: Do you think we could further break down the config?
It seems PhmDaemon only requires the evaluation cycle, so I think we could take as && param and move it here, then somehow move the vector of config into `FlatCfgFactory`.

MaciejKaszynski (changes requested):

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
paulquiring: The code base is inconsistent on this one.

paulquiring: See issue https://github.com/eclipse-score/lifecycle/issues/435. Lets do this for the code base and  not only new code.

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
paulquiring: See Description
"Adds temporary functionality required to copy configuration data until we can, as part of our refactoring efforts, move the configuration data to the entities intended for that purpose." 
The copy is intentional. We need to copy because the object referenced will be moved in a later PR to the ProcessInfoNode.

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/PhmDaemon.cpp
paulquiring: There will be quite a few upcoming PRs changing the structure of the configuration consumers:

The `take*` methods are going to be used. This is why more splitting up does not make sense. I made the split so that we do not have to copy the entire config object.

https://github.com/eclipse-score/lifecycle/blob/main/score/launch_manager/src/daemon/src/configuration/config.hpp

score/launch_manager/src/daemon/src/alive_monitor/details/daemon/AliveMonitorImpl.cpp
paulquiring: Invalidation will happen later, we need the copy for now.

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
paulquiring: I created https://github.com/eclipse-score/lifecycle/issues/435 so that we do this for the code base.

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
paulquiring: Its temporary because of that it should be visible/used only within alive_monitor. 

```c++
/// @brief Adds temporary functionality required to copy configuration data until we can, as part of our refactoring
/// efforts, move the configuration data to the entities intended for that purpose. 
```

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.cpp
MaciejKaszynski: I created an issue as this is a bigger change but in general the alive monitor shouldn't need the name of the component. https://github.com/eclipse-score/lifecycle/issues/436 Then this would become a IdentifierHash and be just a std::uint64_t copy instead of a string.

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
paulquiring: The whole alive_monitor module currently lives under `score::lcm::saf` (timers, factory, daemon, etc.), so this file is consistent with its neighbours. Renaming to `score::mw::lifecycle::internal::alive` would make it inconsistent with the rest of the module. I would prefer to keep `score::lcm::saf` here and do any namespace renaming as a separate, module-wide change rather than in this PR.

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig.hpp
paulquiring: done

score/launch_manager/src/daemon/src/alive_monitor/details/common/AliveMonitorConfig_UT.cpp
paulquiring: fixed

MaciejKaszynski (approved):

PR #441, opened 2026-08-10
Extend crash_on_startup test to also cover comp_req__launch_man__retries_configurable

TimoSteuerwaldETAS:
Extended test to verify if a component fails to start, the amount of retries is configurable.


tests/integration/crash_on_startup/control_client_mock.cpp
```diff
@@ -43,6 +43,23 @@ TEST(CrashOnStartup, ControlClientMock)
         EXPECT_FALSE(std::filesystem::exists(fallback_file)) << "Fallback run target should not be activated yet";
     }
 
+    EXPECT_TRUE(std::filesystem::remove(crash_count_file)) << "Count file must be removed successfully, before reused in the next run target";
+
+    // Given a process that crashes on startup five times, but is configured to retry five times - so still succeeds
```
WilliamRoebuck: Do we need to start it 5 times? That might increase the chance of a timeout on a slow runner and the test would be the same with 3

tests/utils/test_helper/process_crashing_on_startup_n_times.cpp
WilliamRoebuck, line 1: This should be in `utils/` I think, it could be useful for other tests

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include <iostream>
+#include <string_view>
+
+#include "tests/utils/test_helper/test_helper.hpp"
+#include <score/mw/lifecycle/report_running.h>
+
+/// @brief Default number of times the process crashes before starting up successfully.
+constexpr int kDefaultCrashesUntilSuccess = 3;
+
+TEST(CrashOnStartup, ProcessCrashingOnStartupTwice)
```
WilliamRoebuck: The test should be renamed

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
```diff
@@ -0,0 +1,69 @@
...
+/// @brief Persists how many times the process has crashed so far to the crash count file.
+void write_crash_count(const std::string_view file_path, const int count)
+{
+    std::ofstream file{std::string{file_path}};
+    file << count;
+}
+
+int main(int argc, char** argv)
+{
+    // The number of crashes before a successful startup is taken from the command line, defaulting to two.
```
WilliamRoebuck:
```suggestion
    // The number of crashes before a successful startup is taken from the command line, defaulting to kDefaultCrashesUntilSuccess.
```

WilliamRoebuck (changes requested):
Looks good, just a few small comments

tests/integration/crash_on_startup/crash_on_startup.json
```diff
@@ -124,6 +148,12 @@
                 "process_crashing_on_startup_twice"
```
SimonKozik, line 118: Looks like this Component and this Run Target is a dead configuration now.

We should either add a test that uses it or remove this config

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
SimonKozik: If we make this a utility then I suggest, this program should also support `-h` and `--help` params to make it easier to understand.

tests/integration/crash_on_startup/crash_on_startup.json
TimoSteuerwaldETAS: It is still used. See control_client_mock.cpp.

tests/integration/crash_on_startup/control_client_mock.cpp
TimoSteuerwaldETAS: Good point — reduced to avoid the slow-runner timeout risk. The success cases now crash once (retries=2) and three times (retries=3), so nothing is started 5 times anymore.

tests/utils/test_helper/process_crashing_on_startup_n_times.cpp
TimoSteuerwaldETAS: Agreed. Moved it to `tests/utils/test_helper/` with `//tests:__subpackages__` visibility so other integration tests can reuse it.

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
TimoSteuerwaldETAS: Renamed the test to `ProcessCrashingOnStartupNTimes` to match the parametrized behaviour.

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
TimoSteuerwaldETAS: Applied, thanks.

tests/integration/crash_on_startup/crash_on_startup.json
TimoSteuerwaldETAS: `run_target_crash_on_startup_twice` was actually used in the test, so it wasn't dead. That said, the scenarios have been restructured/renamed: the success cases are now `once` (retries=2) and `three_times` (retries=3), and `no_retries` (number_of_attempts=0) replaces the old `process_crashing_on_startup_always` binary to exercise the fallback.

tests/integration/crash_on_startup/process_crashing_on_startup_n_times.cpp
TimoSteuerwaldETAS: Added `-h`/`--help` output describing the optional crash-count argument, now that the binary lives under utils.

WilliamRoebuck (approved):
Thanks, changes look good

WilliamRoebuck (approved):

SimonKozik (approved):
Great work! Thx for working on this.

MaciejKaszynski (approved):

PR #442, opened 2026-08-10
Extend switch_run_target test to also cover feat_req__lifecycle__process_ordering and comp_req__launch_man__launch_manager_shutdown

TimoSteuerwaldETAS:
Extends text to verify correct shutdown order.


tests/integration/switch_run_target/control_client_mock.cpp
```diff
@@ -17,6 +17,23 @@
 #include <score/mw/lifecycle/control_client.h>
 #include <score/mw/lifecycle/report_running.h>
 
+namespace
+{
+void resetFilesystem(
```
WilliamRoebuck: `check_clean()` with `strict=false` should have the same effect as this

tests/integration/switch_run_target/control_client_mock.cpp
```diff
@@ -65,9 +84,29 @@ TEST(SwitchRunTarget, ControlClientMock)
...
+        auto result = client.ActivateRunTarget("run_target_a").Get(stop_token);
+        EXPECT_TRUE(result.has_value()) << "Activating target run_target_a failed: " << result.error().Message();
+    }
+    TEST_STEP("Verify running processes")
+    {
+        for (const auto proc : running_processes)
+        {
+            EXPECT_TRUE(std::filesystem::exists(proc)) << "A process depended on by run target A was not started!";
+        }
+    }
```
WilliamRoebuck: Is this a repeat of the code above? The comment says this verifies shutdown order, but this is already verified by the processes themselves. I'm not sure this is needed

tests/integration/switch_run_target/switch_run_target.py
```diff
@@ -23,16 +23,18 @@
         "comp_req__launch_man__process_state_comm",
         "feat_req__lifecycle__process_termination",
         "feat_req__lifecycle__terminationn_dependency",
+        "feat_req__lifecycle__process_ordering",
+        "comp_req__launch_man__launch_manager_shutdown",
```
WilliamRoebuck, line 27: I read this requirement as "When LM is shutdown, processes terminate in the correct order", so the test would need to wait until the control client receives SIGTERM from the automatic off state transition. This is just my opinion though, I could be misunderstanding

WilliamRoebuck (changes requested):
I think the test already satisfied the feature requirement and, if you agree with me on the component one, we'd just need to change the control client to wait for SIGTERM rather than requesting the "off" target

tests/integration/switch_run_target/control_client_mock.cpp
TimoSteuerwaldETAS: Fixed.

tests/integration/switch_run_target/control_client_mock.cpp
TimoSteuerwaldETAS: You are completely correct.

tests/integration/switch_run_target/switch_run_target.py
TimoSteuerwaldETAS: The document is phrased like this:

> The :term:`Launch Manager` shall support normal shutdown by terminating all
    process in the dependency order.

To my understanding it is rather about the order of processes. So, when shutting down processes the dependency order needs to be respected. Not only in launch manager shut down, but also there. As in launch manager shut down Off Run Target is activated anyway, I think this test already tests the requirement fully. The order is verified in the component cpp files.

WilliamRoebuck: That makes sense, but there's also `comp_req__launch_man__launcher_exit_shutdown` which says
>  The :term:`Launch Manager` shall exit after performing shutdown operation by
    stopping all the :term:`Processes <Process>` it owns in the dependency order when requested.

Do you think we could add this requirement as well and wait for the shutdown? Otherwise we'd need another very similar test to verify that one

TimoSteuerwaldETAS: There are even two separate test for this requirement on my internal branch in etas-eng. 😃 
So I would like to keep that one separated for now, as the `switch_run_target` test has the most complex component and run target configuration, which is not needed for verifying `comp_req__launch_man__launcher_exit_shutdown `

tests/integration/switch_run_target/switch_run_target.py
```diff
@@ -23,16 +23,18 @@
...
     derivation_technique="requirements-analysis",
 )
 def test_switch_run_target(target, setup_test, assert_test_results, remote_test_dir):
     """
-    Objective: Verifies that the launch manager respects component and run target dependencies when switching run targets, enforcing correct startup and termination order.
+    Objective: Verifies that the launch manager respects component and run target dependencies when switching run targets and shuting down, enforcing correct startup and termination order.
 
-    The control client activates run_target_a, which depends on run_target_c (containing component_d) and component_a (which depends on component_b). After activation it switches back to Startup and then Off.
-    Expected Behaviour: Component B starts before component A, component D is started, component A terminates before component B, and component E (not in the dependency chain) is never launched.
+    The control client activates run_target_a, which depends on run_target_c (containing component_d) and component_a (which depends on component_b). After activation it switches back to Startup, then again to run_target_a, and then Off.
```
WilliamRoebuck:
```suggestion
    The control client activates run_target_a, which depends on run_target_c (containing component_d) and component_a (which depends on component_b). After activation it switches back to Startup, and then Off.
```

WilliamRoebuck (commented):
Looks good, just one minor fix

tests/integration/switch_run_target/switch_run_target.py
WilliamRoebuck: Sounds good, if we can split it up into simpler test cases that's better anyway

WilliamRoebuck (approved):

MaciejKaszynski (approved):

Issue #150, opened 2026-04-02
Remove config schema description and doc description


MaciejKaszynski:
Right now the schema file has descriptions of each field, but this is also duplicated in the docs in https://github.com/eclipse-score/lifecycle/blob/65893bc2807e60f9a333f83354ce191282cd1372/src/launch_manager_daemon/config/config_schema/docs/launch_manager_configuration.rst

The docs should either point to the schema file or the docs should be generated from the schema but its bad to have duplication.

MaciejKaszynski:
I have found an extension that would generate the docs and created an issue in the central docs repo to add. https://github.com/eclipse-score/docs-as-code/issues/472

PR #443, opened 2026-08-10
docs: split towards multiple docs_bundles

AlexanderLanin:
This pull request introduces significant improvements to the documentation build system by modularizing documentation bundles for components and subcomponents, updating Sphinx configuration, and improving navigation structure. The changes make it easier to manage, build, and browse documentation for the `launch_manager` and `health_monitor` components and their subcomponents.

**Documentation Build System Improvements:**

* Introduced `docs_bundle` Bazel targets for both `score/launch_manager` and `score/health_monitor`, enabling modular, component-specific documentation builds. Subcomponent documentation is now bundled and mountable at specified paths for better organization.
* Updated the main `BUILD` file to aggregate component documentation using the new bundle system and added project metadata for improved documentation context.
* Refactored documentation indices to use new `components/index.rst` and `subcomponents/index.rst` files, allowing dynamic population of navigation trees via the `docs_bundle` system. This replaces hardcoded paths with a more maintainable structure.
* Removed the custom `conf.py` Sphinx configuration in favor of Bazel-driven configuration, simplifying the build process and reducing maintenance overhead.
* Updated filegroup visibility for configuration schema files to be public, ensuring full documentation builds have access to necessary data.

These changes collectively modernize and modularize the documentation workflow, making it more scalable and maintainable as the project grows.


conf.py
MaciejKaszynski, line 1: Is the removal of the `conf.py` needed for moving towards the bundle method, or is this a separate change?

I am asking as we were at some point planning to automate creating the config description into docs. See https://github.com/eclipse-score/lifecycle/issues/150. From what I understand if we create some custom script to generate this text we'd need to declare it in the `conf.py` so going with bundles would prevent this. Or is there a better approach?

MaciejKaszynski (commented):
Please could you also update the description of the PR to be more descriptive of the changes.

conf.py
AlexanderLanin: Custom additions to conf.py were never really possible. It just looked like it works. You can run a custom script to generate something which is included in docs, or you can raise a feature request in docs to support something specific. Additions in the local conf.py would not work for the final documentation, so they are rather dangerous as the look fine locally.

AlexanderLanin: Looking at #150 the easiest way might be to generate an rst or html file before starting docs-as-code.

MaciejKaszynski (approved):

docs/index.rst
```diff
@@ -21,21 +21,20 @@ Module / Feature Documentation
...
 
 Component documentation
 ------------------------
 
 .. toctree::
    :maxdepth: 1
 
-   score/launch_manager/docs/index
-   score/health_monitor/docs/index
+   components/index
```
a-zw, line 37: This changes URLs but I guess that is ok for lifecycle as it is unlikely there are many links floating around.

PR #445, opened 2026-08-10
Faster `pre-commit`

danth:
- Avoid running Bazel once per file when those files had already been formatted by the previous hook
- Use a faster implementation of the copyright fixer

Should be merged after #447


.pre-commit-config.yaml
MaciejKaszynski: add cpp files now that the format supports it

MaciejKaszynski (changes requested):

MaciejKaszynski (approved):

PR #447, opened 2026-08-10
Add missing copyright headers

danth:
Note: this was automated, I have not checked that the dates are correct.

### DIFF
--- a/.vscode/rustfmt.sh
+++ b/.vscode/rustfmt.sh
@@ -1,3 +1,16 @@
 #!/usr/bin/env bash
 
+# *******************************************************************************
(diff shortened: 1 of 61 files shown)


.github/PULL_REQUEST_TEMPLATE/bug_fix.md
MaciejKaszynski, line 1: I think we shouldn't add these, I think in the pre-commit we can exclude directories and the .github directory should probably be excluded for copyrights

MaciejKaszynski (changes requested):
Just unsure of the gh issue template.

MaciejKaszynski (approved):

MaciejKaszynski (approved):

PR #450, opened 2026-08-11
Add tests which are verifying comp_req__lifecycle__launcher_exit_shutdown

TimoSteuerwaldETAS:


score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -542,6 +544,27 @@ void Graph::forceKillProcesses()
...
+    std::chrono::milliseconds max_timeout{0};
+    for (const auto& component : nodes_)
+    {
+        if (const ProcessInfoNode* process = std::get_if<ProcessInfoNode>(&component))
+        {
+            // Only processes with a live OS process still to stop actually consume the
+            // SIGTERM+timeout wait during the transition to Off. Already terminated/failed
+            // processes short-circuit deactivation, and never-started nodes (pid == 0) have
+            // nothing to wait for, so neither should inflate the bound. This excludes
+            // one-shot / ready-on-termination processes that have already exited.
```
WilliamRoebuck: This comment should be removed, it makes the code seem more complex than it is

score/launch_manager/src/daemon/src/process_group_manager/details/graph.hpp
```diff
@@ -288,6 +288,13 @@ class Graph final
     /// @brief For forced shutdown, kill all leftover processes
     void forceKillProcesses();
 
+    /// @brief Returns the largest configured shutdown_timeout across the processes
+    /// @details The largest configured shutdown_timeout (termination_timeout_ms_) across the process
+    /// nodes of this process group that still have a live process to stop (pid > 0 and not yet
+    /// terminated), or zero if there are none.
+    /// @return The timeout in milliseconds, or zero if there are no live processes to stop.
```
WilliamRoebuck:
```suggestion
    /// @brief Returns the largest configured shutdown_timeout across all running processes
```

WilliamRoebuck: I'm not sure if it shows up my suggested change but I think the @details section should be removed

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.hpp
```diff
@@ -99,6 +100,10 @@ class ProcessInfoNode final : public IComponent
     /// @return The current state of this process.
     score::mw::lifecycle::ProcessState getState() const;
 
+    /// @return The configured shutdown_timeout (termination_timeout_ms_) for this process, or zero
```
WilliamRoebuck:
```suggestion
    /// @return The configured shutdown_timeout for this process, or zero
```

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -153,10 +153,17 @@ void ProcessGroupManager::deinitialize()
...
+    // Stop and join the worker threads BEFORE destroying the process groups.
+    // Worker threads run ProcessInfoNode::doWork(), which dereferences its Graph
+    // (nodeExecuted(), getState(), ...) via a raw back-pointer. If a transition is
+    // still completing on a worker thread (e.g. an in-progress switch to Off that
+    // is allowed to continue during shutdown), destroying the graphs first would be
+    // a use-after-free.
     worker_threads_.reset();
     worker_jobs_.reset();
+
+    process_groups_.clear();
```
WilliamRoebuck: I think this might not be necessary. The situation the comment describes no longer exists and the worker threads don't have any access to Graph

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -350,6 +358,8 @@ bool ProcessGroupManager::run()
 
             watchdog_->serviceWatchdog();
         }
+        LM_LOG_WARN() << "ProcessGroupManager::run() - received SIGTERM, exiting";
```
WilliamRoebuck: Should be debug level I think

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -457,15 +467,30 @@ void ProcessGroupManager::allProcessGroupsOff()
...
+    // Each worker deactivating a process already honours that process's configured
+    // shutdown_timeout: it sends SIGTERM, waits up to termination_timeout_ms_ for a
+    // graceful exit and only then SIGKILLs (waiting a further kMaxSigKillDelay).
+    // Bound the wait for the whole transition to Off by the slowest such process
+    // rather than a fixed grace period, so every component's shutdown_timeout is
+    // respected. Processes deactivate in parallel, so the largest per-process
+    // timeout (plus the SIGKILL grace) bounds how long the transition can take.
+    // Scope this to the process nodes of the graph being shut down (its own
+    // process group) that still have a live process to stop, rather than the whole
+    // configuration - already terminated or never-started processes are not waited on.
```
WilliamRoebuck: Far too long of a comment 😆 . I don't think this is needed at all

tests/integration/lm_shutdown_during_rt_switch/component_a.cpp
```diff
@@ -0,0 +1,61 @@
...
+/// run-target switch the launch manager runs the STOP phase (terminating
+/// no-longer-needed processes) fully before the START phase (activating newly
+/// needed processes). By stalling here, component_a keeps the switch in the STOP
+/// phase, giving the test a deterministic window to send SIGTERM to the launch
+/// manager before run_target_c could ever be activated.
+///
+/// It must be larger than the time the test needs to observe `a_terminating` and
+/// deliver the SIGTERM, and smaller than component_a's configured
+/// shutdown_timeout so the process still exits gracefully (and writes its XML
+/// result) rather than being SIGKILLed.
```
WilliamRoebuck: This could probably be two lines

tests/integration/lm_shutdown_during_rt_switch/lm_shutdown_during_rt_switch.py
```diff
@@ -0,0 +1,160 @@
...
+        )
+
+        # Request shutdown: send SIGTERM to the launch manager process only, so
+        # that the launch manager itself stops the processes it owns (rather than
+        # the OS terminating the whole process group directly).
+        lm_pid = _launch_manager_pid(target)
+        assert lm_pid is not None, "Could not find the running launch_manager process"
+        logger.info(f"Sending SIGTERM to launch_manager (pid {lm_pid})")
+        exit_code, _ = target.execute(f"kill -15 {lm_pid}")
+        assert exit_code == 0, "Failed to send SIGTERM to the launch manager"
```
WilliamRoebuck: Would it be possible to achieve the same behaviour simply using the normal run until file deployed method? This sends a sigterm to lm and lets it terminate gracefully

tests/integration/lm_shutdown_during_switch_to_off/component_a.cpp
```diff
@@ -0,0 +1,64 @@
...
+/// the "Off" run target) in the STOP phase, giving the test a deterministic window to send SIGTERM to the launch
+/// manager while the switch to Off is still in progress.
+///
+/// It must be comfortably larger than the time the test needs to observe
+/// `a_terminating` and deliver the SIGTERM to the launch manager, so the switch
+/// to Off is still in progress when that SIGTERM arrives. It must also be
+/// comfortably smaller than component_a's configured shutdown_timeout: the launch
+/// manager honours that per-process shutdown_timeout during its own shutdown, so
+/// component_a is given time to exit on its own and terminates gracefully (and
+/// writes its XML result) rather than being force-terminated (SIGKILLed).
```
WilliamRoebuck: This can probably be 2 lines also

tests/integration/lm_shutdown_during_switch_to_off/lm_shutdown_during_switch_to_off.py
```diff
@@ -0,0 +1,189 @@
...
+        _wait_for_file(target, a_terminating, proc, timeout_s=10.0)
+
+        # Request shutdown: send SIGTERM to the launch manager process only, so
+        # that the launch manager itself stops the processes it owns (rather than
+        # the OS terminating the whole process group directly).
+        lm_pid = _launch_manager_pid(target)
+        assert lm_pid is not None, "Could not find the running launch_manager process"
+        logger.info(f"Sending SIGTERM to launch_manager (pid {lm_pid})")
+        exit_code, _ = target.execute(f"kill -15 {lm_pid}")
+        assert exit_code == 0, "Failed to send SIGTERM to the launch manager"
```
WilliamRoebuck: Same comment as the other python file, hopefully we can remove these two methods

tests/integration/lm_shutdown_during_switch_to_off/control_client_test_driver.cpp
```diff
@@ -0,0 +1,75 @@
...
+// shutdown must simply let that in-progress switch to Off continue to completion
+// - it must NOT cancel the explicit switch to Off and redo it. Either way the
+// launch manager must end up stopping everything it owns and exit cleanly.
+TEST(LmShutdownDuringSwitchToOff, ControlClient)
+{
+    score::mw::lifecycle::ControlClient client{};
+    ASSERT_TRUE(check_clean({test_end_location, a_started, a_terminating}));
+
+    const auto pid = getpid();
+    const std::string step_msg = "Report running with pid == " + std::to_string(pid);
```
WilliamRoebuck, line 42: Please don't rely on log messages for tests! If we use the normal shutdown procedure as I mentioned in the other test's python file, this shouldn't be needed

tests/utils/test_helper/test_helper.hpp
```diff
@@ -166,6 +169,31 @@ class TestRunner
...
+    ///
+    /// Poll the atomic flag instead of calling pause() directly. A process-directed
+    /// signal (from kill) is delivered to an arbitrary one of the process's threads
+    /// that has the signal unblocked. If the test process has more than one thread
+    /// (e.g. a ControlClient, whose constructor spawns a background IPC thread), the
+    /// handler may run on a background thread and set exitRequested there. pause()
+    /// only returns for a signal handled on the calling thread, so a main thread
+    /// blocked in pause() would not wake - it would stay parked until the launch
+    /// manager SIGKILLs it at its shutdown_timeout, and never write its XML result.
+    /// Polling observes the flag regardless of which thread handled the signal.
```
WilliamRoebuck: This doesn't sound correct to me. If the child thread caught the signal, would exit requested actually be set? I ran the test using the normal `pause()` 450 times and didn't get any failures. I would leave these changes out of this PR and create an issue if something pops up again

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
TimoSteuerwaldETAS: I would stick at least to info level, as this is a major state change to the whole launch manager.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
TimoSteuerwaldETAS: Shortened

tests/integration/lm_shutdown_during_rt_switch/lm_shutdown_during_rt_switch.py
TimoSteuerwaldETAS: Thanks for this great comment!
This is actually true after looking into the implementation of https://github.com/eclipse-score/itf/blob/1be7deaf8f535984d214aa97db0bada4b89c2ed4/score/itf/plugins/docker.py#L108

tests/integration/lm_shutdown_during_switch_to_off/component_a.cpp
TimoSteuerwaldETAS: May be not two, but it should be shorter for sure

tests/integration/lm_shutdown_during_switch_to_off/lm_shutdown_during_switch_to_off.py
TimoSteuerwaldETAS: Yes, we can. 😃

tests/integration/lm_shutdown_during_switch_to_off/control_client_test_driver.cpp
TimoSteuerwaldETAS: The PID is not evaluated in the test. I have added it to ease manual debugging. As launch manager logs the PID, I think it's useful to have them at hand for a failed test run in pipeline.

tests/utils/test_helper/test_helper.hpp
TimoSteuerwaldETAS: `exitRequested` is set to true by the signalHandler. However it is not clear on which thread the signal handler gets executed, therefore it is also not clear if `pause()` returns.

See also on manpage for [signal](https://man7.org/linux/man-pages/man7/signal.7.html):

> A process-directed signal may be delivered to any one of the threads that does not currently have the signal blocked.  If more than one of the threads has the signal unblocked, then the kernel chooses an arbitrary thread to which to deliver the signal.

See also this example: https://godbolt.org/z/3fjPoPMos :
In my test runs the background thread which is hanging on `pause()` is never waked up, but of course this also happened accidentally. The difference to our code: In our code the main thread calls `pause()`, here the background thread, however I would expect this does not influence the behavior.

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
TimoSteuerwaldETAS: After offline clarification, I have left the code as it is, but updated the comment.

tests/integration/lm_shutdown_during_rt_switch/lm_shutdown_during_rt_switch.py
```diff
@@ -0,0 +1,66 @@
...
+
+    new_config_path = str(remote_test_dir / "etc/lm_shutdown_during_rt_switch.bin")
+    a_terminating = remote_test_dir / "component_a_terminating"
+
+    # Run the launch manager until component_a signals it is stalling mid-termination
+    # (file `component_a_terminating`): the switch to run_target_c is then in progress
+    # but run_target_c has not been activated yet. run_until_file_deployed stops the
+    # launch manager at that point by sending it a SIGTERM (to the launch manager
+    # process only, so it performs its own orderly shutdown) and asserts it exits
+    # cleanly (code 0).
```
WilliamRoebuck:
```suggestion
    # Run until `component_a_terminating` is deployed so we can request shutdown during
    # the transition to run target c
```

tests/integration/lm_shutdown_during_switch_to_off/control_client_test_driver.cpp
WilliamRoebuck: Ah I see, I misunderstood

tests/integration/lm_shutdown_during_switch_to_off/lm_shutdown_during_switch_to_off.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+
+    new_config_path = str(remote_test_dir / "etc/lm_shutdown_during_switch_to_off.bin")
+    a_terminating = remote_test_dir / "component_a_terminating"
+
+    # Run the launch manager until component_a signals it is stalling mid-termination
+    # (file `component_a_terminating`): the explicit switch to Off is then in progress.
+    # run_until_file_deployed stops the launch manager at that point by sending it a
+    # SIGTERM (to the launch manager process only, so it performs its own orderly
+    # shutdown) and asserts it exits cleanly (code 0).
```
WilliamRoebuck:
```suggestion
    # Run until `component_a_terminating` is deployed so we can send SIGTERM 
    # to launch manager during the transition to Off
```

tests/integration/readme.md
```diff
@@ -23,3 +23,74 @@ Currently the following configs are supported:
...
+  it manually:
+  ```
+  cat /proc/sys/kernel/core_pattern
+  # On WSL, restore the default handler:
+  docker run --rm --privileged debian:bookworm-slim \
+    bash -c 'echo "|/wsl-capture-crash %t %E %p %s" > /proc/sys/kernel/core_pattern'
+  ```
+  The default `core_pattern` differs per system (e.g. `|/wsl-capture-crash ...`
+  on WSL, `|/lib/systemd/systemd-coredump ...` on systemd hosts, or plain
+  `core`). Check yours *before* the first run so you know the value to restore.
```
WilliamRoebuck: If this is important to do before running, we should put this operation under a config flag I think. Its very unlikely that anyone will see this before they run tests the first time (after this PR)

tests/utils/test_helper/test_helper.hpp
```diff
@@ -166,6 +169,20 @@ class TestRunner
     /// @brief True if the test process has received SIGINT or SIGTERM
     inline static std::atomic<bool> exitRequested = false;
 
+    /// @brief Block the calling thread until a SIGINT/SIGTERM sets exitRequested.
+    /// @details Polls the flag instead of pause(): a process-directed signal may be
+    /// handled on a background thread (e.g. the ControlClient's IPC thread), which a
+    /// main thread parked in pause() would never observe.
+    // TODO: adopt this in the remaining tests/integration/*/*control*.cpp binaries
+    // that still wait with a bare pause().
```
cameron-craig-etas: Do we have a Github issue for this? We could make sure we have one and remove this TODO.

tests/utils/test_helper/process_hanging_on_sigterm.cpp
```diff
@@ -0,0 +1,73 @@
...
+#include <string_view>
+
+#include "tests/utils/test_helper/test_helper.hpp"
+#include <score/mw/lifecycle/report_running.h>
+
+namespace
+{
+/// @brief How long the process stalls while being terminated, keeping an in-progress
+/// run-target switch alive so the test has a deterministic window to send SIGTERM to
+/// the launch manager. Must be smaller than the configured shutdown_timeout so the
```
WilliamRoebuck:
```suggestion
/// @brief How long the process stalls while being terminated. 
/// Must be smaller than the configured shutdown_timeout so the
```

cameron-craig-etas (changes requested):

tests/utils/test_helper/test_helper.hpp
WilliamRoebuck: Ah, but I think that is the key difference. From my testing, a pause on the main thread **always** takes precedence to pauses on child threads. Though this might be OS dependent. I think this should be a separate issue/PR anyway as it is not needed for this test.

WilliamRoebuck (commented):
Looks good now, my only remaining concerns are the core dump and test wait changes

tests/utils/test_helper/test_helper.hpp
TimoSteuerwaldETAS: Good catch. I can drop this. It is only used in test applications, but not in control_client_test_drivers.

TimoSteuerwaldETAS: After another call with William I removed the whole method again.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.hpp
TimoSteuerwaldETAS: Fixed.

tests/integration/readme.md
TimoSteuerwaldETAS: Moved to separate branch.

cameron-craig-etas (approved):

MaciejKaszynski (approved):

TimoSteuerwaldETAS:
@WilliamRoebuck After discussion with @NicolasFussberger I have changed the implementation to use the transition timeout of the Off Run Target. However, as this is not a complete support of the transition timeout as described in #496 , I have not set this issue to done.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -92,6 +99,7 @@ void CreateDependencyGraph(
     {
         const auto off_index = graph.emplace(std::in_place_type<RunTarget>, graph.size());
         run_target_map.insert({IdentifierHash{Graph::off_state_name}.data(), off_index});
+        off_state_transition_timeout = std::chrono::milliseconds(3000);
```
NicolasFussberger: Maybe make this a static constant somewhere so the default is not hidden as a magic number in the code

score/launch_manager/src/daemon/src/process_group_manager/details/graph_UT.cpp
```diff
@@ -411,6 +411,47 @@ TEST_F(GraphImplicitOffTargetTest, offRunTargetIsCreatedWhenNotConfigured)
...
+/// @brief Fixture that configures the "Off" run target with a distinctive transition timeout.
+class GraphOffStateTimeoutTest : public GraphTest
+{
+  protected:
+    void SetConfig() override
+    {
+        auto procs = generateProcessComponents(1);
+        auto rts = generateRunTargets(1);
+        rts[1].depends_on = {procs[0].name};
+        // The Off run target is the last entry generateRunTargets() appends.
```
NicolasFussberger: Maybe add an assertion for this. Someone might modify the function in the future and this is no longer true

NicolasFussberger (approved):

PR #455, opened 2026-08-12
Removing Config Adapter

### LINKED ISSUE #420: Remove Configuration Adapter
### What

Now that the new config is being used we shall remove the temporary adapter that we made.

Note from the code:
```
Temporary bridge that translates the new Config model (RunTargets, Components) into the
 legacy API (ProcessGroups, ProcessGroupStates, OsProcess) consumed by ProcessGroupManager, Graph, and related code
```

### Acceptance Criteria (DoD)

- Removed Configuration Adapter

### How

_No response_

MaciejKaszynski:
Fixes https://github.com/eclipse-score/lifecycle/issues/420

Currently we have a lot of code to bridge the old config and new. This PR removes the bridge and changes. Process Group Manager, Graph and Alive Monitor to use the new config.


score/launch_manager/src/control_client/src/details/control_client_impl.cpp
```diff
@@ -333,6 +334,8 @@ score::concurrency::InterruptibleFuture<void> ControlClientImpl::SetState(
 {
     score::concurrency::InterruptibleFuture<void> retVal_{};
 
+    std::cout << "Setting state " << pg_name << pg_state << std::endl;
```
danth: Use `LM_LOG_DEBUG`, or remove this.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
+/// @param supervision_event_publisher Publisher for supervision events.
+/// @param process_interface Interface for OS process operations.
+/// @param process_map Map for tracking process PIDs.
+/// @return A populated dependency graph with all components and run targets.
+DependencyGraph<Graph::Component> CreateDependencyGraph(
+    configuration::Config* config,
+    ISupervisionEventPublisher& supervision_event_publisher,
+    osal::IProcess* process_interface,
+    std::shared_ptr<SafeProcessMapInserter> process_map,
+    std::unordered_map<std::size_t, GraphIndex>& run_target_map)
```
danth, line 43: `run_target_map` is not documented

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
-    off_state_ = configuration_->getNameOfOffState(pg_name);
-    requested_state_.pg_state_name_ = off_state_;
-    requested_state_.pg_name_ = pg_name;
+    const auto& run_targets = config->runTargets();
+    auto components = config->takeComponents();
+    const auto num_of_components = components.size();
 
-    LM_LOG_DEBUG() << "Process group index" << index << "(with name" << pg_name << ") has" << num_processes
-                   << "processes";
+    DependencyGraph<Graph::Component> graph(components.size() + run_targets.size() + 1);
```
danth: Should we add a comment explaining why the `+ 1` is necessary?

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
+            static_cast<uint32_t>(graph_index),
+            supervision_event_publisher,
+            process_interface,
+            process_map);
+
+        LM_LOG_DEBUG() << "Creating component node:" << component_name << "at index:" << index;
+        component_name_to_index[component_name] = index;
+
+        // all of this relies on graphindex having sequential indexes
+        // TODO use IdHash later...
```
danth: Maybe add a link to #463 and explain that this method will be simplified after that change.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
+    };
+
+    for (std::size_t i = 0; i < run_targets.size(); ++i)
     {
-        if (state_name == pg_state)
+        for (const auto& dep_name : run_targets[i].depends_on)
         {
-            return static_cast<int32_t>(index);
+            auto it = component_name_to_index.find(dep_name);
+            LM_LOG_ERROR() << run_targets[i].name << "HAS DEP TO " << dep_name;
```
danth:
```suggestion
            LM_LOG_DEBUG() << run_targets[i].name << "HAS DEP TO " << dep_name;
```

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
-}
 
-bool Graph::nodeHasTerminatedDeps(IdentifierHash pg_name, uint32_t node_index)
-{
-    const DependencyList* dep_list = configuration_->getOsProcessDependencies(pg_name, node_index).value_or(nullptr);
+    auto fallback_index = graph.emplace(std::in_place_type<RunTarget>, graph.size());
+    run_target_map.insert({IdentifierHash{"fallback"}.data(), fallback_index});
 
-    if (dep_list && dep_list->size() > 0)
+    LM_LOG_ERROR() << "FALLBACK at index:" << fallback_index;
```
danth:
```suggestion
    LM_LOG_DEBUG() << "FALLBACK at index:" << fallback_index;
```

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
+    for (const auto& dep_name : config->fallbackRunTarget().depends_on)
     {
-        return (*dep_list)[0].process_state_ == ProcessState::kTerminated;
+        auto it = component_name_to_index.find(dep_name);
+        LM_LOG_ERROR() << dep_name;
+        LM_LOG_ERROR() << "FALLBACK" << fallback_index << "HAS DEP TO " << dep_name;
```
danth:
```suggestion
        LM_LOG_DEBUG() << dep_name;
        LM_LOG_DEBUG() << "FALLBACK" << fallback_index << "HAS DEP TO " << dep_name;
```

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -360,6 +371,7 @@ void Graph::startTransition(IdentifierHash pg_state)
     queueReadyNodes();
     if (current_transition_->isFinished())
     {
+        LM_LOG_DEBUG() << "Finsihed?";
```
danth: This message should be improved (or deleted).

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
```diff
@@ -33,24 +34,33 @@ ProcessInfoNode::ProcessInfoNode(
...
+            "LCM_ALIVE_INTERFACE_PATH", "/lifecycle_health_" + config_.name);
+    }
 }
 
 IComponent::RequestResult ProcessInfoNode::tryReportCompletion(score::mw::lifecycle::ProcessState new_state)
 {
     ProcessState desired_state{};
-    switch (ready_condition_)
+
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(config_.component_properties.ready_condition.has_value() == true, "ADS");
```
danth: What is ADS?

score/launch_manager/src/daemon/src/process_group_manager/details/transition.hpp
```diff
@@ -24,6 +24,8 @@
 #include <bitset>
 #include <cstddef>
 #include <cstdint>
+#include <ios>
+#include <iostream>
```
danth: Not needed.

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -370,47 +309,39 @@ void ProcessGroupManager::processComponentEvents()
...
+    graph_->startInitialTransition(IdentifierHash(initial_target));
+    return true;
 }
 
 void ProcessGroupManager::allProcessGroupsOff()
 {
-    // Wait for process group states to change while actively draining shutdown events.
+    if (!graph_)
+    {
+        return;
```
danth: Should we log a warning here?

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -553,21 +479,22 @@ void ProcessGroupManager::controlClientRequests(Graph& pg)
...
-                       << scc->request().process_group_state_.pg_state_name_ << "of PG"
-                       << scc->request().process_group_state_.pg_name_;
+                       << scc->request().process_group_state_.pg_state_name_;
 
+        LM_LOG_DEBUG() << "HERER++++++++";
         // Now process the request
         switch (scc->request().request_or_response_)
         {
             case ControlClientCode::kSetStateRequest:
+                LM_LOG_DEBUG() << "HERER++++++++";
```
danth: Remove print debugging.

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -794,9 +713,10 @@ void ProcessGroupManager::processGroupHandler(Graph& pg)
 
             ProcessGroupStateID recovery_state;
             recovery_state.pg_name_ = pg.getProcessGroupName();
-            recovery_state.pg_state_name_ = configuration_.getNameOfRecoveryState(recovery_state.pg_name_);
+            // TODO: Determine recovery state from configuration
+            recovery_state.pg_state_name_ = IdentifierHash("fallback");  // TODO: Get from configuration
```
danth: Does this need to be done?

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,168 +28,176 @@
...
-        if (state_name == pg_state)
+        for (const auto& dep_name : run_targets[i].depends_on)
         {
-            return static_cast<int32_t>(index);
+            auto it = component_name_to_index.find(dep_name);
+            LM_LOG_ERROR() << run_targets[i].name << "HAS DEP TO " << dep_name;
+            SCORE_LANGUAGE_FUTURECPP_PRECONDITION_MESSAGE(
+                it != component_name_to_index.end(), "RunTarget dependency not found in component list");
+
+            graph.addDependency(num_of_components + i, it->second);
```
danth: I think this would be clearer if we had something like

```cpp
for (std::size_t run_target_i = 0; run_target_i < run_targets.size(); ++run_target_i) {
  std::size_t node_i = num_of_components + run_target_i;
```

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
MaciejKaszynski: Think I will actually make all graph_ checks to asserts as this shall never happen. We could actually make this not a shared_ptr and then remove the checks at all, but I will make an issue for this to do in another PR

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
MaciejKaszynski: forgot to add a message :D

score/launch_manager/src/control_client/src/details/control_client_impl.cpp
```diff
@@ -14,6 +14,7 @@
 #include <sys/stat.h>
 #include <cerrno>
 #include <cstdint>
+#include <iostream>
```
WilliamRoebuck: Remove

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_config_loader_UT.cpp
```diff
@@ -294,6 +297,28 @@ TEST_F(FlatbufferConfigLoaderTest, LoadRunTargets)
     EXPECT_THAT(target.recovery_action.run_target, Eq("SafeState"));
 }
 
+TEST_F(FlatbufferConfigLoaderTest, ConfiguredOffRunTargetIsLoadedVerbatim)
```
WilliamRoebuck, line 300: Can we add or extend a test to check that an "Off" target is created if none exists?

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
 
-Graph::Graph(
-    uint32_t max_num_nodes,
-    ConfigurationAdapter* configuration,
-    std::shared_ptr<WorkerQueue> job_queue,
+namespace
+{
+
+using namespace score::mw::lifecycle::internal;
+namespace configuration = score::mw::lifecycle::internal::configuration;
```
WilliamRoebuck: These are probably not needed anymore

score/launch_manager/src/daemon/src/process_group_manager/details/graph.hpp
```diff
@@ -146,11 +146,14 @@ static constexpr GraphState state_results[][static_cast<uint>(GraphState::kUndef
 class Graph final
 {
   public:
+    /// @brief All possible implementation of a Component.
```
WilliamRoebuck:
```suggestion
    /// @brief All currently supported component implementations
```

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
+    for (std::size_t comp_dep_i = 0; comp_dep_i < component_dependencies.size(); ++comp_dep_i)
+    {
+        for (const auto& dep_name : component_dependencies[comp_dep_i])
+        {
+            auto it = component_name_to_index.find(dep_name);
+            SCORE_LANGUAGE_FUTURECPP_PRECONDITION_MESSAGE(
+                it != component_name_to_index.end(), "Component dependency not found in component list");
 
-    SCORE_LANGUAGE_FUTURECPP_ASSERT_DBG_MESSAGE(states != nullptr, "Process group states not found for process group");
+            graph.addDependency(comp_dep_i, it->second);
```
WilliamRoebuck: This addDependency section is repeated 3 times, can we add a helper function/lambda? Or even better maybe we can iterate through all the dependencies in one loop somehow?

WilliamRoebuck: Maybe you can add everything to the graph and then iterate through graph items, accessing the config by index?

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -360,6 +359,7 @@ void Graph::startTransition(IdentifierHash pg_state)
     queueReadyNodes();
     if (current_transition_->isFinished())
     {
+        LM_LOG_DEBUG() << "Finsihed?";
```
WilliamRoebuck: Improve or remove these logs

score/launch_manager/src/daemon/src/process_group_manager/details/graph_UT.cpp
```diff
@@ -107,11 +120,10 @@ class GraphTest : public ::testing::Test
...
     }
 
+    /// @brief Run targets are registered in the graph under their plain configured name.
     IdentifierHash state_name(std::string_view run_target)
     {
-        const auto left = std::string{pg_string};
-        const auto right = std::string{run_target};
-        return IdentifierHash{left + "/" + right};
+        return IdentifierHash{run_target};
     }
```
WilliamRoebuck: We should probably remove this function now

score/launch_manager/src/daemon/src/process_group_manager/details/graph_UT.cpp
```diff
@@ -163,36 +175,28 @@ class GraphTest : public ::testing::Test
...
         }
 
-        ASSERT_EQ(graph_.getState(), GraphState::kSuccess);
-        ASSERT_EQ(graph_.getProcessGroupState(), target);
+        ASSERT_EQ(graph_->getState(), GraphState::kSuccess);
+        ASSERT_EQ(graph_->getProcessGroupState(), target);
     }
 
-    ConfigurationAdapter config_{};
+    std::optional<Config> config_{};
```
WilliamRoebuck, line 182: Why is this an optional?

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
```diff
@@ -185,9 +198,17 @@ IComponent::RequestResult ProcessInfoNode::tryHandleTermination(int32_t process_
...
+                   << config_.deployment_config.bin_dir << "/" << config_.component_properties.binary_name;
+
+    // Read configured restart attempts
+    uint32_t restart_counter = 0;
+    if (config_.deployment_config.ready_recovery_action.has_value())
+    {
+        restart_counter = config_.deployment_config.ready_recovery_action->number_of_attempts;
+    }
+
+    LM_LOG_DEBUG() << "Process" << config_.name << "configured with" << restart_counter << "restart attempts";
```
WilliamRoebuck: I think we should do this at initialisation instead

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.hpp
```diff
@@ -36,14 +38,10 @@ namespace score::mw::lifecycle::internal
...
 {
-  public:
-    /// @brief The criteria for when a process is considered "ready"
-    enum class ReadyCondition : uint8_t
-    {
-        kRunning,     // Running reported in the case of a reporting process, process launched if non-reporting
-        kTerminated,  // Process has terminated with status 0
-    };
 
+    using ReportStateFn = std::function<bool(IdentifierHash, ProcessState, timespec)>;
```
WilliamRoebuck: This shouldn't be here anymore

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -242,22 +257,25 @@ OsalReturnType ProcessLauncher::startProcess(ProcessID* pid, IpcCommsP* block, c
...
+    int& fd,
+    const score::mw::lifecycle::internal::configuration::ComponentConfig& config)
 {
     bool comms_result = true;
     char shm_name[static_cast<uint32_t>(score::mw::lifecycle::internal::ProcessLimits::maxLocalBuffSize)];
     size_t length = sizeof(IpcCommsSync);
 
-    if (CommsType::kControlClient == config.comms_type_)
+    auto app_type = config.component_properties.application_profile.application_type;
+    if (score::mw::lifecycle::internal::configuration::ApplicationType::StateManager == app_type)
```
WilliamRoebuck:
```suggestion
    if (configuration::ApplicationType::StateManager == app_type)
```

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -242,22 +257,25 @@ OsalReturnType ProcessLauncher::startProcess(ProcessID* pid, IpcCommsP* block, c
...
     }
 
     return result;
 }
 
-bool ProcessLauncher::setupComms(IpcCommsP& block, int& fd, const OsalConfig& config)
+bool ProcessLauncher::setupComms(
+    IpcCommsP& block,
+    int& fd,
+    const score::mw::lifecycle::internal::configuration::ComponentConfig& config)
```
WilliamRoebuck:
```suggestion
    const configuration::ComponentConfig& config)
```

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
WilliamRoebuck: There are a few more of these that can be shortened

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -298,10 +319,23 @@ bool ProcessLauncher::setupComms(IpcCommsP& block, int& fd, const OsalConfig& co
...
+                block->comms_type_ = CommsType::kControlClient;
+            }
+            else if (app_type == score::mw::lifecycle::internal::configuration::ApplicationType::Native)
+            {
+                block->comms_type_ = CommsType::kNoComms;
+            }
+            else
+            {
+                block->comms_type_ = CommsType::kReporting;
+            }
```
WilliamRoebuck: Switch case here?

score/launch_manager/src/daemon/src/process_group_manager/details/process_launcher.cpp
```diff
@@ -346,7 +382,8 @@ bool ProcessLauncher::initializeSemaphores(IpcCommsP shared_block)
 }
 
 /// @details The implementation should be async signal safe.
-OsalReturnType ProcessLauncher::setSchedulingAndSecurity(const OsalConfig& config)
+OsalReturnType ProcessLauncher::setSchedulingAndSecurity(
+    const score::mw::lifecycle::internal::configuration::ComponentConfig& config)
```
WilliamRoebuck: Can we pass just the deployment config here?

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -152,8 +134,8 @@ void ProcessGroupManager::deinitialize()
     os_handler_.reset();
     process_monitor_.reset();
     alive_monitor_thread_->stop();
-    configuration_.deinitialize();
-    process_groups_.clear();
+    // No deinitialize needed - Config destructor handles cleanup
```
WilliamRoebuck: I don't think we need this comment

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -370,47 +305,33 @@ void ProcessGroupManager::processComponentEvents()
...
-            result = true;
-        }
-    }
-    else
-    {
-        LM_LOG_ERROR() << "No startup state, exiting from process group manager";
-    }
-    return result;
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(bool(graph_), "Graph not initialized");
+    // Convert string_view to IdentifierHash
```
WilliamRoebuck: There's a constructor at IdentifierHash.cpp:66, doesn't it work?

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -794,9 +697,10 @@ void ProcessGroupManager::processGroupHandler(Graph& pg)
 
             ProcessGroupStateID recovery_state;
             recovery_state.pg_name_ = pg.getProcessGroupName();
-            recovery_state.pg_state_name_ = configuration_.getNameOfRecoveryState(recovery_state.pg_name_);
+            // TODO: Determine recovery state from configuration
+            recovery_state.pg_state_name_ = IdentifierHash("fallback");  // TODO: Get from configuration
```
WilliamRoebuck: Double `// TODO:` 🙂

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -794,9 +697,10 @@ void ProcessGroupManager::processGroupHandler(Graph& pg)
...
 
-            LM_LOG_WARN() << "Problem discovered in PG" << recovery_state.pg_name_ << "Activating Recovery state.";
+            LM_LOG_WARN() << "Problem discovered, activating recovery state: " << recovery_state.pg_state_name_;
```
WilliamRoebuck:
```suggestion
            LM_LOG_WARN() << "Problem discovered, activating recovery state:" << recovery_state.pg_state_name_;
```
Extra space

WilliamRoebuck (changes requested):
Looks good, I have mostly small comments, one or two suggestions for code changes

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
-    for (const auto& [state_name, index] : run_targets_)
+    // handle the off target
+    if (!off_rt_defined)
     {
-        if (state_name == pg_state)
-        {
-            return static_cast<int32_t>(index);
-        }
+        auto off_index = graph.emplace(std::in_place_type<RunTarget>, graph.size());
+        run_target_map.insert({IdentifierHash{"Off"}.data(), off_index});
```
NicolasFussberger: The "Off" name exists in many locations as hardcoded string. Maybe we can move this to a single constant

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
-Graph::~Graph()
+    std::unordered_map<std::size_t, GraphIndex>& run_target_map)
 {
-    LM_LOG_DEBUG() << "Graph destroyed";
-}
+    // this is a temporary (bad) implementation, all shall be cleandup
+    // on https://github.com/eclipse-score/lifecycle/issues/463
+    // making the dep_graph a hash map would make this much cleaner as we
+    // wouldn't have to keep track of stuff...
+    const std::vector<configuration::RunTargetConfig> run_targets = config.takeRunTargets();
```
NicolasFussberger: process group manager still uses configuration_.runTargets() even though its been moved out here

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
-{
-    pg_index_ = index;
-    off_state_ = configuration_->getNameOfOffState(pg_name);
-    requested_state_.pg_state_name_ = off_state_;
-    requested_state_.pg_name_ = pg_name;
+    // the Off and fallback run targets are special.
+    // Off can be configred by the user and would mean we need +1 but
+    // over-reserving is ok.
+    // fallback is always created.
+    const std::size_t graph_size = components.size() + run_targets.size() + 2U;
```
NicolasFussberger: This calculation exists in a duplicate way also here in process_group_manager.cpp:207

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
         SCORE_LANGUAGE_FUTURECPP_ASSERT_DBG_MESSAGE(
-            emplaced_index == node_index, "RunTarget index must match its position in the graph");
-        run_targets_.emplace_back(state.name_, node_index);
+            index == graph_index, "Component graph indices must match component order");
 
-        for (const auto process_index : state.process_indexes_)
+        off_rt_defined |= bool(run_target.name == "Off");
+        component_name_to_index[run_target.name] = index;
+        run_target_map.insert({IdentifierHash{run_target.name}.data(), index});
+    };
```
NicolasFussberger:
```suggestion
    }
```

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
+        LM_LOG_DEBUG() << "Created RunTarget node:" << run_target.name << "at index" << index;
+
         SCORE_LANGUAGE_FUTURECPP_ASSERT_DBG_MESSAGE(
-            emplaced_index == node_index, "RunTarget index must match its position in the graph");
-        run_targets_.emplace_back(state.name_, node_index);
+            index == graph_index, "Component graph indices must match component order");
 
-        for (const auto process_index : state.process_indexes_)
+        off_rt_defined |= bool(run_target.name == "Off");
+        component_name_to_index[run_target.name] = index;
```
NicolasFussberger: is it ensured that component and run target names are unique, otherwise this may silently overwrite a preexisting name

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
```diff
@@ -27,221 +28,217 @@
...
+namespace configuration = score::mw::lifecycle::internal::configuration;
+
+/// @brief Creates a dependency graph from the configuration.
+/// @param config The configuration containing components and run targets.
+/// @param supervision_event_publisher Publisher for supervision events.
+/// @param process_interface Interface for OS process operations.
+/// @param process_map Map for tracking process PIDs.
+/// @param run_target_map Map to keep the translation between IDHash to Index
+/// @return A populated dependency graph with all components and run targets.
+DependencyGraph<Graph::Component> CreateDependencyGraph(
```
NicolasFussberger: Its a rather long private method with many parameters. I wonder if this can be refactored to make it simpler and more easily testable by having e.g. a separate GraphBuilder class or something like that

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -220,48 +202,18 @@ bool ProcessGroupManager::initializeControlClientHandler()
...
-    }
-    else
-    {
-        LM_LOG_ERROR() << "Failed to initialize process groups";
-    }
-
-    return success;
+    graph_ = std::make_shared<Graph>(
+        // size is +2 for fallback + off
+        configuration_.components().size() + configuration_.runTargets().size() + 2,
```
NicolasFussberger, line 206: I wonder if this size can be inferred from the configuration inside the Graph, so that this calculation does not need to exist multiple times

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -358,9 +295,7 @@ bool ProcessGroupManager::run()
 
 void ProcessGroupManager::processComponentEvents()
 {
-    // Single-graph assumption: PGM creates one ProcessMonitor bound to the first (only) graph, so
-    // every event always applies to it. Multi-graph routing is deferred to a future revision.
-    Graph& graph = *process_groups_.front();
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(bool(graph_), "Graph not initialized");
```
NicolasFussberger, line 295: I suppose we should probably use static_cast over c-style casts

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -609,30 +524,26 @@ void ProcessGroupManager::controlClientRequests(Graph& pg)
...
-    }
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(bool(graph_), "Graph not initialized");
 
-    const IdentifierHash old_state = pg->getProcessGroupState();
-    const IdentifierHash recovery_state = configuration_.getNameOfRecoveryState(pg->getProcessGroupName());
-    const GraphState graph_state = pg->getState();
+    const IdentifierHash old_state = graph_->getProcessGroupState();
+    // the fallback state doesn't have a name in the config, so we use
+    // "fallback", it doesn't actually matter...
+    const IdentifierHash recovery_state("fallback");
```
NicolasFussberger: This name exists in multiple places as hardcoded string and they need to be identical for things to work, so I would recommend to move this to a single constant.

NicolasFussberger (changes requested):

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
danth: This method will already need to be rewritten as part of #463, so I think it could be left for a separate PR?

NicolasFussberger: Will this ticket really dramatically change the scope of this method?

danth: I suppose not from an external perspective, but I think the implementation inside will change a lot.

MaciejKaszynski: I don't think we need a builder, after https://github.com/eclipse-score/lifecycle/issues/463 this will simply just loop over all components and RTs (We can probably also start creating the dep graph inside of config during loading) as we can add dependencies without inserting them before. I will make the proc map, proc launcher and the supervision publisher into a struct as it would also make the graph constructor smaller.

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
MaciejKaszynski: I actually found that the only method in PGM that uses it is not used at all so deleted the PGM method. [92c9c13](https://github.com/eclipse-score/lifecycle/pull/455/commits/92c9c134328274afecac20ee18f4cb84ab277668)

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
MaciejKaszynski: That would mean that if a process is restarts N times during one activation it would carry over when changing the run target. This is just restarting the counter

score/launch_manager/src/daemon/src/configuration/details/flatbuffer_config_loader_UT.cpp
MaciejKaszynski: I've moved that to the graph to handle now, so will add to graph_ut

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
MaciejKaszynski: I've simplified without lambdas, now we just iterate throught comps, then RTs then wire up deps

score/launch_manager/src/daemon/src/process_group_manager/details/graph_UT.cpp
MaciejKaszynski: It's not easily changeable because it's not default construable, and gtests classes don't have a constructor so.

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
MaciejKaszynski: I removed the second calc, I'm thinking the graph could be constructed during config load, but this should be done in #463

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
MaciejKaszynski: That's not a cast, I'm using the bool operator to check if the pointer is valid

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
MaciejKaszynski: Added this to the script now

score/launch_manager/src/daemon/src/process_group_manager/details/graph.cpp
MaciejKaszynski: This is the single one now.

score/launch_manager/src/daemon/src/process_group_manager/details/process_info_node.cpp
```diff
@@ -12,45 +12,57 @@
...
-      process_map_(std::move(process_map))
+      config_(std::move(config)),
+      process_handling_(std::move(process_handling))
 {
+
+    if (config.component_properties.application_profile.application_type ==
+        configuration::ApplicationType::ReportingAndSupervised)
+    {
+        config_.deployment_config.environmental_variables.add(
+            "LCM_ALIVE_INTERFACE_PATH", "/lifecycle_health_" + config_.name);
```
WilliamRoebuck: I think the value should be `alive_interface_path(config_.name)`. It's in `score/launch_manager/src/daemon/src/common/alive_interface_path.hpp`

WilliamRoebuck (commented):
Looks good, just one suggestion based on something I found in my PR

WilliamRoebuck (approved):

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
```diff
@@ -370,47 +302,31 @@ void ProcessGroupManager::processComponentEvents()
...
-        }
-    }
-    else
-    {
-        LM_LOG_ERROR() << "No startup state, exiting from process group manager";
-    }
-    return result;
+    LM_LOG_DEBUG() << "=============STARTING STARTUP STATE============";
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(bool(graph_), "Graph not initialized");
+    graph_->startInitialTransition(IdentifierHash{configuration_.initialRunTarget()});
```
NicolasFussberger, line 314: Nice, this now supports having arbitrary initial target.
Previously, this was hardcoded to "Startup" and this was also enforced on the config side.
If this limitation is gone, could you remove these parts:
* [score/launch_manager/docs/product_documentation/known_limitations.rst:41](https://github.com/eclipse-score/lifecycle/blob/8303fd39c50d73000edb527cd33fbd9b3bdfcf2f/score/launch_manager/docs/product_documentation/known_limitations.rst?plain=1#L41-L42)
* https://github.com/eclipse-score/lifecycle/blob/8303fd39c50d73000edb527cd33fbd9b3bdfcf2f/scripts/config_mapping/lifecycle_config.py#L498-L502

NicolasFussberger (approved):

score/launch_manager/src/daemon/src/process_group_manager/process_group_manager.cpp
NicolasFussberger: Though I am not sure if this was already made working in some previous PR. In that case we can also do it in a separate PR

NicolasFussberger (approved):

PR #456, opened 2026-08-12
Combine check workflows

danth:
Needed for https://github.com/eclipse-score/.eclipsefdn/pull/214

Checks to mark as required will be `PR / Ready to merge` and `PR Target / Ready to merge`.


MaciejKaszynski (approved):

.github/workflows/on-pr-target.yml
```diff
@@ -0,0 +1,100 @@
...
+      score-qnx-password: ${{ secrets.SCORE_QNX_PASSWORD }}
+
+  # This is only here as a dependency of the documentation, the main test job is in on-pr.yml.
+  testlogs:
+    name: Upload test logs
+    uses: eclipse-score/cicd-workflows/.github/workflows/tests.yml@93aac16ada7d247bbb6ae926509ddea74cf5213a # v0.0.2
+    permissions:
+      contents: read
+    with:
+      bazel-target: 'test --lockfile_mode=error //score/... //tests/... //scripts/... --config=x86_64-linux'
```
MaciejKaszynski, line 61: With this we would have a duplicate test stage. I understand why this is needed but I think we should remove this in the future.
I made an issue for this as I think we should still merge this https://github.com/eclipse-score/lifecycle/issues/459

Issue #422, opened 2026-08-06
Use common naming convention for mocks


NicolasFussberger:
Currently, in score/launch_manager/ there are mocks defined with various filename patterns:

score/launch_manager/src/lifecycle_client/src/mwlifecyclemanagermock.h
score/launch_manager/src/lifecycle_client/src/lifecyclemanagermock.cpp
score/launch_manager/src/lifecycle_client/src/report_running_mock.h
score/launch_manager/src/lifecycle_client/src/report_running_mock.cpp
score/launch_manager/src/lifecycle_client/src/applicationcontextmock.cpp
score/launch_manager/src/lifecycle_client/src/lifecyclemanagermock.h
score/launch_manager/src/lifecycle_client/src/mwlifecyclemanagermock.cpp
score/launch_manager/src/lifecycle_client/src/applicationcontextmock.h
score/launch_manager/src/daemon/src/process_group_manager/details/mock_component_event_queue.hpp
score/launch_manager/src/daemon/src/process_group_manager/details/mock_termination_callback.hpp
score/launch_manager/src/daemon/src/process_group_manager/details/mock_component.hpp

We should decide for a single pattern and rename all files accordingly.

Note that mocks under score/launch_manager/src/lifecycle_client/src/ are exposed to users, so renaming those files is a breaking change.

cameron-craig-etas:
The `mock_` convention seems to make the most sense. This is what the official GTest docs are using e.g https://google.github.io/googletest/gmock_cook_book.html#making-the-compilation-faster.

cameron-craig-etas:
I have also created a PR to add a pre-commit hook, that will detect and fix these problems in the future.

PR #457, opened 2026-08-12
Consistent naming of mocks

cameron-craig-etas:
#422

This PR tidies up the naming of our mocks. I have gone with the `mock_` prefix convention, as it:
- conforms to GTest conventions of `mock_component` files and `MockComponent` classes.
- is an existing style used within the code base, just not consistently

(Feel free to suggest an alternative naming convention. This is just what made sense to me)

This is a breaking change because mocks under `score/launch_manager/src/lifecycle_client/src/` are exposed to users.


TimoSteuerwaldETAS:
I like the prefix convention. 👍  

However for the integration tests, I would prefer using completely different names here.
In fact these aren't Mocks at all, these are rather Stubs or Fake objects. 
See also:  [Mocks Aren't Stubs](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs)

What do you think?

It is not needed to be done in this PR. I can also add a github issue for that and for conformity just leave it as proposed by you. :-)
 
Renaming to ControlClientStub or ControlClientFake would involve more than just a file rename anyways.

cameron-craig-etas:
> I like the prefix convention. 👍
> 
> However for the integration tests, I would prefer using completely different names here. In fact these aren't Mocks at all, these are rather Stubs or Fake objects. See also: [Mocks Aren't Stubs](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs)
> 
> What do you think?
> 
> It is not needed to be done in this PR. I can also add a github issue for that and for conformity just leave it as proposed by you. :-)
> 
> Renaming to ControlClientStub or ControlClientFake would involve more than just a file rename anyways.

Ah ok, I haven't looked too deep into these files, so I hadn't spotted that. I agree that these mocks, stubs (or whatever they may be) should be named more appropriately. I'll do this in a follow-up PR to prevent this one from getting any bigger.

TimoSteuerwaldETAS (approved):
Created successor story for renaming the MockControlClients to Stubs: https://github.com/eclipse-score/lifecycle/issues/458

Beside that, everything is fine.

danth (approved):

MaciejKaszynski (approved):

PR #460, opened 2026-08-12
Run `arm64` tests on GitHub Actions

danth:
Build and test on a real ARM machine provided by GitHub. This will help ensure that our code is cross platform.

Also, merged the build and test jobs, so that the build results can be reused for the tests rather than compiling everything twice.


danth:
The S-CORE fork of `rules_rust` has a compiler which is too old to build some of our dependencies on ARM, when not cross compiling.

Avoiding the fork is not easy because all our dependencies use it.

Baselibs uses QEMU for their `arm64-linux` unit tests rather than a real machine - maybe we could look into that.

